PR #551, opened 2021-04-09
fix: use direct link to the operator repository in the status

wtrocki:
Update link to the operator from the status

### DIFF
--- a/locales/cmd/cluster/status/active.en.toml
+++ b/locales/cmd/cluster/status/active.en.toml
@@ -34,5 +34,5 @@ Command requires configuration using kubeconfig for Kubernetes cluster running R
 Operator can be installed using OperatorHub.
 You can find operator by typing "rhoas" in search.
 
-For more information please refer to https://operatorhub.io/how-to-install-an-operator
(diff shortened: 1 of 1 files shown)


locales/cmd/cluster/status/active.en.toml
```diff
@@ -34,5 +34,5 @@ Command requires configuration using kubeconfig for Kubernetes cluster running R
 Operator can be installed using OperatorHub.
 You can find operator by typing "rhoas" in search.
 
-For more information please refer to https://operatorhub.io/how-to-install-an-operator
+For more information please visit Operator Github repository: https://github.com/redhat-developer/app-services-operator
```
craicoverflow, line 37:
```suggestion
For more information please visit Operator GitHub repository: https://github.com/redhat-developer/app-services-operator
```

craicoverflow (approved):

locales/cmd/cluster/status/active.en.toml
craicoverflow: Without running make pkger, this change would not be applied.

wtrocki: I though we had ci/cd job for checking this

PR #556, opened 2021-04-12
refactor(topic): refactor how config entries get created

### LINKED ISSUE #555: Refactor util method to create ConfigEntries for TopicSettings
### Enhancement description

The util method to create config entries while creating/updating a topic uses specific field as an argument (which is retenPeriod for now). This method can get bit cumbersome if we are to add multiple config entries.

### Proposed solution

The [CreateConfig](https://github.com/redhat-developer/app-services-cli/blob/main/pkg/kafka/topic/util.go#L14) should accept a map and return array of configEntries according to it. It should check for null strings in order to determine which fields are to be updated.

craicoverflow:
### Description

Resolves #555 

### Verification Steps

1. Create a topic and update a topic, be sure to provide a value for `retention.ms` to exercise this change.

### Checklist

- [x] Code Review completed
- [x] Verified independently by reviewer

### DIFF
--- a/pkg/cmd/kafka/topic/create/create.go
+++ b/pkg/cmd/kafka/topic/create/create.go
@@ -5,12 +5,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strconv"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/redhat-developer/app-services-cli/internal/localizer"
 
 	"github.com/redhat-developer/app-services-cli/pkg/connection"
-	"github.com/redhat-developer/app-services-cli/pkg/kafka/topic"
+	topicutil "github.com/redhat-developer/app-services-cli/pkg/kafka/topic"
 
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/flag"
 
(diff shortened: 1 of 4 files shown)


pkg/cmd/kafka/topic/create/create.go
```diff
@@ -280,10 +281,18 @@ func runInteractivePrompt(opts *Options) (err error) {
...
-	err = survey.AskOne(retentionPrompt, &opts.retentionMs, survey.WithValidator(topic.ValidateMessageRetentionPeriod))
+	err = survey.AskOne(retentionPrompt, &opts.retentionMs, survey.WithValidator(topicutil.ValidateMessageRetentionPeriod))
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
+
+func createConfigEntries(opts *Options) *[]strimziadminclient.ConfigEntry {
```
rkpattnaik780, line 292: Should we use topicUtil's CreateConfigEntries method here as well?

rkpattnaik780 (commented):
Able to create and update a topic, dropped a minor query.

pkg/cmd/kafka/topic/create/create.go
craicoverflow: It does, see line 297.

rkpattnaik780 (approved):

PR #571, opened 2021-04-13
Update the cli to use new SSO URL

abstractj:


abstractj:
@pmuir @wtrocki marked as draft to prevent accidental merge, as I understand we should only merge next week.

pmuir:
I've added this to #577 which also updates the kas-fleet-manager to prod as well

wtrocki:
@abstractj what we should use for operator? 

"https://identity.api.openshift.com/auth/realms/rhoas" is specific for CLI?

Our current URL is `https://sso.redhat.com/auth/realms/redhat-external`

Issue #595, opened 2021-04-19
RHOAS CLI bind should detect if operator exist on cluster and create CR instead of doing bindings


wtrocki:
### Feature or problem description

RHOAS cluster bind command currently does binding without any indication of binding created (traceability) and topology connection visible as we do not create CR/Require operator to be installed. When operator is available on cluster we should create CR instead and let the operator to deal with the binding giving users extra capability. This should be quite simple and effective way to satisfy all use cases. 

As follow up we can document entire detection functionality and make it available as mini SDK/contribut it back to SBO operator if possible

CC @DuncanDoyle @pedjak @kadel

wtrocki:
Allow forcing options for using SDK or operator etc.

wtrocki:
We have added `Operator First and SDK second capability`. I have started separate thread to discuss if we want to keep this feature in the RHOAS CLI or somewhere else

PR #599, opened 2021-04-19
fix: add support for creating operator based resource

wtrocki:
Add option to create CR instead of using SDK for binding.
Issue https://github.com/redhat-developer/app-services-cli/issues/595


wtrocki:
@DuncanDoyle This is working end to end, but I need to perform some minor code cleanup for it.
In case this is not merged I will send you google drive with binary containing this feature.

pkg/cluster/serviceBinding.go
```diff
@@ -33,7 +34,9 @@ type KubernetesClients struct {
 
 var deploymentResource = schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}
 
-func ExecuteServiceBinding(logger logging.Logger, serviceName string, ns string, appName string, forceCreationWithoutAsk bool) error {
+// TODO extract arguments
```
wtrocki, line 37: I have this almost there - moving arguments to structure.

pkg/cmd/cluster/bind/bind.go
```diff
@@ -69,6 +70,7 @@ func NewBindCommand(f *factory.Factory) *cobra.Command {
 	cmd.Flags().BoolVarP(&opts.forceCreationWithoutAsk, "yes", "y", false, localizer.MustLocalizeFromID("cluster.common.flag.yes.description"))
 	cmd.Flags().StringVarP(&opts.namespace, "namespace", "n", "", localizer.MustLocalizeFromID("cluster.common.flag.namespace.description"))
 	cmd.Flags().BoolVarP(&opts.ignoreContext, "ignore-context", "", false, localizer.MustLocalizeFromID("cluster.common.flag.ignoreContext.description"))
+	cmd.Flags().BoolVarP(&opts.useOperator, "force-operator", "", false, localizer.MustLocalizeFromID("cluster.bind.flag.onlyOperator.description"))
```
wtrocki, line 73: @pmuir We assume to use Operator by default when installed. If this flag is added and CLI fails to detect operator it will fail. Otherwise if operator is not detected SDK is used. 

Would you prefer to have more complex flags like:
```
--useOperator=true
--useSDK=true
```

Additionally we should provide name override for the service binding (as it is currently generated which might look bad)

pmuir: I like `--force-operator` - it's clear what's it's doing :-D What about `--force-sdk` as well?

wtrocki: Yes. Adding SDK here will be easy. I was hoping to have single flag but that is not possible due to this fallback logic

wtrocki:
@craicoverflow ready for review :)

pkg/cmd/cluster/bind/bind.go
```diff
@@ -102,21 +107,27 @@ func runBind(opts *Options) error {
...
-		opts.forceCreationWithoutAsk)
+	err = cluster.ExecuteServiceBinding(logger, &cluster.ServiceBindingOptions{
+		ServiceName:             *kafkaInstance.Name,
+		Namespace:               opts.namespace,
+		AppName:                 opts.appName,
+		ForceCreationWithoutAsk: opts.forceCreationWithoutAsk,
+		ForceUseOperator:        opts.forceOperator,
+		ForceUseSDK:             opts.forceSDK,
+		BindingName:             opts.bindingName,
+		BindAsFiles:             true,
```
wtrocki, line 129: Not sure if we need flag for it (added it to SDK only)

pkg/cmd/cluster/bind/bind.go
```diff
@@ -92,7 +97,7 @@ func runBind(opts *Options) error {
 	// In future config will include Id's of other services
 	if cfg.Services.Kafka == nil || opts.ignoreContext {
 		// nolint
-		selectedKafka, err := kafka.InteractiveSelect(connection, logger)
+		selectedKafka, err := kafka.InteractiveSelect(apiConnection, logger)
```
craicoverflow, line 100: It is possible that `selectedKafka` will be nil if there are no Kafka instances available.

Same issue caused a nil-pointer error in the `use` command: https://github.com/redhat-developer/app-services-cli/pull/602/files

pkg/cmd/cluster/bind/bind.go
```diff
@@ -102,21 +107,27 @@ func runBind(opts *Options) error {
 		opts.selectedKafka = cfg.Services.Kafka.ClusterID
 	}
 
-	api := connection.API()
-	kafkaInstance, _, error := api.Kafka().GetKafkaById(context.Background(), opts.selectedKafka).Execute()
+	api := apiConnection.API()
+	kafkaInstance, _, err2 := api.Kafka().GetKafkaById(context.Background(), opts.selectedKafka).Execute()
```
craicoverflow, line 111:
```suggestion
	kafkaInstance, _, err := api.Kafka().GetKafkaById(context.Background(), opts.selectedKafka).Execute()
```

nit: Why create new error?

pkg/cmd/cluster/bind/bind.go
```diff
@@ -102,21 +107,27 @@ func runBind(opts *Options) error {
...
 	if kafkaInstance.Name == nil {
 		return errors.New(localizer.MustLocalizeFromID("cluster.bind.error.emptyResponse"))
 	}
 
-	err = cluster.ExecuteServiceBinding(logger, *kafkaInstance.Name,
-		opts.namespace,
-		opts.appName,
-		opts.forceCreationWithoutAsk)
+	err = cluster.ExecuteServiceBinding(logger, &cluster.ServiceBindingOptions{
+		ServiceName:             *kafkaInstance.Name,
```
craicoverflow, line 122: Safest option to avoid nil-pointer error:

```suggestion
		ServiceName:             kafkaInstance.GetName(),
```

craicoverflow (approved):

pkg/cmd/cluster/bind/bind.go
wtrocki: Lint issues

wtrocki: Lint issues - creating new instance here

craicoverflow:
Is that a fix or feat?

wtrocki:
I seen that as fix initially but when you have asked I think it is feat :D

Issue #618, opened 2021-04-22
Login fails if there is not a default browser


b1zzu:
## Description

**Version**: 0.23.1

### Steps to reproduce

1. Remove the default browser for the OS settings (can be done on Gnome)
2. Run `rhoas login`

### Expected vs actual behaviour

The login fails with this error
```
Error: exit status 4
```
but I expected an error message that explains that a default browser is not set and therefore the `--print-sso-url` should be used, or even it could fall-back to `--print-sso-url` automatically

PR #624, opened 2021-04-22
fix(login): handle scenario when user is trying to log in through headless environment. 

### LINKED ISSUE #617: RHOAS login unavailable in headless environments
Because rhoas login pops up a browser window, it does not work in a headless environment.  Instead, it fails with a cryptic "Error: exit status 3" message. Many customers work in virtualized / headless environments or have rsh / ssh access only to connected environments, so this may be a blocker for them.

craicoverflow:
Fixes #617 #618 

This PR does a couple of things:

- Provides better explanations in the login help text that you should use the token-login when running in a virtual or headless environment.
- If a default browser cannot be open (or if xdg-open does not exist, instead print the SSO url to the terminal instead).
- Replace occurrences of "MAS-SSO" with "identity.api.openshift.com".


hawkdshrh:
@craicoverflow from localhost, it seems to work.  I get something like:

$ /data/nfs/development/git/managed-kafka/app-services-cli/rhoas login
Logging in...
Open the following URL in your browser to log in: 

https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/auth?client_id=rhoas-cli-prod&code_challenge=XXXXXXXXXXXXXXXXXXXXXXX&code_challenge_method=S256&grant_type=authorization_code&redirect_uri=http%3A%2F%2Flocalhost%3A2542%2Fsso-redhat-callback&response_type=code&scope=openid&state=yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy

Logged in successfully
Logging in to MAS-SSO...
Open the following URL in your browser to log in: 

https://identity.api.openshift.com/auth/realms/rhoas/protocol/openid-connect/auth?client_id=rhoas-cli-prod&code_challenge=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX&code_challenge_method=S256&grant_type=authorization_code&redirect_uri=http%3A%2F%2Flocalhost%3A19773%2Fmas-sso-callback&response_type=code&scope=openid&state=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa


Logged in successfully to MAS-SSO

You are now logged in as "kafka-double-ipa"

Let me try from another host

hawkdshrh:
@craicoverflow Doesn't seem to work from a remote host as the callback is being registered as localhost instead of the ssh target host, but I think it is close.

craicoverflow:
> @craicoverflow Doesn't seem to work from a remote host as the callback is being registered as localhost instead of the ssh target host, but I think it is close.

I think this will have to come down to documenting this as a limitation when running from a remote host - that you have to use token-based login. So you do not see the cryptic exit code 3 any more?

craicoverflow:
> Doesn't seem to work from a remote host as the callback is being registered as localhost instead of the ssh target host, but I think it is close.

@pmuir I think for this scenario as long as we have documented you should use offline tokens for logging in when in a remote session, we should be okay. What do you think?

pmuir:
There may be further improvements we can make but this seems like a good start.

craicoverflow:
> There may be further improvements we can make but this seems like a good start.

One way I have found which could actually improve this is by detecting if the current terminal session is an SSH session, by detecting if `$SSH_CLIENT` or `$SSH_CLIENT` environment variables are set. I've tested it out and it seems a reliable way to tell.

Ref: https://unix.stackexchange.com/a/120081

I can explore it a bit more on Monday, if you think we should still look to improve upon this.

I haven't actually experienced the original bug myself, even when SSHing into a RHEL VM.

```shell
❯ ssh ephelan@192.168.122.149
ephelan@vm-ip's password: 
Last login: Fri Apr 23 10:03:55 2021 from 192.168.122.1
[ephelan@rhel83 ~]$ rhoas login
Logging in...
Error: exec: "xdg-open": executable file not found in $PATH
[ephelan@rhel83 ~]$ 
```

^ This PR fixes above error by falling back to printing the URL when xdg-open fails.

The original error seen in  #617:
```
Instead, it fails with a cryptic "Error: exit status 3" message.
```

I have not actually seen this at all, so it's been quite difficult to verify if this is fixed or not. @hawkdshrh has this been resolved from the current fix?

hawkdshrh:
@craicoverflow Let me do a pull and check it and I will post up the results

hawkdshrh:
@craicoverflow I no longer get the "Error: exit status 3" message, but pasting the URL into a browser on the ssh client host doesn't work (login hangs), because the browser can't reach the callback at localhost:{port}/sso-redhat-callback?state=....

craicoverflow:
Thanks for verifying that the original error is gone - I can reproduce what you are getting currently. Will pick this up Monday :)

craicoverflow:
Postponing this to post-Summit.

Issue #627, opened 2021-04-26
Connect fails with NPE when there is no kafkas on the account


wtrocki:
## Description

Running connect with account that doesn't have any Kafkas returns error
```
 sh-4.4$ rhoas cluster connect
No Kafka instances were found.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1621650]

goroutine 1 [running]:
github.com/redhat-developer/app-services-cli/pkg/cmd/cluster/connect.runBind(0xc0001d89a0, 0xc000127c78, 0xc000127c78)
        /home/runner/work/app-services-cli/app-services-cli/pkg/cmd/cluster/connect/connect.go:98 +0x1d0
github.com/redhat-developer/app-services-cli/pkg/cmd/cluster/connect.NewConnectCommand.func1(0xc0003c8000, 0x27a97f0, 0x0, 0x0, 0x0, 0x0)
        /home/runner/work/app-services-cli/app-services-cli/pkg/cmd/cluster/connect/connect.go:56 +0x34
github.com/spf13/cobra.(*Command).execute(0xc0003c8000, 0x27a97f0, 0x0, 0x0, 0xc0003c8000, 0x27a97f0)
        /home/runner/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:852 +0x47c
github.com/spf13/cobra.(*Command).ExecuteC(0xc000207180, 0x1c37314, 0x6, 0xc000207180)
        /home/runner/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:960 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
        /home/runner/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:897
main.main()
        /home/runner/work/app-services-cli/app-services-cli/cmd/rhoas/main.go:72 +0x229

```

PR #628, opened 2021-04-26
fix: panic when no kafkas in organization

wtrocki:
Resolves
https://github.com/redhat-developer/app-services-cli/issues/627

This fixing problem with CLI crashing when there is no Kafka List returned due to list command returning nil.
We need to add check for nil for the commands.

### DIFF
--- a/pkg/cmd/cluster/bind/bind.go
+++ b/pkg/cmd/cluster/bind/bind.go
@@ -101,6 +101,9 @@ func runBind(opts *Options) error {
 		if err != nil {
 			return err
 		}
+		if selectedKafka == nil {
+			return nil
+		}
 		opts.selectedKafka = selectedKafka.GetId()
 	} else {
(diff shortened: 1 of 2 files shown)


craicoverflow:
Should it exit when there are no Kafka instances?

```
❯ ./rhoas cluster connect
No Kafka instances were found.
This command will link your cluster with Cloud Services by creating custom resources and secrets.
In case of problems please execute "rhoas cluster status" to check if your cluster is properly configured

Connection Details:

Apache Kafka instance:  
Kubernetes Namespace:   wtrocki-testing
Service Account Secret: rh-cloud-services-service-account

? Do you want to continue? (y/N) 
```

craicoverflow:
LGTM - but you need others to approve this.

craicoverflow (approved):

craicoverflow (approved):

PR #632, opened 2021-04-27
refactor: use the standard `os.UserConfigDir` API to get the config dir

### LINKED ISSUE #626: Use `os.UserConfigDir()` to store config file
We should use the standard library to get the config directory: https://golang.org/pkg/os/

craicoverflow:
Closes #626 

This PR achieves two things:

- Use the standard library for determining the config directory. We were implementing the same in the CLI with the exact same results.
- Change the default file name from `$HOME/.config/.rhoascli.json` to `$HOME/.config/rhoas/config.json`. Since the `.config` directory is already hidden, there is no longer a need to make the config hidden also. This change is backwards compatible by having a migration function in the code.
 
### Verification Steps

1. Run `rhoas whoami`
2. Check the location of your config file (it is different across OSs). It will have been renamed, but you will still be logged in.

### Type of change

- [ ] Documentation change
- [x] Refactor


cmd/rhoas/main.go
```diff
@@ -128,3 +137,24 @@ func initConfig(f *factory.Factory) {
...
+	}
+	oldFilePath := filepath.Join(userCfgPath, ".rhoascli.json")
+	if os.Getenv("RHOASCLI_CONFIG") == oldFilePath {
+		return nil
+	}
+	if _, err = os.Stat(oldFilePath); err == nil {
+		return os.Rename(oldFilePath, cfgPath)
+	}
+	return nil
+}
```
craicoverflow, line 160: Migrating the old config file. I want to start a habit of non-breaking changes now that we have users. This function can be removed after a few releases.

cmd/rhoas/main.go
```diff
@@ -128,3 +137,25 @@ func initConfig(f *factory.Factory) {
 		os.Exit(1)
 	}
 }
+
+// check if the config file is located in the old default location
+// if so, move it to the new location
+// TODO: Remove this function after 0.28.0 is released.
```
wtrocki, line 143: I would remove this and transform that to the issue (possibly not instantly removed since we do not control when people update etc.)

wtrocki: Since login to CLI is required everyday I would question if migration would be critical thing to do, but let's have it :D

craicoverflow: Here's a scenario:

1. User logs in - sees there is an update
2. Downloads the update
3. They have to log in again!!

Another scenario:

User is logged in and using their Kafka which is saved to their config. They update, have to log in and their Kafka is no long er saved as the current one.

These are not drastic things to haver happen. But I do want to build these good habits early.

wtrocki (approved):

wtrocki:
Verified cases of upgrade or existing config etc. All works well

PR #638, opened 2021-04-30
refactor: replace pkger with go:embed

### LINKED ISSUE #453: Replace pkger with go embed
Currently we are using pkger to embed static assets into the release binary. Since then, go 1.16 has been released with native support for this.

craicoverflow:
Closes #453 

## Description

This is a massive refactor which aims to address the poor developer experience of using [markbates/pkger](https://github.com/markbates/pkger) with [nicksnyder/go-i18n](https://github.com/nicksnyder/go-i18n). 

Since the latest release of Go (1.16), static file embedded is now natively supported. This meant that we could remove pkger as it was really a pain to work with in combination with go-i18n as the static files had to be regularly updated and  a slow re-compile was needed. Thankfully, go embed just works - no need to do any extra compiling which is great.

I've also abstracted from the i18n implementation while I was doing this refactor. This way, if a better library than go-18n comes along and we want to adopt it it should be more straightforward.

I've added some basic test cases: https://github.com/redhat-developer/app-services-cli/blob/a87734ffa8d6a308774200a1a716eabb8b974b2a/pkg/localize/goi18n/go_i18n_test.go

## Notes

I think we have gone too far with i18n - covering every single string file is not practical nor easy with Go. In TS/JS it is easier as you can natively imported a JSON file. I will form a discussion around refactoring this, so that we place clear limits on what should be internationalized.


craicoverflow:
@wtrocki what's your take on this?

Issue #346, opened 2021-02-16
[Idea] turn on debug mode for all commands


craicoverflow:
It would be quite nice to be able to persist debug mode without having to pass `--debug` all the time.

This could be in the form of a new `config` command:

```shell
$ rhoas config set debug true
Debug mode is now enabled for all commands..
```

wtrocki:
Working on this as we agreed that we cannot release CLi with current implementation.

wtrocki:
@craicoverflow all are `set debug true` an positional characters?

craicoverflow:
Yes in that example, but maybe there is a better way without so many sub-commands?

```
rhoas config set-debug true
```

As an example. Not sure if this is the best way.

wtrocki:
So I'm happy eather way. 

Do you feel that this command should enable editing any value of the config or we should support some well known use cases?

My take is that it will be easier to have multiple subcomands like set-debug etc. that handle specific cases rather than generic config writer that somewhat makes less sense as you might just open config and make edits there.

craicoverflow:
I think we should definitely only enable specific cases and not let this give the user the ability to modify anything. Let's only enable editing the debug mode for now.

PR #668, opened 2021-05-18
feat: add dev preview capability

wtrocki:
Creating as separate PR to review but this PR targets srs-cli branch that has SDK and initial commands.

### DIFF
--- a/pkg/cmd/root/root.go
+++ b/pkg/cmd/root/root.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/registry"
+	"github.com/redhat-developer/app-services-cli/pkg/profile"
 
 	"github.com/redhat-developer/app-services-cli/internal/build"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/login"
@@ -36,7 +37,6 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
 	cmd.Version = version
 
 	cmd.SetVersionTemplate(f.Localizer.MustLocalize("version.cmd.outputText", localize.NewEntry("Version", build.Version)))
-
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 
 	fs := cmd.PersistentFlags()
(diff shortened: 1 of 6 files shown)


pkg/arguments/arguments.go
```diff
@@ -1,13 +1,43 @@
...
+// Enables dev preview in config
+func EnableDevPreview(f *factory.Factory, enablement *string) (*config.Config, error) {
+	if *enablement == "" {
+		// Flag not present no action needed.
+		return nil, nil
+	}
+
+	logger, err := f.Logger()
+	if err != nil {
+		logger.Info("Cannot enable dev preview")
```
wtrocki, line 27: I will do i18n as last step on the feature branch. Ignore non externalized strings for now

pkg/cmd/root/root.go
```diff
@@ -46,17 +46,25 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
 	var help bool
 	fs.BoolVarP(&help, "help", "h", false, f.Localizer.MustLocalize("root.cmd.flag.help.description"))
 
+	var devpreview *string = nil
+	cmd.Flags().StringVar(devpreview, "devpreview", "", f.Localizer.MustLocalize("root.cmd.flag.devpreview.description"))
```
wtrocki, line 50: I spent most of the time thinking how to enable this properly. 
Choose string finally as the best way. 
Technically we can validate it for the proper values as well :)
This is as string because:

- boolean has only 2 possible values - we need 3
- It will simplify adding tiers later without introducing other flags etc.

pkg/cmd/root/root.go
```diff
@@ -46,17 +46,25 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
...
-	cmd.AddCommand(registry.NewServiceRegistryCommand(f))
 	cmd.AddCommand(serviceaccount.NewServiceAccountCommand(f))
 	cmd.AddCommand(cluster.NewClusterCommand(f))
 	cmd.AddCommand(status.NewStatusCommand(f))
 	cmd.AddCommand(completion.NewCompletionCommand(f))
 	cmd.AddCommand(whoami.NewWhoAmICmd(f))
 	cmd.AddCommand(cliversion.NewVersionCmd(f))
 
+	// Dev preview commands
+	if profile.DevPreviewEnabled(f) {
```
wtrocki, line 65: I have investigated using custom help message template based on the annotations, however that will require pulling some large parts of the internal cobra functions etc. so decided to simplify it by listing those commands bellow. This will be mostly seen by us so there is no big deal with it.

pkg/cmd/root/root.go
craicoverflow: What will the valid values be?

wtrocki: Open to your feedback.

Currently true, yes etc. but we can rename flqg and have alpha beta etc.

craicoverflow: https://github.com/spf13/cobra#persistent-flags

This should use persistent flags.

craicoverflow: The thing about string flags is the need an argument so you cannot just do `rhoas srs --devpreview` - its a bit more awkward to achieve.

pkg/arguments/arguments.go
craicoverflow, line 27: That will fail because the logger is not created.

pkg/cmd/root/root.go
```diff
@@ -46,17 +46,25 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
...
+		cmd.AddCommand(registry.NewServiceRegistryCommand(f))
```
craicoverflow, line 66:
```shell
❯ ./rhoas --devpreview=yes registry
Error: unknown command "registry" for "rhoas"
```

How do I get this to work?

pkg/cmd/root/root.go
wtrocki: Acording to this issue boolean ones are not going to help us

https://github.com/spf13/cobra/issues/613

This doesn't work this way. Srs will not be visible. DevPreview flag on it's own modifies config so its not needed when calling dev preview commands. 
```
rhoas srs --devpreview
```

wtrocki:
Update from call. We will refactor this command post POC like in the #346 , however for the moment, I will keep flags this way.

Pushing changes to feature branch.

wtrocki:
This is still printing help when dev preview is enabled.

PR #670, opened 2021-05-18
docs(kafka topic): remove rferences to replication factor

### LINKED ISSUE #665: Replication factor should be removed from example for topic update command (both CLI help and docs)
## Description

**Version**: 0.24.3

In the CLI help and in the CLI reference doc entry for `rhoas kafka topic update` https://access.redhat.com/documentation/en-us/red_hat_openshift_streams_for_apache_kafka/1/guide/bd840e21-c3f9-4978-b726-b24effb5c4e8#_b9095d68-7cfc-4236-a37c-7cee14ccab6a, the list of supported options includes only retention size and retention time, but the example command shown is for replication factor. 

```
sterobin@sterobin ~$ rhoas kafka topic update --help
Update a topic in the current Apache Kafka instance.

Usage:
  rhoas kafka topic update [flags]

Examples:
# update the number of replicas for a topic
$ rhoas kafka topic update topic-1 --replication-factor 


Flags:
  -o, --output string            Format in which to display the Kafka topic. Choose from: "json", "yml", "yaml" (default "json")
      --retention-bytes string   The maximum total size of a partition log segments before old log segments are deleted to free up space
      --retention-ms string      The period of time in milliseconds the broker will retain a partition log before deleting it

Global Flags:
  -d, --debug   Enable debug mode
  -h, --help    Show help for a command

```
As of today, my understanding is that the replication factor is pre-defined by default, and that the only available options for updating are the retention size and time. So the example should reflect one of the retention options only. Confirmed the same with @pmuir .

rkpattnaik780:
Remove references to replication factor in docs of `kafka topic` commands.

Closes #665 

### Type of change

- [X] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/docs/commands/rhoas_kafka_topic_create.adoc
+++ b/docs/commands/rhoas_kafka_topic_create.adoc
@@ -9,7 +9,7 @@ Create a topic
 Create a topic in the current Apache Kafka instance.
 
 This command lets you create a topic, set a desired number of 
-partitions, replicas and retention period or else use the default values.
+partitions, retention size and retention period or else use the default values.
 
 
 ....
--- a/docs/commands/rhoas_kafka_topic_update.adoc
+++ b/docs/commands/rhoas_kafka_topic_update.adoc
@@ -16,7 +16,7 @@ rhoas kafka topic update [flags]
 === Examples
 
 ....
-# update the number of replicas for a topic
+# update the message retention period for a topic
 $ rhoas kafka topic update topic-1 --retention-ms -1
 
 ....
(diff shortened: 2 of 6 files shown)


wtrocki:
So this is good but in future we might need versioning for docs as making change on unreleased cli will remove docs for people that still have this value

craicoverflow:
> So this is good but in future we might need versioning for docs as making change on unreleased cli will remove docs for people that still have this value

https://github.com/redhat-developer/app-services-cli/blob/0.24.3/docs/commands/rhoas.adoc

You get that with tags yes. I am not sure whether the published docs use these versions though.

wtrocki:
@craicoverflow Sorry for that - I meant strictly for what users will see on the page thru midstream. 
Will ask for details.

PR #676, opened 2021-05-19
chore: update open API specs

rkpattnaik780:
Updated client according to latest Open API specs.

### Type of change

- [ ] Documentation change
- [ ] Update Open API specs

### Checklist

- [ ] Code Review completed


openapi/kafka-service.yaml
```diff
@@ -1276,89 +1281,88 @@ components:
...
         reason: "Synchronous action is not supported, use async=true parameter"
         operation_id: "1iWIimqGcrDuL61aUxIZqBTqNRa"
     400CreationExample:
       value:
         id: "103"
         kind: "Error"
-        href: "/api/managed-services-api/v1/errors/103"
-        code: "MGD-SERV-API-103"
+        href: "/api/kafkas_mgmt/v1/errors/103"
+        code: "KAFKAS-MGMT-103"
```
craicoverflow, line 1293: You may have noticed in the OpenAPI document that the error codes have changed from `MGD-SERV-API-` to `KAFKAS-MGMT-`. We are using these error codes in some places and so they also need to be updated.

craicoverflow (changes requested):

openapi/kafka-service.yaml
rkpattnaik780: I had made some changes around it, for example, changing ERROR_CODE_PREFIX, Currently doing a global search for "MGD-SERV-API" returns no result from code, am I missing something?

craicoverflow (approved):

openapi/kafka-service.yaml
craicoverflow: My bad, misread it :)

PR #686, opened 2021-05-24
feat(completion): static completion for suitable flags

### LINKED ISSUE #682: Add auto completion for flags

CLI should have auto completion for flags wherever possible.

### Proposed solution

Flags such as `--output` should be able to autocomplete themselves based on allowed values.

rkpattnaik780:
Adds completions for flags such as `--output` and `--file-format`.

Closes #682 

### Verification Steps

1. Press tab after typing in the following command:
```
./rhoas kafka consumergroup describe -o
```
2. It should suggest the valid options for the output flag - `json`, `yaml` and `yml`

Screenshot from 2021-05-24 16-51-01

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/kafka/consumergroup/describe/describe.go
+++ b/pkg/cmd/kafka/consumergroup/describe/describe.go
@@ -16,6 +16,7 @@ import (
 	strimziadminclient "github.com/redhat-developer/app-services-cli/pkg/api/strimzi-admin/client"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/factory"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/flag"
+	flagutil "github.com/redhat-developer/app-services-cli/pkg/cmdutil/flags"
 	"github.com/redhat-developer/app-services-cli/pkg/color"
 	"github.com/redhat-developer/app-services-cli/pkg/connection"
 	"github.com/redhat-developer/app-services-cli/pkg/dump"
@@ -91,6 +92,10 @@ func NewDescribeConsumerGroupCommand(f *factory.Factory) *cobra.Command {
 	// flag based completions for ID
 	_ = cmd.RegisterFlagCompletionFunc("id", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return cmdutil.FilterValidConsumerGroupIDs(f, toComplete)
+	})
+
+	_ = cmd.RegisterFlagCompletionFunc("output", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return flagutil.ValidOutputFormats, cobra.ShellCompDirectiveNoSpace
 	})
 
 	return cmd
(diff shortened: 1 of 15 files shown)


pkg/cmd/kafka/consumergroup/describe/describe.go
```diff
@@ -93,6 +94,10 @@ func NewDescribeConsumerGroupCommand(f *factory.Factory) *cobra.Command {
 		return cmdutil.FilterValidConsumerGroupIDs(f, toComplete)
 	})
 
+	_ = cmd.RegisterFlagCompletionFunc("output", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
craicoverflow, line 97: Good work but this is being duplicated in every file and they are the same. Could you extract this into the flagutil package?

pkg/cmd/serviceaccount/resetcredentials/reset_credentials.go
```diff
@@ -79,6 +79,10 @@ func NewResetCredentialsCommand(f *factory.Factory) *cobra.Command {
 	cmd.Flags().StringVar(&opts.fileFormat, "file-format", "", opts.localizer.MustLocalize("serviceAccount.common.flag.fileFormat.description"))
 	cmd.Flags().BoolVarP(&opts.force, "yes", "y", false, opts.localizer.MustLocalize("serviceAccount.resetCredentials.flag.yes.description"))
 
+	_ = cmd.RegisterFlagCompletionFunc("file-format", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
craicoverflow, line 82: This is also duplicated. I would say any one with more than one occurrence can be extracted to flagutil.

craicoverflow (commented):
Verified as working but would like to see deduplication. Re-request my review when that is done! :)

pkg/cmd/kafka/consumergroup/describe/describe.go
```diff
@@ -93,6 +94,8 @@ func NewDescribeConsumerGroupCommand(f *factory.Factory) *cobra.Command {
 		return cmdutil.FilterValidConsumerGroupIDs(f, toComplete)
 	})
 
+	flagutil.EnableStaticFlagCompletion(cmd, "output", flagutil.ValidOutputFormats)
```
craicoverflow, line 97: nitpick: this could be further simplified by having a specific util for "output". There is still repetition here and the values will likely always be the same for the value flag name and the valid values.

Something like `flagutil.RegisterOutputFlagCompletion()`

craicoverflow (approved):
Great work 👍🏻 

There may be more flags worth doing this for too.

pkg/cmd/kafka/consumergroup/describe/describe.go
rkpattnaik780: Having a separate method for `--output` will indeed reduce repetition. I will suggest to have `RegisterOutputFlagCompletion` for output flag and `RegisterOutputFlagCompletion` for other flags, else we'll have to define a new method for every flag that will require static completion. What do you think?

craicoverflow: I don't think we should define one for every flag, but for the extremly common ones like this it is good.

rkpattnaik780: Kept the `EnableStaticFlagCompletion` with a separate method to enable output flag.

craicoverflow (approved):

Issue #704, opened 2021-06-04
Erroneous message shown when no consumergroups are available for a topic


rkpattnaik780:
## Description

Erroneous message printed when no consumer groups are available for a topic in a given Kafka instance.

**Version**: #

### Steps to reproduce

1. Run `./rhoas kafka consumergroup list --topic topic-1` where topic-1 is a topic not associated with any consumer group.
2. It should throw the error `Kafka instance "<no value>" has no consumer groups for topic "<no value>" &{InstanceName test-instance} &{TopicName topic-1}`

### Expected vs actual behaviour

The error message thrown should be:

> Kafka instance "<instance name>" has no consumer groups for topic "topic-1"

### Logs

![Screenshot from 2021-06-04 17-02-23](https://user-images.githubusercontent.com/23582438/120804371-30bb9a00-c562-11eb-87d0-f9734c3998ee.png)

PR #705, opened 2021-06-05
fix(kafka create): run interactive if no name provided and fix typos

### LINKED ISSUE #703: `rhoas kafka create` should prompt for name if not provided
## Description

`rhoas kafka create` doesn't run in interactive mode as expected.

**Version**: 0.24.3

### Steps to reproduce

Run `rhoas kafka create --provider aws`. It should output:
> Error: name does not match ^[a-z]([-a-z0-9]*[a-z0-9])?$

The cli should prompt for name but instead sends request with name as `""`

### Logs

Screenshot from 2021-06-04 17-02-15

rkpattnaik780:
- Open interactive mode when `rhoas kafka create` is run with flags but no name.
- Fix faulty error message while filtering consumergroups by topic.

Closes #703, #704

### Verification Steps

1. Run `rhoas kafka create --provider aws`, it should open the interactive mode.
2. Run `rhoas kafka consumergroup list --topic <topic-name>` where the topic has no associated consumergroup, it should display the message `Kafka instance "<instance name>" has no consumer groups for topic "<topic-name>"`

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/localize/locales/en/cmd/kafka_create.en.toml
+++ b/pkg/localize/locales/en/cmd/kafka_create.en.toml
@@ -69,6 +69,9 @@ one = "Geographical region where the Kafka instance will be deployed"
 [kafka.create.argument.name.error.requiredWhenNonInteractive]
 one = 'name is required. Run "rhoas kafka create <name>"'
 
+[kafka.create.flags.notAllowedWhenInteractive]
+one = 'region and provider flags not allowed for interactive Kafka instance creation'
+
 [kafka.create.debug.autoUseSetMessage]
 one = 'Auto-use Kafka is set, updating the current Kafka instance'
 
--- a/pkg/localize/locales/en/cmd/kafka_create.en.toml
+++ b/pkg/localize/locales/en/cmd/kafka_create.en.toml
@@ -69,9 +69,6 @@ one = "Geographical region where the Kafka instance will be deployed"
 [kafka.create.argument.name.error.requiredWhenNonInteractive]
 one = 'name is required. Run "rhoas kafka create <name>"'
 
-[kafka.create.flags.notAllowedWhenInteractive]
-one = 'region and provider flags not allowed for interactive Kafka instance creation'
-
 [kafka.create.debug.autoUseSetMessage]
 one = 'Auto-use Kafka is set, updating the current Kafka instance'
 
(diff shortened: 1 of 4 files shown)


craicoverflow:
image

It looks like the **Closing** keyword does not link multiple issues automatically. In this case you can manually link them in the side bar.

craicoverflow (commented):
When I pass the `--provider` flag, it will re-prompt me to enter the provider value in the interactive prompt:

```shell
❯ ./rhoas kafka create --provider aws
? Name: my-kafka
? Cloud Provider:  [Use arrows to move, type to filter]
> aws
```

The way it was (supposed to be) working before now was as follows: 

If no name argument is passed, and any value flag _is_ passed, such as provider, then throw an error as the user is trying to mix interactive with non-interactive Kafka creation.

I think that this is more maintainable in the long run, and I believe it is consistent with the other **create** commands (not sure though).

rkpattnaik780:
> I think that this is more maintainable in the long run, and I believe it is consistent with the other **create** commands (not sure though).

Had used `topic create` as a reference. We should update that one later as well.
Screenshot from 2021-06-09 16-06-18

pkg/localize/locales/en/cmd/kafka_create.en.toml
```diff
@@ -69,6 +69,9 @@ one = "Geographical region where the Kafka instance will be deployed"
 [kafka.create.argument.name.error.requiredWhenNonInteractive]
 one = 'name is required. Run "rhoas kafka create <name>"'
 
+[kafka.create.flags.notAllowedWhenInteractive]
+one = 'region and provider flags not allowed for interactive Kafka instance creation'
```
rkpattnaik780, line 73: Suggestion for a better error message @craicoverflow ?

craicoverflow: Swap it round like something like this: `"name" argument is required when creating a Kafka instance`

rkpattnaik780: Changed it to `'name is required. Run "rhoas kafka create <name>"'`

craicoverflow (approved):

PR #706, opened 2021-06-07
chore: expose binding name

wtrocki:
1. Expose binding name to support multiple bindings

> NOTE: The only flag that we do not expose now is `bindAsFiles` 

This is because we do not have any documentation for environment variables binding and using it will not work without any doc.

### DIFF
--- a/pkg/localize/locales/en/cmd/cluster_bind.en.toml
+++ b/pkg/localize/locales/en/cmd/cluster_bind.en.toml
@@ -29,6 +29,9 @@ $ rhoas cluster bind --namespace=ns --app-name=myapp
 [cluster.bind.flag.appName]
 one = '''Name of the kubernetes deployment to bind'''
 
+[cluster.bind.flag.bindName]
+one = '''Name of the Service binding object to create when using operator'''
+
 [cluster.bind.flag.forceOperator.description]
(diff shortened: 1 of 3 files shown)


pkg/localize/locales/en/cmd/cluster_bind.en.toml
```diff
@@ -29,6 +29,9 @@ $ rhoas cluster bind --namespace=ns --app-name=myapp
 [cluster.bind.flag.appName]
 one = '''Name of the kubernetes deployment to bind'''
 
+[cluster.bind.flag.bindName]
+one = '''Name of the Service binding object to create when using operator'''
```
craicoverflow, line 33:
```suggestion
one = 'Name of the Service binding object to create when using operator'
```

The three quotes are only necessary for multi-line strings.

craicoverflow (approved):
This is a **feat** type change, not a **chore**.

> chore: (updating grunt tasks etc; no production code change)

Since the changelog is reliant on our conventional commits please change this before merging. Thanks!

wtrocki:
@craicoverflow Thank you. Sorry for using chore (lots of chores done recently). 

My recommendation would be to squash this into:

feat: Add ability to specify a binding name using binding-name flag

PR #709, opened 2021-06-10
feat(kafka topic): add search flag to list subcommand

### LINKED ISSUE #697: Add search flag for `kafka topic list`

Currently there doesn't exist any means to filter the kafka topics.

### Proposed solution

Add a `--search` flag for `rhoas kafka topic list` to filter topics by name.

```
rhoas kafka topic list --search "topic-1"
```

rkpattnaik780:
Allow user to filter list of kafka topics by passing `--search` flag to `rhoas kafka topic list`

Closes #697 

Screenshot from 2021-06-10 16-00-37

### Verification Steps

1. > go run ./cmd/rhoas kafka topic list --search "pic(-1"
2. It should throw the error message `Error: illegal search value "pic(-1"; only letters (Aa-Zz), numbers, "_", "." and "-" are accepted`
3. Run `go run ./cmd/rhoas kafka topic list --search <search input>`

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/kafka/topic/validators.go
+++ b/pkg/kafka/topic/validators.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 
 	"github.com/redhat-developer/app-services-cli/pkg/common/commonerr"
+	"github.com/redhat-developer/app-services-cli/pkg/localize"
 	kafkainstanceclient "github.com/redhat-developer/app-services-sdk-go/kafkainstance/apiv1internal/client"
 )
 
@@ -43,6 +44,23 @@ func ValidateName(val interface{}) error {
 	}
 
 	return fmt.Errorf(`invalid topic name "%v"; only letters (Aa-Zz), numbers, "_", "." and "-" are accepted`, name)
+}
+
+func ValidateSearchInput(val interface{}, localizer localize.Localizer) error {
+
+	search, ok := val.(string)
+	if !ok {
+		return commonerr.NewCastError(val, "string")
+	}
+
+	matched, _ := regexp.Match(legalNameChars, []byte(search))
+
+	if matched {
+		return nil
+	}
+
+	return fmt.Errorf(localizer.MustLocalize("kafka.topic.list.error.illegalSearchValue", localize.NewEntry("Search", search)))
+
 }
 
 // ValidatePartitionsN performs validation on the number of partitions v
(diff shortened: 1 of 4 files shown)


pkg/kafka/topic/validators.go
```diff
@@ -45,6 +46,23 @@ func ValidateName(val interface{}) error {
 	return fmt.Errorf(`invalid topic name "%v"; only letters (Aa-Zz), numbers, "_", "." and "-" are accepted`, name)
 }
 
+func ValidateSearchInput(val interface{}, localizer localize.Localizer) error {
```
craicoverflow, line 49: Could you just re-use `ValidateName`?

pkg/localize/locales/en/cmd/kafka_topic_list.en.toml
```diff
@@ -19,4 +19,16 @@ $ rhoas kafka topic list -o json
 '''
 
 [kafka.topic.list.log.info.noTopics]
-one = 'Kafka instance "{{.InstanceName}}" has no topics. Run "rhoas kafka topic create <topic-name>" to create one'
\ No newline at end of file
+one = 'Kafka instance "{{.InstanceName}}" has no topics. Run "rhoas kafka topic create <topic-name>" to create one'
```
craicoverflow, line 22: Probably this message should not be used when the search flag is use - there could be topics but not showing up because of the filter value.

craicoverflow (changes requested):
Couple of small changes needed.

pkg/kafka/topic/validators.go
rkpattnaik780: what if the user would want to filter with "." or ".."?

pkg/localize/locales/en/cmd/kafka_topic_list.en.toml
rkpattnaik780: Indeed! `--search` for kafka instances returns `No kafka instances were found` for both no instance created and no instance found after searching. Should we modify the message in same manner for consistency?

pkg/kafka/topic/validators.go
craicoverflow: Good point.

pkg/localize/locales/en/cmd/kafka_topic_list.en.toml
craicoverflow: `No Kafka instance were found` is good because it does not call out that there are no topics - just that we did not find any. Sure, let's change the topic message to be consistent :)

craicoverflow (approved):

PR #720, opened 2021-06-15
fix(serviceaccount reset-credentials): validate serviceaccount ID in prompt

### LINKED ISSUE #716: Add validation for the Service Account ID in interactive mode
```shell
? Service Account ID: https://api.stage.openshift.com/?urls.primaryName=Service%20Registry%20service%20fleet%20manager%20service
? Credentials file format:  [Use arrows to move, type to filter, ? for more help]
> env
  json
  properties
❯ rhoas serviceaccount reset-credentials
```

If I interactively enter the service account ID, it should validate that the ID is a UUID.

rkpattnaik780:
Client side validation for service account ID in interactive mode.

Closes #716

### Verification Steps

1. Run command to reset credentials for service account in interactive mode.
```
go run ./cmd/rhoas serviceaccount reset-credentials
```
2. Input any random string that is not an UUID, e.g `"kafka"`. It should throw the error:
```
Sorry, your reply was invalid: "kafka" is not a valid UUID
```

Screenshot from 2021-06-15 18-53-53

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/serviceaccount/validation/validation.go
+++ b/pkg/serviceaccount/validation/validation.go
@@ -14,7 +14,7 @@ const (
 	legalNameChars = "^[a-z]([-a-z0-9]*[a-z0-9])?$"
 	maxNameLength  = 50
 	minNameLength  = 1
-	leagalUUID     = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+	legalUUID      = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
 	// description validation rules
 	legalDescriptionChars = "^[a-zA-Z0-9.,\\-\\s]*$"
 	maxDescriptionLength  = 255
@@ -67,14 +67,14 @@ func ValidateDescription(val interface{}) error {
 }
 
 // ValidateID validates if ID is a valid UUID
-func ValidateID(localizer localize.Localizer) func(v interface{}) error {
+func ValidateUUID(localizer localize.Localizer) func(v interface{}) error {
 	return func(val interface{}) error {
 		id, ok := val.(string)
 		if !ok {
 			return commonerr.NewCastError(val, "string")
 		}
 
-		matched, _ := regexp.Match(leagalUUID, []byte(id))
+		matched, _ := regexp.Match(legalUUID, []byte(id))
 
 		if matched {
 			return nil
--- a/pkg/serviceaccount/validation/validation.go
+++ b/pkg/serviceaccount/validation/validation.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 
 	"github.com/redhat-developer/app-services-cli/pkg/common/commonerr"
+	"github.com/redhat-developer/app-services-cli/pkg/localize"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


rkpattnaik780:
Should client side validation be extended to non-interactive mode as well?
```
validateIDFlag := validation.ValidateID(opts.localizer)
validID := validateIDFlag(opts.id)
if validID != nil {
return validID
}
```
Also, it seems bit complex to write tests for higher order functions being used as validators.

pkg/serviceaccount/validation/validation.go
```diff
@@ -63,3 +65,22 @@ func ValidateDescription(val interface{}) error {
 
 	return errors.New(`invalid service account description; only alphanumeric characters and "-", ".", "," are accepted`)
 }
+
+// ValidateID validates if ID is a valid UUID
+func ValidateID(localizer localize.Localizer) func(v interface{}) error {
```
craicoverflow, line 70:
```suggestion
func ValidateUUID(localizer localize.Localizer) func(v interface{}) error {
```

pkg/serviceaccount/validation/validation.go
```diff
@@ -6,13 +6,15 @@ import (
...
 	"github.com/redhat-developer/app-services-cli/pkg/common/commonerr"
+	"github.com/redhat-developer/app-services-cli/pkg/localize"
 )
 
 const (
 	// name validation rules
 	legalNameChars = "^[a-z]([-a-z0-9]*[a-z0-9])?$"
 	maxNameLength  = 50
 	minNameLength  = 1
+	leagalUUID     = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
```
craicoverflow, line 17:
```suggestion
	legalUUID     = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
```

craicoverflow (commented):
> Should client side validation be extended to non-interactive mode as well?

Yes, we should. Any client-side validation improves UX/DX so this would be nice to have.

> Also, it seems bit complex to write tests for higher order functions being used as validators.

Yeah, I see what you mean! Now that we are needing to pass `localizer` to validators they are no longer simple functions.

What would be better (and you can do this in this PR) is to create a `Validator` type which contains all the validator functions for service accounts.

pkg/serviceaccount/validation/validation.go
```diff
@@ -63,3 +65,22 @@ func ValidateDescription(val interface{}) error {
 
 	return errors.New(`invalid service account description; only alphanumeric characters and "-", ".", "," are accepted`)
 }
+
+// ValidateUUID validates if ID is a valid UUID
+func ValidateUUID(localizer localize.Localizer) func(v interface{}) error {
```
craicoverflow, line 70: I think it will be a good idea for readibility if we do not use anonymous return functions anymore for all validators. I'll explain in a follow up issue

pkg/serviceaccount/validation/validation_test.go
```diff
@@ -69,6 +71,58 @@ func TestValidateName(t *testing.T) {
 	}
 }
 
+func TestValidateUUID(t *testing.T) {
```
craicoverflow, line 74: ❤️

craicoverflow (approved):

pkg/serviceaccount/validation/validation.go
rkpattnaik780: > PS: It would also be nice to validate the flag for UUId like you had suggested.

It has been done in `reset_credentials.go`. Did you expect something else?

rkpattnaik780: > I think it will be a good idea for readibility if we do not use anonymous return functions anymore for all validators. I'll explain in a follow up issue

Maked #721 as WIP, it has localizations for validators using anonymous functions, should this PR wait for the issue?

craicoverflow: No you can merge this PR beforehand if you like.

Issue #633, opened 2021-04-27
Golang versions are not matching our release versions


wtrocki:
## Description

https://pkg.go.dev/github.com/redhat-developer/app-services-cli?tab=versions

craicoverflow:
From [Modules Overview](https://golang.org/ref/mod#modules-overview):

> A version identifies an immutable snapshot of a module, which may be either a release or a pre-release. Each version starts with the letter v, followed by a semantic version. See Semantic Versioning 2.0.0 for details on how versions are formatted, interpreted, and compared.

> To summarize, a semantic version consists of three non-negative integers (the major, minor, and patch versions, from left to right) separated by dots. The patch version may be followed by an optional pre-release string starting with a hyphen. The pre-release string or patch version may be followed by a build metadata string starting with a plus. For example, v0.0.0, v1.12.134, v8.0.5-pre, and v2.0.9+meta are valid versions.

From this, I think the reason this is happening is because the versions have not been prefixed with "v".

We should really start prefixing the release tags with "v" going forward. This will keep the project consistent with other tools like [Odo](https://pkg.go.dev/github.com/redhat-developer/odo?tab=versions).

craicoverflow:
Fixed now that we have prefixed tags with "v": https://pkg.go.dev/github.com/redhat-developer/app-services-cli?tab=versions

PR #723, opened 2021-06-16
chore: fix install script

craicoverflow:
Fixes a bug where the binary could not be extracted as the tag has a `v` prefix, but the asset does not, and the tag is used to dynamically build the asset name.

## Verification

Install latest:

```shell
curl -o- https://raw.githubusercontent.com/redhat-developer/app-services-cli/fix-install-script/scripts/install.sh | bash
```

Install with "v" prefix:

```shell
curl -o- https://raw.githubusercontent.com/redhat-developer/app-services-cli/fix-install-script/scripts/install.sh | bash -s v0.24.4
```

Install without prefix:

```shell
curl -o- https://raw.githubusercontent.com/redhat-developer/app-services-cli/fix-install-script/scripts/install.sh | bash -s 0.24.4
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -61,6 +61,10 @@
     DOWNLOAD_TAG="$RELEASE_TAG"
   fi
 
+  if [[ $DOWNLOAD_TAG == v* ]]; then
+    DOWNLOAD_TAG="${DOWNLOAD_TAG:1}"
+  fi
+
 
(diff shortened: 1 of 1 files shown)


wtrocki:
[Offtopic] I never got why we were adding v to the tag :D

craicoverflow:
> [Offtopic] I never got why we were adding v to the tag :D

We dicussed this here: https://github.com/redhat-developer/app-services-cli/issues/633

Issue #774, opened 2021-07-01
Add `-` to service-account and consumer-groups commands


wtrocki:

wtrocki:
Closing. Leaving merge to @craicoverflow

PR #783, opened 2021-07-02
Renames for service account and consumer group commands

wtrocki:
Use hyphenated names for all multiword commands.
This effort will help us to standarize format for all commands we have.
This is breaking change but it will be delivered with breaking backend change so everyone will be advised to update CLI.

Docs were updated:
https://github.com/redhat-developer/app-services-guides/pull/191/files
Fixes
https://github.com/redhat-developer/app-services-cli/issues/774


craicoverflow (changes requested):
i18n files need to be updated. All good otherwise

wtrocki:
> i18n files need to be updated

Which ones? The only ones were the top level ones

craicoverflow:
> > i18n files need to be updated
> 
> Which ones? The only ones were the top level ones

Nope. There are lots of `examples` defined.

wtrocki:
Yes.. Looking at this now.

rkpattnaik780:
Global search shows few things still need to be changed in toml files

wtrocki:
Done changes in examples

rkpattnaik780 (approved):

PR #795, opened 2021-07-05
fix: remove old workaround for migrating config file name

wtrocki:
Creating for discussion rather than merge directly.
Might be to early to remove it, but I counted +5 releases since this was introduced.


craicoverflow (approved):

wtrocki:
Thank you for review. 
Reviewing now entire CLI for such things. I will keep couple small PR's like this and merge them together for better testing.

PR #798, opened 2021-07-06
Config command support

wtrocki:
## Verification

```
[wtrocki@graphapi app-services-cli (config)]$ rhoas config devPreview true
Developer Preview commands activated. Use help command to view them.
[wtrocki@graphapi app-services-cli (config)]$ rhoas config devPreview trueds
Error: invalid argument "trueds" for "rhoas config devPreview"
[wtrocki@graphapi app-services-cli (config)]$ rhoas config devPreview false 
Developer Preview commands deactivated.
[wtrocki@graphapi app-services-cli (config)]$ 
```

## Approach

In 0b79620 I have used single command approach and 2 arguments - one which is config key second is value. 
I think that is overly complex and convoluted.
Current approach uses separate command for each key. 

Drawback is for that approach we getting documentation file generated and there is little bit of duplication in docs, but much simpler and easier to work with. 

## What is outstanding

- [x] base implementation
- [x] unit tests
- [x] Error handling mapping instead of system error
- [x] I18n :)

### DIFF
--- a/pkg/cmd/config/config.go
+++ b/pkg/cmd/config/config.go
@@ -1,0 +1,79 @@
+package status
+
+import (
+	"errors"
+	"strconv"
+	"strings"
+
+	"github.com/redhat-developer/app-services-cli/pkg/localize"
+	"github.com/redhat-developer/app-services-cli/pkg/profile"
+
+	"github.com/redhat-developer/app-services-cli/internal/config"
+	"github.com/redhat-developer/app-services-cli/pkg/iostreams"
+	"github.com/redhat-developer/app-services-cli/pkg/logging"
+
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/factory"
+	"github.com/spf13/cobra"
+)
+
+type Options struct {
+	IO        *iostreams.IOStreams
+	Config    config.IConfig
+	Logger    func() (logging.Logger, error)
+	localizer localize.Localizer
+
+	devPreview bool
+}
+
+const DevPreviewConfigArgument = "devPreview"
+
+var validArguments = []string{DevPreviewConfigArgument}
+
+func NewConfigCommand(f *factory.Factory) *cobra.Command {
+	opts := &Options{
+		IO:        f.IOStreams,
+		Config:    f.Config,
+		Logger:    f.Logger,
+		localizer: f.Localizer,
+	}
+
+	cmd := &cobra.Command{
+		Use:       "config",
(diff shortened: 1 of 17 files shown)


wtrocki:
@craicoverflow that will be quick review (small command/refactoring here)

pkg/cmd/config/config.go
```diff
@@ -0,0 +1,58 @@
...
+
+	cmd := &cobra.Command{
+		Use:     "config",
+		Short:   opts.localizer.MustLocalize("config.cmd.shortDescription"),
+		Long:    opts.localizer.MustLocalize("config.cmd.longDescription"),
+		Example: opts.localizer.MustLocalize("config.cmd.example"),
+	}
+
+	devPreview := &cobra.Command{
+		Use:       "devPreview",
```
craicoverflow, line 41:
```suggestion
		Use:       "dev-preview",
```

Id prefer to use kebab-case for commands.

pkg/localize/locales/en/cmd/config.en.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[config.cmd.longDescription]
+one = '''
+Change configuration of the cli by executing one of the available subcommands
+'''
+
+[config.cmd.example]
+one = '''
+# change dev preview configuration
+$ rhoas config devpreview true
```
craicoverflow, line 12:
```suggestion
$ rhoas config dev-preview true
```

pkg/localize/locales/en/cmd/config.en.toml
```diff
@@ -0,0 +1,31 @@
...
+[devpreview.cmd.longDescription]
+one = '''
+Dev preview is used to enable commands in the CLI that are not available for general use. 
+Those commands usually work against alpha versions of the API and might often have breaking changes.
+'''
+
+[devpreview.cmd.example]
+one = '''
+# change dev preview configuration
+$ rhoas config devpreview true
```
craicoverflow, line 27:
```suggestion
$ rhoas config dev-preview true
```

craicoverflow (approved):
I suggest changing the command to `dev-preview` - all good otherwise

pkg/cmd/config/config.go
wtrocki: WOW. Totally forgot! 
Actually our ruleset should catch it once we add regexp rule. CC @ankithans

PR #806, opened 2021-07-07
fix: create folder for the config

wtrocki:


cmd/rhoas/main.go
```diff
@@ -94,12 +95,33 @@ func generateDocumentation(rootCommand *cobra.Command) {
 	}
 }
 
-func initConfig(f *factory.Factory) {
+func initConfig(f *factory.Factory, logger logging.Logger) {
+	rhoasCfgDir, err := config.DefaultDir()
+	if err != nil {
+		logger.Errorf("Error fetching config dir.", err)
```
craicoverflow, line 101: nit: would it be better to return the error from `initConfig` and log it in `main`? Saves having to pass the logger.

cmd/rhoas/main.go
```diff
@@ -94,12 +95,33 @@ func generateDocumentation(rootCommand *cobra.Command) {
...
+	}
+
+	customFolder, isCustomFolderSet := os.LookupEnv("RHOASCONFIG")
+	if isCustomFolderSet {
+		rhoasCfgDir = customFolder
```
craicoverflow, line 106: `RHOASCONFIG` actually points to the file itself, not the directory. So it could be `$HOME/rhoascustomfile.json`. Here you are assigning it to be the holding folder, and so later it is creating the folder called `$HOME/rhoascustomfile.json`

craicoverflow (changes requested):
I had issues when setting the custom config location:

```shell
❯ export RHOASCONFIG=~/.rhoascustom
❯ ./rhoas whoami
unable to read config file: read /home/ephelan/.rhoascustom: is a directory
```

I think if `RHOASCONFIG` is set, and the directory does not exist we should not try to create the directory for them. Only when the system defaulr dir does not exist should we create it.

cmd/rhoas/main.go
wtrocki: I had that and changed it as I thought it will be folder.

wtrocki:
Actually handling for RHOASCONFIG is done in other place.

wtrocki:
This is much cleaner now

wtrocki:
The only thing that is see right now is that you still going to get default folder created even if custom location is set. 
My take is that we should always create that folder.

craicoverflow:
> The only thing that is see right now is that you still going to get default folder created even if custom location is set.
> My take is that we should always create that folder.

We could also check if the env is set, and only create the folder if it is not set. But I'm happpy with how it is right now.

cmd/rhoas/main.go
```diff
@@ -94,20 +99,35 @@ func generateDocumentation(rootCommand *cobra.Command) {
...
+
 	cfgFile, err := f.Config.Load()
 
 	if cfgFile != nil {
-		return
+		return err
 	}
+
 	if !os.IsNotExist(err) {
 		fmt.Fprintln(f.IOStreams.ErrOut, err)
```
craicoverflow, line 123: Any need? This will duplicate the message being printed.

cmd/rhoas/main.go
```diff
@@ -94,20 +99,35 @@ func generateDocumentation(rootCommand *cobra.Command) {
...
-		os.Exit(1)
+		return err
 	}
 
 	cfgFile = &config.Config{}
 	if err := f.Config.Save(cfgFile); err != nil {
 		fmt.Fprintln(f.IOStreams.ErrOut, err)
```
craicoverflow, line 129: Any need? This will duplicate the message being printed.

craicoverflow (approved):

wtrocki:
@craicoverflow I tried to add `	HasCustomLocation: func() bool ` to config but I seem to have problem with mock generator right now. Do not want to use this env variable in two places

wtrocki:
I think that would be good right now

craicoverflow (approved):

PR #808, opened 2021-07-07
fix: invalid location of shared service i18n files

wtrocki:
The shared file were not placed properly in the folder that had language. 

I verified is using account without AMS enablement


wtrocki:
@famartinrh this should solve your problem

craicoverflow (approved):

PR #817, opened 2021-07-08
docs: remove registry commands

wtrocki:
files are shown downstream and they should be removed


craicoverflow:
Is there a way we can prevent this from happening after the next generation? Like a filter or something for dev-preview commands.

wtrocki:
this is already supported by dev preview config. Dev preview commands aren't generated by default. I specifically enabled dev preview before generation to include them. It turns out this was bad call.

PR #823, opened 2021-07-15
feat(kafka topic): partitions flag for update

### LINKED ISSUE #762: Add ability to update the number of partitions for a Kafka topic

It should be possible for a user to update the number of partitions for an existing topic.

### Proposed solution

Add a flag to the existing `kafka topic update` command to specify the number of partitions.

### Additional context

The number of partitions cannot be decreased below the current number.

rkpattnaik780:
User should be able to update number of partitions in a topic. The number of partitions can only increase.

Closes #762 

### Verification Steps

1. Create a Kafka topic and update its partition count using `--partitions` flag.
```
./rhoas kafka topic update rama-topic --partitions 5
```
2. Partitions can only increase, trying to reduce number of partitions should throw error.
3. Validate the interactive update flow. Try providing values smaller and bigger than current partitions count.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/kafka/topic/update/update.go
+++ b/pkg/cmd/kafka/topic/update/update.go
@@ -172,6 +172,7 @@ func NewUpdateTopicCommand(f *factory.Factory) *cobra.Command {
 	cmd.Flags().StringVar(&opts.retentionMsStr, "retention-ms", "", opts.localizer.MustLocalize("kafka.topic.common.input.retentionMs.description"))
 	cmd.Flags().StringVar(&opts.retentionBytesStr, "retention-bytes", "", opts.localizer.MustLocalize("kafka.topic.common.input.retentionBytes.description"))
 	cmd.Flags().StringVar(&opts.cleanupPolicy, "cleanup-policy", "", opts.localizer.MustLocalize("kafka.topic.common.input.cleanupPolicy.description"))
+	cmd.Flags().StringVar(&opts.partitionsStr, "partitions", "", opts.localizer.MustLocalize("kafka.topic.common.input.partitions.description"))
 
 	flagutil.EnableOutputFlagCompletion(cmd)
 
@@ -202,6 +203,14 @@ func runCmd(opts *Options) error {
 				return err
 			}
 		}
+
+		if opts.partitionsStr != "" {
+			partitionCount, err = topicutil.ConvertPartitionsToInt(opts.partitionsStr)
+			if err != nil {
+				return err
+			}
+		}
+
 	}
 
 	conn, err := opts.Connection(connection.DefaultConfigRequireMasAuth)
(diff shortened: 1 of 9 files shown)


rkpattnaik780:
While updating non-interactive mode for updating partition topic, this is the displayed error:
Screenshot from 2021-07-15 14-24-36
Should we set the error message in `case statuscode.badrequest`? But again a bad request can mean many things and not only invalid partitions.

wtrocki:
> Should we set the error message in case statuscode.badrequest? But again a bad request can mean many things and not only invalid partitions.

That would be typically only status code. We need backend to return error codes like kafka and map them properly.  If we not getting any info back that can be used to know type of errror I would recomend to:

- Create issue in kafka backend to discuss it
- create issue in cli to adjust error messages/handling when backend support it.

pkg/cmd/kafka/topic/update/update.go
```diff
@@ -202,6 +203,14 @@ func runCmd(opts *Options) error {
 				return err
 			}
 		}
+
+		if opts.partitionsStr != "" {
+			partitionCount, err = topicutil.ConvertPartitionsToInt(opts.partitionsStr)
+			if err != nil {
+				return err
```
wtrocki, line 210: This could be some generic error from system that is not i18n
Best to return explicit error.

pkg/cmd/kafka/topic/update/update.go
```diff
@@ -202,6 +203,14 @@ func runCmd(opts *Options) error {
 				return err
 			}
 		}
+
+		if opts.partitionsStr != "" {
```
wtrocki, line 207: Partition count is used only in of bellow. 
Best to move it into if scope or closer to where it is used

pkg/kafka/topic/validators.go
```diff
@@ -81,6 +86,10 @@ func (v *Validator) ValidatePartitionsN(val interface{}) error {
 		return errors.New(v.Localizer.MustLocalize("kafka.topic.common.validation.partitions.error.invalid.minValue", localize.NewEntry("Partitions", partitions), localize.NewEntry("Min", minPartitions)))
 	}
 
+	if partitions < v.CurPartitions {
```
wtrocki, line 89: Equality also might not make sense? 
Cna we do <= ?

wtrocki (approved):
I have no means to verify but all changes look good

pkg/kafka/topic/validators.go
rkpattnaik780: Making a patch with the same number of partitions doesn't throw any errors from server, hence I resorted to not using equality here, wdyt?

pkg/cmd/kafka/topic/update/update.go
rkpattnaik780: Regarding this and comment above, it aligns with how we handle validations for other flags.

pkg/cmd/kafka/topic/update/update.go
rkpattnaik780: The validators return i18n error messages.

pkg/kafka/topic/validators.go
wtrocki: ok. no issue here

Issue #828, opened 2021-07-19
CLI not working on Windows 10


EricWittmann:
## Description

**Version**:  0.26.0-alpha5

Getting an error trying to run the CLI on Windows 10:  "open locales\en: file does not exist"

### Steps to reproduce

1. Download CLI for Windows
2. Run CLI, get error

![image](https://user-images.githubusercontent.com/1890703/126193708-4f47f6f4-64af-4259-9751-9f13723c095d.png)

EricWittmann:
Also tested with 0.25.0 with the same result.

wtrocki:
@rkpattnaik780 Do you have windows machine?

We should enable tests and build on windows for github actions in long run

rkpattnaik780:
> @rkpattnaik780 Do you have windows machine?
Nope I have only linux here.
> We should enable tests and build on windows for github actions in long run
Yes, that will be great to have.

wtrocki:
Looks like path problem:
https://github.com/redhat-developer/app-services-cli/blob/main/pkg/localize/goi18n/go_i18n.go#L47

We can fix it and verify in CI/CD

wtrocki:
Replicated problem and identified issue. On PTO so looking if anyone else could make change.

CC @pmuir

PR #829, opened 2021-07-19
ci: build binary and run tests on Windows

wtrocki:


wtrocki:
Enable CI/CD for tracking issues on windows.
https://github.com/redhat-developer/app-services-cli/issues/828

wtrocki:
Problem replicated. It appears to be inside the library we use

wtrocki:
Problem is with embed.
Even on windows path should be forward slash https://pkg.go.dev/embed

CC @rkpattnaik780 @ankithans

wtrocki:
we should add fmt print to strings loaded by embed.
If that works then we have to log issue with lib

rkpattnaik780 (approved):

.github/workflows/gobuild.yml
```diff
@@ -3,33 +3,23 @@ name: Go Build
 on:
   push:
     branches: [main]
-    paths:
-      - '**.go'
-      - 'go.mod'
```
craicoverflow, line 8: Why was this removed?

craicoverflow: I'd rather keep filters in, because this build will run on every change when it does not need to - for example, if I change the Makefile then this will change.

wtrocki: This PR was used for testing. Changes here are experimental and allowing us to replicate/check if we fixed problem.

craicoverflow:
Obsolete.

PR #845, opened 2021-07-27
feat: rename debug flag to verbose

### LINKED ISSUE #842: Introduce verbose mode that is printed on every error
## What

Rename debug to verbose flag for better alignment with other CLI's
Print verbose flag on errors (especially API errors). For example:
`To view more details use --verbose flag`
Verbose flag will also check for cli updates.

## Discussion from another issue

If we document that if something wrong happens with CLI use verbose mode. 
That will give us:

1) Exact and fully error from backend which should return you information that your API call is outdated
2) Check if the version is up to date and suggest new one
3) When error happens without `--verbose` we can have generic message that suggest user to use --verbose flag for more details.
4) user gets info when it is needed in context (not in 4 hours later when timer kicks in etc.)

This approach is what I have seen in other CLI's 

We do not need to change debug to verbose if we print message (point nr3) so there is no big drawbacks.

_Originally posted by @wtrocki in https://github.com/redhat-developer/app-services-cli/issues/841#issuecomment-886567606_

craicoverflow:
BREAKING: The `debug` flag is now `verbose`. There is also no longer any shorthand for `version`.

Closes #842 

### Verification Steps

Run `rhoas kafka list -v` and it will enable verbose mode.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed


pkg/cmd/debug/flag.go
```diff
@@ -4,21 +4,21 @@ package debug
 
 import "github.com/spf13/pflag"
 
-// AddFlag adds the debug flag to the given set of command line flags.
+// AddFlag adds the veerbose flag to the given set of command line flags.
```
rkpattnaik780, line 7:
```suggestion
// AddFlag adds the verbose flag to the given set of command line flags.
```

cmd/rhoas/main.go
```diff
@@ -57,24 +52,16 @@ func main() {
...
-			_ = dump.JSON(cmdFactory.IOStreams.ErrOut, errJSON)
+			build.CheckForUpdate(context.Background(), logger, localizer)
 		}
-		os.Exit(1)
+		return
 	}
 
-	if err = cmdutil.CheckSurveyError(err); err != nil {
-		logger.Error("Error:", err)
+	if err != nil {
```
wtrocki, line 62: This brings patter that our commands should no longer log message when error happens but return error that is exact message of what have happened.

wtrocki (approved):

wtrocki:
@craicoverflow - I have been told that for breaking changes that *can* affect docs it is important to have jira and ping docs team so they can verify if missing flag is not affecting them. I have done quick check and we should be ok.

cmd/rhoas/main.go
craicoverflow: None of our commands log the error, they all already return the error and let the main program handle it.

craicoverflow:
> @craicoverflow - I have been told that for breaking changes that _can_ affect docs it is important to have jira and ping docs team so they can verify if the missing flag is not affecting them. I have done a quick check on the guides repo and we should be ok so leaving decision for you.

Okay, I was unaware. I will create a JIRA.

rkpattnaik780 (approved):

PR #848, opened 2021-07-28
fix: update version of the kafka admin SDK

wtrocki:
New API splits for topics/groups/acls.
I did not add ACL API yet but have concerns how to do update CLI.

We can either do this:

Screenshot 2021-07-28 at 15 46 09
Or this:
Screenshot 2021-07-28 at 15 45 37

Opinions? FIrst approach is implemented in PR

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -11,7 +11,7 @@ import (
 type API struct {
 	Kafka          func() kafkamgmtclient.DefaultApi
 	ServiceAccount func() kafkamgmtclient.SecurityApi
-	KafkaAdmin     func(kafkaID string) (kafkainstanceclient.DefaultApi, *kafkamgmtclient.KafkaRequest, error)
+	KafkaAdmin     func(kafkaID string) (*kafkainstanceclient.APIClient, *kafkamgmtclient.KafkaRequest, error)
 	AccountMgmt    func() amsclient.DefaultApi
 
 	ServiceRegistryMgmt func() srsmgmtclient.RegistriesApi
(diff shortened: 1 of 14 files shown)


pkg/api/api.go
```diff
@@ -11,7 +11,7 @@ import (
 type API struct {
 	Kafka          func() kafkamgmtclient.DefaultApi
 	ServiceAccount func() kafkamgmtclient.SecurityApi
-	KafkaAdmin     func(kafkaID string) (kafkainstanceclient.DefaultApi, *kafkamgmtclient.KafkaRequest, error)
+	KafkaAdmin     func(kafkaID string) (kafkainstanceclient.TopicsApi, kafkainstanceclient.GroupsApi, *kafkamgmtclient.KafkaRequest, error)
```
craicoverflow, line 14:
```suggestion
	KafkaAdmin     func(kafkaID string) (kafkainstanceclient, *kafkamgmtclient.KafkaRequest, error)
```

If we return the full API client like this, it can be used like: `_ = api.TopicsApi.GetTopic()`

craicoverflow (commented):
I'd like to see us return the full Kafka Admin client as suggested rather than returning multiple API groups

pkg/api/api.go
wtrocki: Yes. This requires refactor of the KeycloakConnection so it will take some time

wtrocki:
@craicoverflow kafkainstanceclient is namespace. There is no common group for those unless we actually fix that in SDK by creating it.
Screenshot 2021-07-28 at 16 00 32

wtrocki:
Actually SDK returns: kafkainstanceclient.APIClient and that should be used, however this conflicts with unit testing capabilities and token refresh

wtrocki:
WDYT now?

craicoverflow (approved):

PR #854, opened 2021-07-29
docs: add official logo to README

craicoverflow:


README.md
```diff
@@ -1,6 +1,9 @@
 # RHOAS CLI
+<p align="center">
+  <img alt="Logo" src="https://user-images.githubusercontent.com/11743717/127519981-97c76ae4-f17b-4ac8-8b4d-365bfa4a6374.png">
+</p>
 
-Repository for the RHOAS CLI.
+`rhoas` is a command-line client for managing all of your managed services on Kubernetes.
```
craicoverflow, line 6: Not sure about this. I want to be more descriptive than "Repository for the RHOAS CLI." while also remaining neutral in terms of upstream versus product language.

wtrocki: `managed` was replaced by `application`

README.md
```diff
@@ -1,6 +1,9 @@
 # RHOAS CLI
+<p align="center">
```
wtrocki, line 2: That works but was replaced by css that looks like this.
No strong opinion to change it but it simply looks better :D
```suggestion
<p style="width:400px; text-align:center;">
```

wtrocki: Actually entire container should be centered and p tag can be omitted

craicoverflow: What would be the functional benefit of changing this?

README.md
```diff
@@ -1,6 +1,9 @@
...
+  <img alt="Logo" src="https://user-images.githubusercontent.com/11743717/127519981-97c76ae4-f17b-4ac8-8b4d-365bfa4a6374.png">
```
wtrocki, line 3: Is this ID coming from GitHub profile page?

In situations when proxy is used github content might take long time to load. Anyway we can load content locally.

Since logo is done in svg I think the best to put it inline as base64 image+svg

wtrocki (approved):
Sorry for so many comments on trivial PR.
Approving considering that changes pointed out are for your consideration only.

README.md
wtrocki: controling with of the content depending on the size of the screen - so image is not stretched on 16:9 screens etc.

wtrocki:
@craicoverflow my bad. Ignore me. I got confused by 2 PR's 
I have added comments that should go to login page to readme. 

My bad. This PR is good to go! Sorry

README.md
wtrocki: Igore me. This is readme. Doesn't matter much

PR #859, opened 2021-07-30
Apicurio CLI 

wtrocki:
## Description

Set of commands to interact with service registry dataplane:
```
Available Commands:
  create      Creates new artifact from file or standard input
  delete      Deletes all of the artifacts that exist in a given group
  download    Download artifacts from registry by using global identifiers
  get         Get artifact by id and group
  list        List artifacts
  metadata    Get artifact metadata
  update      Update artifact
  versions    Get latest artifact versions by id and group
```

# Verification

## Create artifact in my-group from schema.json file
rhoas service-registry artifact create --artifact=my-artifact --group=my-group artifact.json

## Get artifact content
rhoas service-registry artifact get --artifact=my-artifact --group=my-group file.json 

## Delete artifact
rhoas service-registry artifact delete --artifact=my-artifact

## Get artifact metadata
rhoas service-registry artifact metadata --artifact=my-artifact --group=my-group

## Update artifact
rhoas service-registry artifact update --artifact=my-artifact artifact-new.json

## List Artifacts
rhoas service-registry artifact list --group=my-group --limit=10 page=1

## View artifact versions
rhoas service-registry artifact versions --artifact=my-artifact --group=my-group

### DIFF
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -128,7 +128,7 @@ rhoas service-registry artifact create --type=JSON my-artifact.json
 	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
 	cmd.Flags().StringVarP(&opts.file, "file", "f", "", "File location of the artifact")
 
-	cmd.Flags().StringVarP(&opts.artifact, "artifact", "a", "", "Id of the artifact")
+	cmd.Flags().StringVarP(&opts.artifact, "artifact-id", "a", "", "Id of the artifact")
 	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Group of the artifact")
 	cmd.Flags().StringVarP(&opts.artifactType, "type", "t", "", "Type of artifact. Choose from:  "+util.GetAllowedArtifactTypeEnumValuesAsString())
 	cmd.Flags().StringVarP(&opts.registryID, "instance-id", "", "", "Id of the registry to be used. By default uses currently selected registry")
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -159,7 +159,7 @@ func runCreate(opts *Options) error {
 	}
 
 	if opts.group == util.DefaultArtifactGroup {
-		logger.Info("Group was not specified. Using ", util.DefaultArtifactGroup, " artifacts group.")
+		logger.Info("Group was not specified. Using", util.DefaultArtifactGroup, "artifacts group.")
 		opts.group = util.DefaultArtifactGroup
 	}
 
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -87,14 +87,11 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 		Short: "Creates new artifact from file or standard input",
 		Long:  longDescription,
 		Example: `
-## Create artifact in my-group from schema.json file
-rhoas service-registry artifacts create my-group schema.json
-
 # Create an artifact in default group
-rhoas service-registry artifacts create my-artifact.json
+rhoas service-registry artifact create my-artifact.json
 
 # Create an artifact with specified type
-rhoas service-registry artifacts create --type=JSON my-artifact.json
+rhoas service-registry artifact create --type=JSON my-artifact.json
 		`,
 		Args: cobra.RangeArgs(0, 1),
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -2,18 +2,15 @@ package create
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
 
-	"github.com/redhat-developer/app-services-cli/pkg/cmdutil"
 	"github.com/redhat-developer/app-services-cli/pkg/connection"
 	"github.com/redhat-developer/app-services-cli/pkg/dump"
 	"github.com/redhat-developer/app-services-cli/pkg/localize"
 	"github.com/redhat-developer/app-services-cli/pkg/serviceregistry/registryinstanceerror"
 	registryinstanceclient "github.com/redhat-developer/app-services-sdk-go/registryinstance/apiv1internal/client"
-	"gopkg.in/yaml.v2"
 
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/flag"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/registry/artifact/util"
@@ -197,14 +194,7 @@ func runCreate(opts *Options) error {
 	}
 	logger.Info("Artifact created")
 
-	switch opts.outputFormat {
-	case "json":
-		data, _ := json.MarshalIndent(metadata, "", cmdutil.DefaultJSONIndent)
-		_ = dump.JSON(opts.IO.Out, data)
-	case "yaml", "yml":
-		data, _ := yaml.Marshal(metadata)
-		_ = dump.YAML(opts.IO.Out, data)
-	}
+	dump.PrintDataInFormat(opts.outputFormat, metadata, opts.IO.Out)
 
 	return nil
 }
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -129,7 +129,7 @@ rhoas service-registry artifact create --type=JSON my-artifact.json
 	cmd.Flags().StringVarP(&opts.file, "file", "f", "", "File location of the artifact")
 
 	cmd.Flags().StringVarP(&opts.artifact, "artifact-id", "a", "", "Id of the artifact")
-	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Group of the artifact")
+	cmd.Flags().StringVarP(&opts.group, "group", "g", util.DefaultArtifactGroup, "Group of the artifact")
 	cmd.Flags().StringVarP(&opts.artifactType, "type", "t", "", "Type of artifact. Choose from:  "+util.GetAllowedArtifactTypeEnumValuesAsString())
 	cmd.Flags().StringVarP(&opts.registryID, "instance-id", "", "", "Id of the registry to be used. By default uses currently selected registry")
 
@@ -158,8 +158,8 @@ func runCreate(opts *Options) error {
 		return err
 	}
 
-	if opts.group == "" {
-		logger.Info("Group was not specified. Using " + util.DefaultArtifactGroup + " artifacts group.")
+	if opts.group == util.DefaultArtifactGroup {
+		logger.Info("Group was not specified. Using ", util.DefaultArtifactGroup, " artifacts group.")
 		opts.group = util.DefaultArtifactGroup
 	}
 
@@ -171,7 +171,7 @@ func runCreate(opts *Options) error {
 			return err
 		}
 	} else {
-		logger.Info("Reading file content from stdin")
+		logger.Info("Reading file content from standard input")
 		specifiedFile, err = util.CreateFileFromStdin()
 		if err != nil {
 			return err
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -112,12 +112,12 @@ rhoas service-registry artifact create --type=JSON my-artifact.json
 
 			if opts.artifactType != "" {
 				if _, err = registryinstanceclient.NewArtifactTypeFromValue(opts.artifactType); err != nil {
-					return errors.New("Invalid artifact type. Please use one of following values: " + util.GetAllowedArtifactTypeEnumValuesAsString())
+					return errors.New("invalid artifact type. Please use one of following values: " + util.GetAllowedArtifactTypeEnumValuesAsString())
 				}
 			}
 
 			if !cfg.HasServiceRegistry() {
-				return errors.New("No service Registry selected. Use 'rhoas service-registry use' use to select your registry")
+				return errors.New("no service Registry selected. Use 'rhoas service-registry use' use to select your registry")
 			}
 
 			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -1,0 +1,213 @@
+package create
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/redhat-developer/app-services-cli/pkg/cmdutil"
+	"github.com/redhat-developer/app-services-cli/pkg/connection"
+	"github.com/redhat-developer/app-services-cli/pkg/dump"
+	"github.com/redhat-developer/app-services-cli/pkg/localize"
+	"github.com/redhat-developer/app-services-cli/pkg/serviceregistry/registryinstanceerror"
+	registryinstanceclient "github.com/redhat-developer/app-services-sdk-go/registryinstance/apiv1internal/client"
+	"gopkg.in/yaml.v2"
+
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/flag"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/registry/artifact/util"
+	flagutil "github.com/redhat-developer/app-services-cli/pkg/cmdutil/flags"
+
+	"github.com/redhat-developer/app-services-cli/pkg/iostreams"
+
+	"github.com/redhat-developer/app-services-cli/pkg/logging"
+
+	"github.com/spf13/cobra"
+
+	"github.com/redhat-developer/app-services-cli/internal/config"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/factory"
+)
+
+type Options struct {
+	artifact string
+	group    string
+
+	file         string
+	artifactType string
+
+	registryID   string
+	outputFormat string
+
+	IO         *iostreams.IOStreams
+	Config     config.IConfig
+	Connection factory.ConnectionFunc
+	Logger     func() (logging.Logger, error)
+	localizer  localize.Localizer
+}
+
+var longDescription = `
+Creates a new artifact by posting the artifact content to the registry.
+
+Artifacts can be typically in JSON format for most of the supported types, but may be in another format for a few (for example, PROTOBUF).
+The registry attempts to figure out what kind of artifact is being added from the following supported list:
+
+- Avro (AVRO)
+- Protobuf (PROTOBUF)
+- JSON Schema (JSON)
+- Kafka Connect (KCONNECT)
+- OpenAPI (OPENAPI)
+- AsyncAPI (ASYNCAPI)
+- GraphQL (GRAPHQL)
+- Web Services Description Language (WSDL)
+- XML Schema (XSD)
+
+An artifact is created using the content provided in the body of the request.  
+This content is created under a unique artifact ID that can be provided by user.
+If not provided in the request, the server generates a unique ID for the artifact. 
+It is typically recommended that callers provide the ID, because this is a meaningful identifier, and for most use cases should be supplied by the caller.
+If an artifact with the provided artifact ID already exists command will fail with error.
+
+
+When --group parameter is missing the command will create a new artifact under the "default" group.
+when --registry is missing the command will create a new artifact for currently active service registry (visible in rhoas service-registry describe)
+`
+
+func NewCreateCommand(f *factory.Factory) *cobra.Command {
+	opts := &Options{
+		IO:         f.IOStreams,
+		Config:     f.Config,
+		Connection: f.Connection,
+		Logger:     f.Logger,
+		localizer:  f.Localizer,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Creates new artifact from file or standard input",
+		Long:  longDescription,
+		Example: `
+## Create artifact in my-group from schema.json file
+rhoas service-registry artifacts create my-group schema.json
+
+# Create an artifact in default group
+rhoas service-registry artifacts create my-artifact.json
+
+# Create an artifact with specified type
+rhoas service-registry artifacts create --type=JSON my-artifact.json
+		`,
+		Args: cobra.RangeArgs(0, 1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			validOutputFormats := flagutil.ValidOutputFormats
+			if opts.outputFormat != "" && !flagutil.IsValidInput(opts.outputFormat, validOutputFormats...) {
+				return flag.InvalidValueError("output", opts.outputFormat, validOutputFormats...)
+			}
+
+			if len(args) > 0 {
+				opts.file = args[0]
+			}
+
+			if opts.registryID != "" {
+				return runCreate(opts)
+			}
+
+			cfg, err := opts.Config.Load()
+			if err != nil {
+				return err
+			}
+
+			if opts.artifactType != "" {
+				if _, err = registryinstanceclient.NewArtifactTypeFromValue(opts.artifactType); err != nil {
+					return errors.New("Invalid artifact type. Please use one of following values: " + util.GetAllowedArtifactTypeEnumValuesAsString())
+				}
+			}
+
+			if !cfg.HasServiceRegistry() {
+				return errors.New("No service Registry selected. Use 'rhoas service-registry use' use to select your registry")
+			}
+
+			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
+			return runCreate(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
+	cmd.Flags().StringVarP(&opts.file, "file", "f", "", "File location of the artifact")
+
+	cmd.Flags().StringVarP(&opts.artifact, "artifact", "a", "", "Id of the artifact")
+	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Group of the artifact")
+	cmd.Flags().StringVarP(&opts.artifactType, "type", "t", "", "Type of artifact. Choose from:  "+util.GetAllowedArtifactTypeEnumValuesAsString())
+	cmd.Flags().StringVarP(&opts.registryID, "instance-id", "", "", "Id of the registry to be used. By default uses currently selected registry")
+
+	flagutil.EnableOutputFlagCompletion(cmd)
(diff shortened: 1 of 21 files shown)


wtrocki:
CC @ankithans

wtrocki:
Looking for early feedback about dataplane crud commands.
More into UX and potential issues than code as I still did not i18n strings

wtrocki:
Requirements chaged for some commands. Rewriting

pkg/cmd/registry/artifacts/crud/get/get.go
```diff
@@ -0,0 +1,164 @@
...
+rhoas service-registry artifacts get --group mygroup --registryId=myregistry --artifact myartifact
+`,
+		Args: cobra.RangeArgs(0, 2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
+				opts.artifact = args[0]
+			}
+
+			if opts.artifact == "" {
+				return fmt.Errorf("Artifact is required. Please specify artifact as positional argument or by using --artifact flag")
```
craicoverflow, line 77: It is probably better to use one or the other (positional or flag). Offering both does not give much functional benefit but adds more documentation.

pkg/cmd/registry/artifacts/download/download.go
```diff
@@ -0,0 +1,170 @@
...
+	opts := &Options{
+		Config:     f.Config,
+		Connection: f.Connection,
+		IO:         f.IOStreams,
+		localizer:  f.Localizer,
+		Logger:     f.Logger,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "get",
```
craicoverflow, line 50: This has the same command name as `get` so it is being overridden I guess?

pkg/cmd/registry/artifacts/download/download.go
```diff
@@ -0,0 +1,170 @@
...
+		logger.Info("Group was not specified. Using 'default' artifacts group.")
+		opts.group = util.DefaultArtifactGroup
+	}
+
+	logger.Info("Fetching artifact")
+
+	ctx := context.Background()
+	var dataFile *os.File
+	// nolint
+	if opts.contentId != -1 {
```
craicoverflow, line 137: What does `-1` mean? We should create a constant for readbility.

wtrocki: Actually rookie mistake. Variables in flag should be *int64 and then we can check for nil as default value

pkg/cmd/registry/artifacts/download/download.go
wtrocki: strangely enough I have changed it locally. Github outage prevented me to update it sorry.

pkg/cmd/registry/artifacts/crud/get/get.go
wtrocki: That is intentional for UX trial.
I do not like our positional flags usage in our current cli. They poorly documented and not isomorphic (some commands take name or id. Some just id and if we mix them we getting 404 with some missleading message.

For registry it is hard to define positionals as there are multiple different Ids, groups etc.

So my take is to see how users will use those commands and then adjust/remove positionals and flags. Can we keep duplicates for now (before adding i18n )?

pkg/cmd/registry/artifacts/crud/list/list.go
```diff
@@ -0,0 +1,198 @@
...
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Artifact group")
+
+	cmd.Flags().Int32VarP(&opts.page, "page", "", 1, "Page number")
+	cmd.Flags().Int32VarP(&opts.limit, "limit", "", 100, "Page limit")
+	cmd.Flags().StringVarP(&opts.orderBy, "orderBy", "", "", "Order by fields (id, name etc.) ")
+
+	cmd.Flags().StringVarP(&opts.registryID, "registryId", "", "", "Id of the registry to be used. By default uses currently selected registry")
```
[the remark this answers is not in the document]
wtrocki: Need clarification. We should change all camelcase to `artifact-id` etc.?

Like charmil rule suggestion

pkg/cmd/registry/artifacts/crud/delete/delete.go
```diff
@@ -0,0 +1,183 @@
...
+		Config:     f.Config,
+		Connection: f.Connection,
+		Logger:     f.Logger,
+		IO:         f.IOStreams,
+		localizer:  f.Localizer,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "delete",
+		Short: "Deletes all of the artifacts that exist in a given group.",
```
rkpattnaik780, line 53: Remove period.

pkg/cmd/registry/artifacts/crud/delete/delete.go
```diff
@@ -0,0 +1,183 @@
...
+		},
+	}
+
+	cmd.Flags().BoolVarP(&opts.force, "yes", "y", false, "Delete without prompt")
+	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
+
+	cmd.Flags().StringVarP(&opts.artifact, "artifact", "a", "", "Id of the artifact")
+	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Group of the artifact")
+	cmd.Flags().StringVarP(&opts.registryID, "registryId", "", "", "Id of the registry to be used. By default uses currently selected registry")
+
```
rkpattnaik780, line 110: Should we add static completion for output flag here?

pkg/cmd/registry/artifacts/metadata/metadata.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
+			return runGet(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.artifact, "artifact", "a", "", "Id of the artifact")
+	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Group of the artifact")
+	cmd.Flags().StringVarP(&opts.registryID, "registryId", "", "", "Id of the registry to be used. By default uses currently selected registry")
```
rkpattnaik780, line 87: Remove camel casing for `registryID` flag

pkg/cmd/registry/artifacts/util/constants.go
```diff
@@ -0,0 +1,25 @@
...
+	"OPENAPI",
+	"ASYNCAPI",
+	"GRAPHQL",
+	"KCONNECT",
+	"WSDL",
+	"XSD",
+	"XML",
+}
+
+// Get AllowedArtifactTypeEnumValues as string
```
rkpattnaik780, line 22:
```suggestion
// GetAllowedArtifactTypeEnumValues gets artifact types as string.
```

pkg/cmd/registry/artifacts/crud/list/list.go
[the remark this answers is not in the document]
craicoverflow: Yes, camelCase should not be used in flags or command names.

pkg/cmd/registry/artifacts/crud/get/get.go
craicoverflow: > They poorly documented and loosely defined (some commands take name or id.
> Positionals have very clear usage in other CLIs but in our CLI they seem little bit overused.

All names using a positional argument. Eg: `rhoas kafka describe [name]`
All IDs use a flag. Eg: `rhoas kafka describe --id [id]`.

I would not say that they are **over**used. They are used very little.

> For registry it is hard to define positionals that would be easy to remember for user as there are multiple different Ids, groups etc. and each command will have different mixture of those.

You should really only use it for a `name` argument, if there is any.

> So my take is to see how users will use those commands and then adjust/remove positionals and flags.

How do you plan to collect usage infrmation?

pkg/cmd/registry/artifacts/download/download.go
craicoverflow: I don't think that is possible.

pkg/cmd/registry/artifacts/metadata/metadata.go
wtrocki: How it should look like?

craicoverflow: Yes, kebab-case.

craicoverflow: Should we call it `instance-id`? We could use this across all instance types..such as Kafka, Connector etc.

wtrocki: Good idea. Changing

pkg/cmd/registry/artifacts/crud/delete/delete.go
wtrocki: Applying

pkg/cmd/registry/artifacts/download/download.go
wtrocki: yes. IT was outdated. Github glitch

pkg/cmd/registry/artifacts/crud/get/get.go
wtrocki: We will send guide to users to try CLI and then have them sent us usage log.
More info:
https://github.com/aerogear/charmil/issues/222

pkg/cmd/registry/artifacts/crud/create/create.go
```diff
@@ -0,0 +1,209 @@
...
+				return fmt.Errorf("No service Registry selected. Use 'rhoas service-registry use' use to select your registry")
+			}
+
+			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
+			return runCreate(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
+	cmd.Flags().StringVarP(&opts.file, "file", "f", "", "File location of the artifact")
```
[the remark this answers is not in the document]
wtrocki: Will remove that later. Created duplicates for UX session

pkg/cmd/registry/artifacts/crud/get/get.go
```diff
@@ -0,0 +1,164 @@
...
+		return err
+	}
+	if opts.outputFile != "" {
+		err := os.WriteFile(opts.outputFile, fileContent, 0600)
+		if err != nil {
+			return err
+		}
+	} else {
+		// Print to stdout
+		fmt.Fprintf(os.Stdout, "%v\n", string(fileContent))
```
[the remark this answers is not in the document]
craicoverflow: Yeah, we probably should not then, unless you add a flag which specifies doing this `--pretty`.

pkg/cmd/registry/artifacts/crud/get/get.go
```diff
@@ -0,0 +1,164 @@
...
+
+			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
+			return runGet(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.artifact, "artifact", "a", "", "Id of the artifact")
+	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Group of the artifact")
+	cmd.Flags().StringVarP(&opts.registryID, "registryId", "", "", "Id of the registry to be used. By default uses currently selected registry")
+	cmd.Flags().StringVarP(&opts.outputFile, "outputFile", "", "", "Filename of the output file")
```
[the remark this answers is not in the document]
craicoverflow: If it was a requirement from BU then that is okay. It seems a little weird to provide so many different ways to do the same thing though. Positional especially seems out of place - I can see a flag being suitable.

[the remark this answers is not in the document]
wtrocki: Ah.. ok. I wanted to give number of options for User trials :D

pkg/cmd/registry/artifacts/crud/get/get.go
```diff
@@ -0,0 +1,164 @@
...
+	opts := &Options{
+		Config:     f.Config,
+		Connection: f.Connection,
+		IO:         f.IOStreams,
+		localizer:  f.Localizer,
+		Logger:     f.Logger,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "get",
```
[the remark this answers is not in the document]
wtrocki: This command doesn't get metadata it gets file. 
There is specific metadata command that is more implicit than describe. I will leave those considerations for user trials.

pkg/cmd/registry/artifacts/crud/get/get.go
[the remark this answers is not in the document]
craicoverflow: I did not have the full context regarding user trials, in that case it is fine to leave it in - but be aware it could make things confusing during the trials.

wtrocki:
Upon consideration, I think it will be good to merge this without i18n and have follow-up issue to address those. 
Without that we would probably have this PR staying for couple weeks unmerged etc.

WDYT?

wtrocki:
https://github.com/redhat-developer/app-services-cli/issues/904

rkpattnaik780:
> WDYT?

We can do that.

wtrocki:
I will create more granular issues to continue work on this and another service registry. This way we can collaborate better.
What we need is to have some proper review of the commands

pkg/dump/dump.go
```diff
@@ -136,3 +136,15 @@ func haveYQ(minVersion int) bool {
 
 	return false
 }
+
+// Dump prints the given data to the given format
```
rkpattnaik780, line 140:
```suggestion
// PrintDataInFormat prints the given data to the given format
```

pkg/cmd/registry/artifact/crud/delete/delete.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	if opts.group == "" {
+		logger.Info("Group was not specified. Using " + util.DefaultArtifactGroup + " artifacts group.")
+		opts.group = util.DefaultArtifactGroup
+	}
+
+	if opts.artifact == "" {
+		logger.Info("Artifact was not specified. Command will delete all artifacts in the group")
+		err = confirmDelete(opts, "Do you want to delete ALL ARTIFACTS from group "+opts.group)
```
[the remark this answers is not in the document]
wtrocki: Adding check for artifact. I cannot see any way to check if group exist (there are no error codes on list command)

pkg/cmd/registry/artifact/crud/get/get.go
```diff
@@ -0,0 +1,165 @@
...
+		logger.Info("Fetching artifact with version: " + opts.version)
+		request := dataAPI.VersionsApi.GetArtifactVersion(ctx, opts.group, opts.artifact, opts.version)
+		dataFile, _, err = request.Execute()
+	} else {
+		logger.Info("Fetching latest artifact")
+		request := dataAPI.ArtifactsApi.GetLatestArtifact(ctx, opts.group, opts.artifact)
+		dataFile, _, err = request.Execute()
+	}
+	if err != nil {
+		return registryinstanceerror.TransformError(err)
```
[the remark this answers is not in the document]
wtrocki: Sadly backend doesn't provide proper error codes thus we have limited ability to map anything. 
What we get right now is 

- message
- Name of the Exception class (that is unlikely to be stable)
- ugly java stacktrace (in dev env)

CC @jsenko do you think that it would be possible to introduce error codes to dataplane in the future?

pkg/cmd/registry/artifact/crud/list/list.go
```diff
@@ -0,0 +1,204 @@
...
+
+			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
+
+			return runList(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Artifact group")
+
+	cmd.Flags().Int32VarP(&opts.page, "page", "", 1, "Page number")
```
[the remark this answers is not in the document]
wtrocki: Yes. Backend should return error in this case. I can hack this on the client for now.

craicoverflow:
```shell
$ rhoas service-registry artifact list --group=my-group
Provided limit and page arguments are larger than total count of elements on the server 1
~/c/gith/redhat-developer/app-services-cli apicurio-cli *21 ❯
```

Getting some weird pagination error.

pkg/cmd/registry/artifact/crud/list/list.go
```diff
@@ -0,0 +1,210 @@
...
+	response, _, err := request.Execute()
+	if err != nil {
+		return registryinstanceerror.TransformError(err)
+	}
+
+	totalCount := opts.page * opts.limit
+	if len(response.Artifacts) != 0 && response.GetCount() < totalCount {
+		logger.Info("Provided limit and page arguments are larger than total count of elements on the server", response.GetCount())
+		return nil
+	}
```
craicoverflow, line 169: I'm not entirely sure why this is needed?

page=1
limit=10

Therefore you will get the first 10 possible items.

totalCount = 10

if the number of artifacts is less than 10, then log this message and return nil?

pkg/cmd/registry/artifact/crud/list/list.go
```diff
@@ -0,0 +1,210 @@
...
+	default:
+		rows := mapResponseItemsToRows(&response.Artifacts)
+		dump.Table(opts.IO.Out, rows)
+		logger.Info("")
+	}
+
+	return nil
+}
+
+func mapResponseItemsToRows(artifacts *[]registryinstanceclient.SearchedArtifact) []artifactRow {
```
craicoverflow, line 192:
```suggestion
func mapResponseItemsToRows(artifacts []registryinstanceclient.SearchedArtifact) []artifactRow {
```

A slice does not store any data, it just describes a section of an underlying array. Therefore a pointer is not needed.

pkg/cmd/registry/artifact/crud/list/list.go
```diff
@@ -0,0 +1,210 @@
...
+	rows := []artifactRow{}
+
+	for i := range *artifacts {
```
craicoverflow, line 195:
```suggestion
	for _, artifact := range *artifacts {
```

This gives you the artifact object.

pkg/cmd/registry/artifact/crud/list/list.go
wtrocki: Actually we can remove it

pkg/cmd/registry/artifact/crud/list/list.go
```diff
@@ -0,0 +1,210 @@
...
+		rows := mapResponseItemsToRows(&response.Artifacts)
+		dump.Table(opts.IO.Out, rows)
+		logger.Info("")
+	}
+
+	return nil
+}
+
+func mapResponseItemsToRows(artifacts *[]registryinstanceclient.SearchedArtifact) []artifactRow {
+	rows := []artifactRow{}
```
craicoverflow, line 193:
```suggestion
	var rows []artifactRow
```

This will create the same value - in Go it is generally preferred to create variables which have not been given a value using `var`.

As an aside, this slice does not have a fixed value - but since we know what the size will be we could set a fixed size.

See https://github.com/redhat-developer/app-services-cli/blob/reset_offset/pkg/cmd/kafka/consumergroup/resetoffset/reset_offset.go#L241

This will have no noticeable impact on performance for the CLI, so i's totally optional, just letting you know that it is possible.

pkg/cmd/registry/artifact/crud/create/create.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+			if opts.artifactType != "" {
+				if _, err = registryinstanceclient.NewArtifactTypeFromValue(opts.artifactType); err != nil {
+					return errors.New("Invalid artifact type. Please use one of following values: " + util.GetAllowedArtifactTypeEnumValuesAsString())
+				}
+			}
+
+			if !cfg.HasServiceRegistry() {
+				return errors.New("No service Registry selected. Use 'rhoas service-registry use' use to select your registry")
```
craicoverflow, line 120: **me thinking** When this is running in the independent upstream CLI, what will the error message say, because here it directly refers to `rhoas` but that would not always be so.

pkg/cmd/registry/artifact/crud/create/create.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+
+			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
```
craicoverflow, line 123: `strconv.FormatInt()` would be marginally better here (it is faster).

pkg/cmd/registry/artifact/crud/delete/delete.go
```diff
@@ -0,0 +1,187 @@
...
+		confirm := &survey.Confirm{
+			Message: message,
+		}
+		err := survey.AskOne(confirm, &shouldContinue)
+		if err != nil {
+			return err
+		}
+
+		if !shouldContinue {
+			return errors.New("")
```
craicoverflow, line 183: Should this be an error with no message?

pkg/cmd/registry/artifact/crud/list/list.go
```diff
@@ -0,0 +1,210 @@
...
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Artifact group")
+
+	cmd.Flags().Int32VarP(&opts.page, "page", "", 1, "Page number")
+	cmd.Flags().Int32VarP(&opts.limit, "limit", "", 100, "Page limit")
+
+	cmd.Flags().StringVarP(&opts.registryID, "instance-id", "", "", "Id of the registry to be used. By default uses currently selected registry")
+	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "", "Output format (json, yaml, yml, table)")
```
craicoverflow, line 124:
```suggestion
	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "", "Output format (json, yaml, yml)")
```

Table is not an option.

pkg/cmd/registry/artifact/crud/list/list.go
```diff
@@ -0,0 +1,210 @@
...
+				return errors.New("No service Registry selected. Use rhoas registry use to select your registry")
+			}
+
+			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
+
+			return runList(opts)
+		},
+	}
+
+	cmd.Flags().StringVarP(&opts.group, "group", "g", "", "Artifact group")
```
craicoverflow, line 118: Could you set the default group here?

pkg/cmd/registry/artifact/download/download.go
```diff
@@ -0,0 +1,173 @@
...
+		request := dataAPI.ArtifactsApi.GetContentById(ctx, opts.contentId)
+		dataFile, _, err = request.Execute()
+	} else if opts.globalId != -1 {
+		request := dataAPI.ArtifactsApi.GetContentByGlobalId(ctx, opts.globalId)
+		dataFile, _, err = request.Execute()
+	} else if opts.hash != "" {
+		request := dataAPI.ArtifactsApi.GetContentByHash(ctx, opts.hash)
+		dataFile, _, err = request.Execute()
+	} else {
+		return errors.New("Please specify at least one flag: [contentId, globalId, hash]")
```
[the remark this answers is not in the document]
craicoverflow: Errors should not begin with a capital letter as they will all be prefixed with "Error: ..."

pkg/cmd/registry/artifact/util/constants.go
```diff
@@ -0,0 +1,25 @@
+package util
+
+import (
+	"strings"
+)
+
+var DefaultArtifactGroup = "default"
```
craicoverflow, line 7:
```suggestion
const DefaultArtifactGroup = "default"
```

pkg/cmd/registry/artifact/crud/list/list.go
wtrocki: Yes, but do you think we should not print that default group is used. I'm assigning group in the business logic instead. 
I can check if group is default and print it so no difference.

pkg/cmd/registry/artifact/crud/delete/delete.go
wtrocki: Actually better to return nil

wtrocki:
```suggestion
			return errors.New("Command stopped")
```

pkg/cmd/registry/artifact/crud/list/list.go
wtrocki: Similariy to others. This is copy paste from kafka list so created issue.

pkg/cmd/registry/artifact/crud/create/create.go
wtrocki: Good point!

wtrocki:
Awesome reviews. I will refine things little bit and ping you next week for final check

pkg/cmd/registry/artifact/crud/delete/delete.go
```diff
@@ -0,0 +1,187 @@
...
+
+			opts.registryID = fmt.Sprint(cfg.Services.ServiceRegistry.InstanceID)
+			return runDelete(opts)
+		},
+	}
+
+	cmd.Flags().BoolVarP(&opts.force, "yes", "y", false, "Delete without prompt")
+	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
+
+	cmd.Flags().StringVarP(&opts.artifact, "artifact", "a", "", "Id of the artifact")
```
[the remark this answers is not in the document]
wtrocki: I will change it

pkg/cmd/registry/artifact/crud/list/list.go
wtrocki: This gives me lint errors for operating on objects.

pkg/cmd/registry/artifact/crud/delete/delete.go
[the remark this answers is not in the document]
craicoverflow: `artifact`is vague - do I pass the artifact name, ID, path to file or who object?

pkg/cmd/registry/artifact/crud/create/create.go
craicoverflow: Special error types would be perhaps be needed which are scoped to the data plane CLI.

From data plane CLI, keep it simple with no reference to what command you should use:

```go
var NoRegistrySelectedError = errors.New("no service registry selected")

return NoRegistrySelectedError
```

When interpreting the error, add additional info:

```go
if rerr, ok := err.(*NoRegistrySelectedError); ok {
  return fmt.Errorf("%w: use 'rhoas service-registry use' to select a registry or pass the `--registry-id` flag")
}
```

That is a simplistic approach which does not scale very well, but perhaps some kind of abstraction would be helpful to match errors.

pkg/cmd/registry/artifact/crud/delete/delete.go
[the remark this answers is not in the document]
wtrocki: Agreed. Will change this all. Original reason was that artifact-id is long to type

pkg/cmd/registry/artifact/crud/create/create.go
wtrocki: Good idea!

wtrocki: Created dedicated issue as this is not instantly actionable in PR. I will discuss that with the team ( CC @ankithans )

PR #863, opened 2021-08-04
chore: auto-add "needs-triage" label to new issues

craicoverflow:
This PR uses https://github.com/marketplace/actions/regex-issue-labeler to add a "needs-triage" label to all new issues.


wtrocki:
Let's test it . Action seeems ok

PR #890, opened 2021-08-11
refactor(dump): create constants for output formats

craicoverflow:
This PR adds a constant for each of the output formats (json, yaml, yml). Everything should remain the same functionally.

### Verification Steps

Run `rhoas kafka describe --output (json|yaml|yml)` and it should print the format that is expected.

### Type of change

- [ ] Documentation change
- [ ] Refactor

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/kafka/consumergroup/describe/describe.go
+++ b/pkg/cmd/kafka/consumergroup/describe/describe.go
@@ -140,10 +140,10 @@ func runCmd(opts *Options) error {
 
 	stdout := opts.IO.Out
 	switch opts.outputFormat {
-	case "json":
+	case dump.JSONFormat:
 		data, _ := json.Marshal(consumerGroupData)
 		_ = dump.JSON(stdout, data)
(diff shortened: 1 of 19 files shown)


wtrocki:
While this is good. Entire dump should be probably helper with generic interface (as there is nothing specific) 
Then constants are less likely to be needed (but still can)

wtrocki:
With future CLI it will be easier to have output flags as a form of SDK. Maybe we can even have it as charmil helper?

wtrocki (approved):
Verified, LGTM - suggested more comprehensive approach

rkpattnaik780 (approved):

craicoverflow:
> While this is good. The entire dump should be probably helper with the generic interface (as there is nothing specific)
> Then constants are less likely to be needed (but still can).

Hmm.."as there is nothing specific" - there would still need to be some form of constant to indicate the output format though right? Even if this was made into some generic interface you still need to specify the output format..

PR #908, opened 2021-08-12
fix!: replace config command with dev env

wtrocki:
Simple and clear implementation.
The bad news is that I could not get completion working


CONTRIBUTING.md
```diff
@@ -22,6 +22,15 @@ You can run the CLI commands files directly with `go run`. All commands and subc
 go run ./cmd/rhoas kafka create "test" --provider=aws --region=us-east-1
 ```
 
+### Development features
+
+To enable development and early access commands please set environment variable in your terminal:
+
+```bash
+RHOAS_DEV=true
```
craicoverflow, line 30:
```suggestion
export RHOAS_DEV=true
```

The help text will appear and completions will work only when RHOAS_DEV is exported.

pkg/profile/profile.go
```diff
@@ -2,60 +2,29 @@
...
-}
+var DevPreviewEnv string = "RHOAS_DEV"
 
-// Annotation used in templates and tools like documentation generation
-func DevPreviewAnnotation() map[string]string {
-	return map[string]string{"channel": "preview"}
-}
+// ApplyDevPreviewLabel adds visual element displayed in help
+func ApplyDevPreviewLabel(cmd *cobra.Command) {
+	cmd.Short = "[preview] " + cmd.Short
```
craicoverflow, line 14: Maybe we could highlight this in a different colour?

wtrocki: I had that done. That will break ascidocs generation sadly and I could not figure out how to not include colours into docs

wtrocki:
That PR  will lands into the dataplane branch. I would keep it open and merge the other one first (to not add more elements to review in the source branch.

cben:
https://github.com/redhat-developer/app-services-cli/blob/main/docs/commands/rhoas_config.adoc file still remains, guess should be deleted.

wtrocki:
@cben Amazing catch! Do you mind creating PR to remove it?

PR #935, opened 2021-08-18
fix: add -f flag shortcut to output file

wtrocki:


pkg/cmd/registry/artifact/download/download.go
```diff
@@ -103,7 +103,7 @@ rhoas service-registry artifact download --hash=c71d239df91726fc519c6eb72d318ec6
 	cmd.Flags().Int64VarP(&opts.globalId, "global-id", "", unusedFlagIdValue, "Global ID of the artifact")
 	cmd.Flags().Int64VarP(&opts.contentId, "content-id", "", unusedFlagIdValue, "ContentId of the artifact")
 
-	cmd.Flags().StringVar(&opts.outputFile, "output-file", "", "Filename of the output file")
+	cmd.Flags().StringVarP(&opts.outputFile, "output-file", "f", "", "Filename of the output file")
```
craicoverflow, line 106: It might be confusing that the shorthand does not begin with the same letter as the longhand.

wtrocki: I had the same feeling. Do you think we should stick with `--file`?

craicoverflow: Does it need to have a shorthand?

wtrocki: I do not have prefference but we received feedback that it should have it :D

craicoverflow: The feedback said "should it have", not "it should have". 

Service accounts use `--file-location`
```shell
~/code/github.com/redhat-developer/app-services-cli main *21 ❯ rhoas service-account create --                                                                                       12:50:17
--debug          -- Enable debug mode
--description    -- Description for the service account. Only alphanumeric characters and '-', '.', ',' are accepted.
--file-format    -- Format in which to save the service account credentials. Choose from: "env", "json", "properties"
--file-location  -- Sets a custom file location to save the credentials.
--help           -- Show help for a command
--name           -- Name of the service account.
--overwrite      -- Forcibly overwrite a credentials file if it already exists.
```

wtrocki: So creating PR to remove -f flag. Anyting else?

craicoverflow: What you mean "anything else"? Should it be called "--file-location" maybe?

wtrocki: @craicoverflow Please create issues with your preferences and assign them to me.

I  jumped to fix this without disscussion. For the future I will create issues for even smaller fixes to avoid going against you preferences.

PR #936, opened 2021-08-19
feat: display registry status in status cmd

### LINKED ISSUE #902: Service registry support in status command

Add service registry to status command

rhoas status should show active registry that is seleteced (if any)

rkpattnaik780:
`rhoas status` should display the status of registry command if selected.

Closes #902

Screenshot from 2021-08-19 14-07-55

### Verification Steps

1. Select a service registry using the `service-registry` use command.
```
./rhoas service-registry use --id 195
```
2. Run the `rhoas status` command, you should be able to see the status of registry along with kafka instance.
3. To see the status of registry specifically, run `./rhoas status registry`

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/status/status.go
+++ b/pkg/cmd/status/status.go
@@ -24,10 +24,11 @@ import (
 )
 
 const (
-	kafkaSvcName = "kafka"
+	kafkaSvcName    = "kafka"
+	registrySvcName = "registry"
 )
 
-var validServices = []string{kafkaSvcName}
+var validServices = []string{kafkaSvcName, registrySvcName}
 
 type Options struct {
 	IO         *iostreams.IOStreams
--- a/pkg/cmd/status/status.go
+++ b/pkg/cmd/status/status.go
@@ -9,6 +9,7 @@ import (
 	flagutil "github.com/redhat-developer/app-services-cli/pkg/cmdutil/flags"
 	"github.com/redhat-developer/app-services-cli/pkg/connection"
 	"github.com/redhat-developer/app-services-cli/pkg/localize"
+	"github.com/redhat-developer/app-services-cli/pkg/profile"
 
 	"github.com/redhat-developer/app-services-cli/pkg/cmdutil/flags"
 
@@ -25,10 +26,10 @@ import (
 
 const (
 	kafkaSvcName    = "kafka"
-	registrySvcName = "registry"
+	registrySvcName = "service-registry"
 )
 
-var validServices = []string{kafkaSvcName, registrySvcName}
+var validServices = []string{kafkaSvcName}
 
 type Options struct {
 	IO         *iostreams.IOStreams
@@ -42,6 +43,11 @@ type Options struct {
 }
 
 func NewStatusCommand(f *factory.Factory) *cobra.Command {
+
+	if profile.DevModeEnabled() {
+		validServices = append(validServices, registrySvcName)
+	}
+
 	opts := &Options{
 		IO:         f.IOStreams,
 		Config:     f.Config,
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -10,18 +10,21 @@ import (
 
 	"github.com/redhat-developer/app-services-cli/pkg/connection"
 	"github.com/redhat-developer/app-services-cli/pkg/kafka/kafkaerr"
+	"github.com/redhat-developer/app-services-cli/pkg/serviceregistry"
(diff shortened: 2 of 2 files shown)


wtrocki:
Verified. Do we need to use dev mode for status? It feels like the only side effect in production would be to see debug message (in else statement).

pkg/status/status.go
```diff
@@ -75,6 +85,21 @@ func Get(ctx context.Context, opts *Options) (status *Status, ok bool, err error
...
+		if registryCfg != nil && registryCfg.InstanceID != "" {
+			// nolint:govet
+			registry, newErr := getRegistryStatus(ctx, api.ServiceRegistryMgmt(), registryCfg.InstanceID)
+			if newErr != nil {
+				return status, ok, err
+			}
+			status.Registry = registry
+			ok = true
+		} else {
+			logger.Debug("No service registry is currently used, skipping status check")
```
wtrocki, line 99: That is going to be printed when dev mode is disabled. Since this is debug message I think we can leave with it. 
Alternatively entire service registry part should be wrapped in dev check (preferred)

pkg/cmd/status/status.go
```diff
@@ -24,10 +24,11 @@ import (
 )
 
 const (
-	kafkaSvcName = "kafka"
+	kafkaSvcName    = "kafka"
+	registrySvcName = "registry"
```
craicoverflow, line 28: The name of the service will be known to the user as "service-registry"

pkg/cmd/status/status.go
```diff
@@ -24,10 +24,11 @@ import (
...
 )
 
-var validServices = []string{kafkaSvcName}
+var validServices = []string{kafkaSvcName, registrySvcName}
```
wtrocki, line 31: Missing documentation for valid services

craicoverflow, line 31: The registry service features should currently only be visible/accessible when the `RHOAS_DEV` flag is enabled.

pkg/status/status.go
```diff
@@ -75,6 +85,21 @@ func Get(ctx context.Context, opts *Options) (status *Status, ok bool, err error
 		}
 	}
 
+	if stringInSlice("registry", opts.Services) {
```
craicoverflow, line 88: This should also be wrapped in a `RHOAS_DEV=true` conditional.

wtrocki:
Verified. Do we need to use dev mode for status? It feels like the only side effect in production would be to see debug message (in else statement).

pkg/cmd/status/status.go
wtrocki:
```
View status information of your currently used services.
Choose to view the status of all services with "rhoas status" or specific services with "rhoas status [service]"

To use a different service run "rhoas <service> use [args] [flags]".

```

There is no list of services here (although they are suggested by cmd

wtrocki (approved):
LGTM - minor questions asked but no need to address those

pkg/cmd/status/status.go
wtrocki, line 31:
```suggestion
var validServices = []string{kafkaSvcName}
if profile.isDevEnaled() {
 validServices = []string{kafkaSvcName, registrySvcName}
 }
```

pkg/status/status.go
```diff
@@ -10,18 +10,21 @@ import (
...
 	"github.com/redhat-developer/app-services-cli/pkg/logging"
+	srsmgmtv1 "github.com/redhat-developer/app-services-sdk-go/registrymgmt/apiv1/client"
 )
 
 const tagTitle = "title"
 
 type Status struct {
-	Kafka *KafkaStatus `json:"kafka,omitempty" title:"Kafka"`
+	Kafka    *KafkaStatus    `json:"kafka,omitempty" title:"Kafka"`
+	Registry *RegistryStatus `json:"registry,omitempty" title:"Registry"`
```
craicoverflow, line 27: Should we call this "Service Registry"?

```suggestion
	Registry *RegistryStatus `json:"registry,omitempty" title:"Service Registry"`
```

craicoverflow (commented):
A few comments, but I think the most important one is that this should only be visible with `RHOAS_DEV`.

pkg/status/status.go
rkpattnaik780: Enclosing all of it inside a condition checking dev preview.

wtrocki (approved):
Verified with and without dev preview

craicoverflow (approved):

PR #943, opened 2021-08-19
fix: rename flag for get and download commands

wtrocki:


wtrocki:
As this is exactly what we agreed on merging to master to continue with fixes

PR #950, opened 2021-08-20
fix!: remove version from persistent flags

wtrocki:
##### Valid output:
rhoas --version
rhoas version

##### Error

rhoas cluster --version

Ignore last commit when reviewing (docs changes)

### DIFF
--- a/cmd/rhoas/main.go
+++ b/cmd/rhoas/main.go
@@ -29,7 +29,7 @@ func main() {
 	}
 
 	buildVersion := build.Version
-	cmdFactory := factory.New(build.Version, localizer)
+	cmdFactory := factory.New(localizer)
 	logger, err := cmdFactory.Logger()
 	if err != nil {
 		fmt.Println(cmdFactory.IOStreams.ErrOut, err)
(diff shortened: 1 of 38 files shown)


cmd/rhoas/main.go
```diff
@@ -58,12 +58,9 @@ func main() {
...
 		return
 	}
-
-	if err != nil {
-		logger.Error(wrapErrorf(err, localizer))
-		build.CheckForUpdate(context.Background(), logger, localizer)
-		os.Exit(1)
-	}
+	logger.Error(wrapErrorf(err, localizer))
+	build.CheckForUpdate(context.Background(), logger, localizer)
```
craicoverflow, line 62: Does this check for update for every command regardless of success rate?

craicoverflow (approved):

cmd/rhoas/main.go
```diff
@@ -58,12 +58,9 @@ func main() {
 		}
 		return
 	}
-
-	if err != nil {
-		logger.Error(wrapErrorf(err, localizer))
-		build.CheckForUpdate(context.Background(), logger, localizer)
-		os.Exit(1)
-	}
+	logger.Error(wrapErrorf(err, localizer))
```
wtrocki, line 61: FIxing logic here. It should be the opposite

wtrocki: Actually I needed to do it this way because of linting issue. 
So logic is as follows:

if not error and debug check update
otherwise (error) print error and check update 

it is confusing but os.Exit(1) typically should be the last statement (there is no return after that ;) :)

Issue #927, opened 2021-08-18
rhoas service-registry use and kafka use has inconsistent flag and positional implementation


wtrocki:
Feedback

```
The command allows the name of the registry instance to be included in the base command but required the id of the instance when using the “--id” flag:
Works: rhoas service-registry use ewittman
Works: rhoas service-registry use --id 3
Does not work: rhoas service-registry use 3
Does not work: rhoas service-registry --id ewittman
Note: this is fine but should be made clear in the help text, which indicates “<id>” in both cases.
```

wtrocki:
CC @craicoverflow

craicoverflow:
https://github.com/redhat-developer/app-services-cli/blob/f8cd89d6305cdc1d507d947acd95d0babd29184f/pkg/localize/locales/en/cmd/registry_crud.en.toml#L92-L95

The example for this command is invalid.

wtrocki:
I think we have 2 issues here:

1) we need to add --name flag to kafka and registry use command
2) we should not have any positional support for use (docs issue)

wtrocki:
Triaged with @craicoverflow and @rkpattnaik780

PR #952, opened 2021-08-20
 fix!: replace positional use with explicity --name flag

wtrocki:
Context and discussion #927 
## Verification

rhoas kafka use (will print help)
rhoas kafka use --name=yourkafka
rhoas kafka use --id=id

### DIFF
--- a/pkg/cmd/kafka/use/use.go
+++ b/pkg/cmd/kafka/use/use.go
@@ -47,14 +47,12 @@ func NewUseCommand(f *factory.Factory) *cobra.Command {
 		Short:   opts.localizer.MustLocalize("kafka.use.cmd.shortDescription"),
 		Long:    opts.localizer.MustLocalize("kafka.use.cmd.longDescription"),
 		Example: opts.localizer.MustLocalize("kafka.use.cmd.example"),
-		Args:    cobra.RangeArgs(0, 1),
+		Args:    cobra.NoArgs,
 		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 			return cmdutil.FilterValidKafkas(f, toComplete)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if len(args) > 0 {
-				opts.name = args[0]
-			} else if opts.id == "" {
+			if opts.id == "" && opts.name == "" {
 				if !opts.IO.CanPrompt() {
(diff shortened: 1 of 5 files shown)


wtrocki:
@craicoverflow as discussed this is something that we can employ to minimize confusion. Let me know if that works for you.

pkg/cmd/kafka/use/use.go
```diff
@@ -70,7 +68,7 @@ func NewUseCommand(f *factory.Factory) *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&opts.id, "id", "", opts.localizer.MustLocalize("kafka.use.flag.id"))
-
+	cmd.Flags().StringVar(&opts.id, "name", "", opts.localizer.MustLocalize("kafka.use.flag.name"))
```
rkpattnaik780, line 71:
```suggestion
	cmd.Flags().StringVar(&opts.name, "name", "", opts.localizer.MustLocalize("kafka.use.flag.name"))
```

pkg/cmd/registry/use/use.go
```diff
@@ -63,6 +61,7 @@ func NewUseCommand(f *factory.Factory) *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&opts.id, "id", "", opts.localizer.MustLocalize("registry.use.flag.id"))
+	cmd.Flags().StringVar(&opts.id, "name", "", opts.localizer.MustLocalize("registry.use.flag.name"))
```
rkpattnaik780, line 64:
```suggestion
	cmd.Flags().StringVar(&opts.name, "name", "", opts.localizer.MustLocalize("registry.use.flag.name"))
```

rkpattnaik780 (changes requested):

wtrocki:
@rkpattnaik780 Forgot to push local change. Thanks for spotting problem. This is good now

rkpattnaik780 (approved):

pkg/localize/locales/en/cmd/registry_crud.en.toml
```diff
@@ -87,6 +87,9 @@ one = 'Use a Service Registry instance'
 [registry.cmd.use.longDescription]
 one = ''' 
 Select a Service Registry instance to use with all instance-specific commands.
+You can specify a Service Registry instance by --name or --id.
+
+When you set the Service Registry instance to be used, it is set as the current instance for all rhoas service-registry artifacts commands.
```
wtrocki, line 92:
```suggestion
When you set the Service Registry instance to be used, it is set as the current instance for all rhoas service-registry artifact commands.
```

PR #981, opened 2021-08-26
metadata commands

wtrocki:
Added new command to edit metadata.
Bonus point - added small utility to open editor in interactive mode (same as in kubernetes)

Interactive mode - opens your system editor and lets you edit values
```
rhoas service-registry artifact metadata-set --artifact-id=test 
```

Explicit mode:

```
rhoas service-registry artifact metadata-set --artifact-id=test --name=test --description=test
```

Verify if changes were applied

```
 rhoas service-registry artifact metadata-get --artifact-id=test
 ```

### DIFF
--- a/pkg/editor/editor.go
+++ b/pkg/editor/editor.go
@@ -38,6 +38,7 @@ func (e *Editor) Run() ([]byte, error) {
 	defer os.Remove(path)
 
 	args := []string{shell, shellCommandFlag, fmt.Sprintf("%s %s", editor, path)}
+	// #nosec 204
 	cmd := exec.Command(args[0], args[1:]...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
--- a/pkg/editor/editor.go
+++ b/pkg/editor/editor.go
@@ -1,0 +1,56 @@
+package editor
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
+type Editor struct {
+	content  []byte
+	filename string
+}
+
+func New(content []byte, filename string) *Editor {
+	return &Editor{
+		content:  content,
+		filename: filename,
+	}
+}
+
+func (e *Editor) Run() ([]byte, error) {
+	shell := defaultShell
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by rkpattnaik780 — APPROVED** [review]
Tested with linux. LGTM!

**Comment by rkpattnaik780** [comment]
Out of the scope for this PR, we should do something about error messages with two periods.
Screenshot from 2021-08-31 12-48-26

**Comment by wtrocki** [comment]
> Out of the scope for this PR, 

Never bother to think in this categories. Mention every problem you see..


I think this is duplicate of error handling bug. Since we do not get error codes from api we have no way to tell what is happening (group can be invalid, artifact invalid etc. so we print message back. There is issue aready to cover it and I have created issues in apicurio as well

**Comment by rkpattnaik780** [comment]
> > Out of the scope for this PR,
> 
> Never bother to think in this categories. Mention every problem you see..
> 
> I think this is duplicate of error handling bug. Since we do not get error codes from api we have no way to tell what is happening (group can be invalid, artifact invalid etc. so we print message back. There is issue aready to cover it and I have created issues in apicurio as well

That's great. Thanks for letting me know.

**Comment by wtrocki** [comment]
Verified on windows. I think we need to document how to change your editor by using env variable.

Live edit on windows used notepad :)

File: pkg/editor/editor.go [hunk 1]
```diff
@@ -0,0 +1,57 @@
+package editor
```
wtrocki, line 1: This is an new helper added in PR 

@craicoverflow it can be used also in other parts of the CLI - for example specifying some complex topic structures and in future for Authz rules.

pkg/editor/windows.go
```diff
@@ -0,0 +1,9 @@
+// +build windows
+
+package editor
+
+const (
+	defaultShell     = "cmd"
+	shellCommandFlag = "/C"
+	defaultEditor    = "notepad"
```
wtrocki, line 8: Done research. There is no other preinstalled editor that does support json formatting etc.

pkg/editor/editor.go
craicoverflow: ooh very nice!

PR #982, opened 2021-08-26
refactor: converted i18n files from TOML to YAML

### LINKED ISSUE #757: Merge kafka crud i18n into single file
## Description

Currently navigation between multiple files/common elements is very hard. 

We should have developer UX oriented i18n (not file oriented that causes a lot of confusion and has number of entries duplicated). Without using single file we cannot detect duplicates which are currently in main branch.

## Proposed files
```
kafka_crud.en.toml
registry_crud.en.toml
serviceaccount_crud.en.toml
cluster_en.toml
kafka_consumer_group.en.toml
kafka_topics.en.toml
```

### LINKED ISSUE #976: Convert i18n files from TOML to YAML
TOML is a sufficient markup format for simple configurations. But we are dealing with 100s of text strings, so it is actually tricky to find the locale ID we need.

It is also hard to "group" strings into a hierarchical tree. TOML does have this notion of dot-notated grouping, but visually it is flat.

This change would really improve our developer experience.

alexal:
Closes #976
Closes #757


alexal:
@craicoverflow just wondering whether you had a chance to review the changes?

craicoverflow:
Hey @alexal - not yet but will get to it soon.

craicoverflow:
We came to the conclusion that while it is easier to group keys in YAML, it actually makes it more difficult to look up the property when looking at the i18n ID.

With TOML, they are exactly the same. Eg:

https://github.com/redhat-developer/app-services-cli/blob/8b3ac77c29ca01e2c68ba59d02e14e153a9cf517/pkg/cmd/cluster/bind/bind.go#L49
https://github.com/redhat-developer/app-services-cli/blob/4930e17e17be946a01c6c5db7b9115d01edf7482/pkg/localize/locales/en/cmd/cluster_bind.en.toml#L1-L2

For that reason, TOML is a better choice. I'm going to close this PR and the issue. Thank you for your contributions 🙏🏻

PR #991, opened 2021-08-30
feat: added a timeout to the login command

### LINKED ISSUE #989: Add a timeout when logging in via the web browser
The `login` command should have a timeout so that it does not "hang" if the command fails.

alexal:
Closes #989

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -3,6 +3,7 @@ package build
 import (
 	"context"
 	"runtime/debug"
+	"time"
 
 	"github.com/google/go-github/github"
 	"github.com/redhat-developer/app-services-cli/pkg/color"
@@ -49,6 +50,7 @@ var (
 	StagingAPIURL               = "https://api.stage.openshift.com"
 	DefaultClientID             = "rhoas-cli-prod"
 	DefaultOfflineTokenClientID = "cloud-services"
+	DefaultLoginTimeout         = 30 * time.Second
 	// #nosec G101
 	OfflineTokenURL      = "https://console.redhat.com/openshift/token"
 	ProductionAuthURL    = "https://sso.redhat.com/auth/realms/redhat-external"
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -50,7 +50,7 @@ var (
 	StagingAPIURL               = "https://api.stage.openshift.com"
 	DefaultClientID             = "rhoas-cli-prod"
 	DefaultOfflineTokenClientID = "cloud-services"
-	DefaultLoginTimeout         = 30 * time.Second
+	DefaultLoginTimeout         = 60 * time.Second
 	// #nosec G101
 	OfflineTokenURL      = "https://console.redhat.com/openshift/token"
 	ProductionAuthURL    = "https://sso.redhat.com/auth/realms/redhat-external"
--- a/pkg/cmd/login/login.go
+++ b/pkg/cmd/login/login.go
@@ -116,6 +116,7 @@ func NewLoginCmd(f *factory.Factory) *cobra.Command {
 			case res := <-ch:
 				return res
 			case <-time.After(build.DefaultLoginTimeout):
+				//nolint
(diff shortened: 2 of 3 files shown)


alexal:
@craicoverflow have you had a chance to review the changes?

craicoverflow:
The build is failing - this needs to be fixed.

> @craicoverflow have you had a chance to review the changes?

We have a number of tasks to complete, but don't worry - this is on the backlog.

pkg/cmd/login/login.go
```diff
@@ -103,7 +104,20 @@ func NewLoginCmd(f *factory.Factory) *cobra.Command {
 				opts.Logger.Info(opts.localizer.MustLocalize("login.log.info.sshLoginDetected", localize.NewEntry("OfflineTokenURL", build.OfflineTokenURL)))
 			}
 
-			return runLogin(opts)
+			ch := make(chan error, 1)
```
wtrocki, line 107: My take would be to have timeout util rather than code directly embedded into login function. 

Timeout can be done many different ways. This  puts timeout on entire login method. 
We can also do it individually using middleware: https://github.com/go-http-utils/timeout

My take is that this approach is better although it needs to be done as separate function/helper

pkg/localize/locales/en/cmd/login.en.toml
```diff
@@ -78,6 +78,9 @@ one = 'You are now logged in to <span style="font-weight:700">{{.Host}}</span> a
 description = 'Error message for when scheme is missing from the Auth URL'
 one = 'scheme missing from URL "{{.URL}}". Please add either "https" or "https"'
 
+[login.error.timeout.expired]
+one = 'login attempt has expired'
```
wtrocki, line 82:
```suggestion
one = 'login operation took too long. Please try again '
```

internal/build/build.go
```diff
@@ -49,6 +50,7 @@ var (
 	StagingAPIURL               = "https://api.stage.openshift.com"
 	DefaultClientID             = "rhoas-cli-prod"
 	DefaultOfflineTokenClientID = "cloud-services"
+	DefaultLoginTimeout         = 30 * time.Second
```
wtrocki, line 53: This is way to low (in case when SSO session is not active it gives only 30 seconds to login)

pkg/cmd/login/login.go
wtrocki: Just checked. We cannot use http middleware in this case. This code is the best way forward IMHO

craicoverflow (commented):
So, the best way to do this would be to use `context.WithTimeout`, but only on the auth requests.

I believe that the timeout should be focused on the network request, not the command itself.

Currently we are using `context.WithCancel` [here](https://github.com/redhat-developer/app-services-cli/blob/b4128d56a571d998634901980f56fb958384c55c/pkg/auth/login/login.go#L253), this could be replaced with `context.WithTimeout`.

Since one context is created for each OpenID client, using 30 seconds for each might be satisfactory (as when combined it gives 60 seconds) - looking for @wtrocki's opinion before we agree with this approach.

wtrocki:
So first callback requires human interaction (if you need to login). Second callback uses SSO so it is almost instant. 
I like global timeout but if we want to go with individual ones it should be 60 for first and 10 seconds for second.

craicoverflow:
> So first callback requires human interaction (if you need to login). Second callback uses SSO so it is almost instant.
> I like global timeout but if we want to go with individual ones it should be 60 for first and 10 seconds for second.

Global timeout makes sense then. We should try to avail of the built-in `context.WithTimeout` for this then.

The context can be passed to `login.Execute(ctx)` from the login command:

https://github.com/redhat-developer/app-services-cli/blob/b4128d56a571d998634901980f56fb958384c55c/pkg/auth/login/login.go#L41

craicoverflow:
Needs rebase.

wtrocki:
Rebase needed.

alexal:
@wtrocki , @craicoverflow please see a new PR #1027. This one can be closed as I lost my old local repo.

PR #996, opened 2021-08-31
fix: invalid i18n strings

wtrocki:
Fixed invalid keys assigned as common key.

Verification:

```
rhoas service-registry delete -h 
rhoas service-registry describe -h 
```


wtrocki:
Minor fix. Merging

PR #1006, opened 2021-08-31
feat: added a custom user-agent

### LINKED ISSUE #994: Cli should use custom user agent when making requests to all SDK
## Description

See Connection structure that does define all APIs. 

When building API objects we need to override User-Agent with RHOAS CLI name and it's version.

alexal:
Closes #994


pkg/cmd/registry/describe/describe.go
```diff
@@ -93,7 +93,7 @@ func runDescribe(opts *Options) error {
 
 	api := connection.API()
 
-	var registry *srsmgmtv1.RegistryRest
+	var registry *srsmgmtv1.Registry
```
wtrocki, line 96: @rkpattnaik780 once new sdk is released it is best to integrate it with cli

scripts/check-docs.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
 set -uo pipefail
-
-if [[ $(git diff --stat docs/commands | wc -l) != 0 ]]; then
-  echo "./docs/commands has changes"
+commands=$(git diff --stat docs/commands | wc -l)
```
wtrocki, line 4: Can we revert this? Why this changes

wtrocki: actually much better implementation.
Lets just try keep prs focused for future.v

wtrocki:
@alexal Not sure who you are and where you from but this is amazing. You have been doing more contributions than our entire dev team combined.

Thank you so much for doing it!

pkg/cmd/registry/describe/describe.go
wtrocki: This was breaking change

wtrocki (approved):
Verified

PR #1008, opened 2021-09-01
fix: use a consistent flag name for output files

craicoverflow:
BREAKING CHANGE: Use a consistent "--output-file" to specify a file location.

The CLI uses a couple of different flag names when specifying a path to save a file to.

The preferred POSIX standard is `--output` (see [here](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html). Unfortunately the rhoas CLI uses `output` much more commonly for a different meaning (to specify the format to print to stdout). It would be confusing the use this same flag name here.

`output-file` is close in name to `output` and is very self descriptive compared to `--file-location`.

I am happy to discuss an alternative name with you @wtrocki and @rkpattnaik780 - we just need to be consistent.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -129,7 +129,7 @@ rhoas service-registry artifact create --type=JSON my-artifact.json
 	}
 
 	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
-	cmd.Flags().StringVar(&opts.file, "output-file", "", "File location of the artifact")
+	cmd.Flags().StringVar(&opts.file, "file", "", "File location of the artifact")
 
 	cmd.Flags().StringVarP(&opts.artifact, "artifact-id", "a", "", "Id of the artifact")
 	cmd.Flags().StringVarP(&opts.group, "group", "g", util.DefaultArtifactGroup, "Artifact group")
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -129,7 +129,7 @@ rhoas service-registry artifact create --type=JSON my-artifact.json
 	}
 
 	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
-	cmd.Flags().StringVar(&opts.file, "file", "", "File location of the artifact")
+	cmd.Flags().StringVar(&opts.file, "output-file", "", "File location of the artifact")
 
 	cmd.Flags().StringVarP(&opts.artifact, "artifact-id", "a", "", "Id of the artifact")
 	cmd.Flags().StringVarP(&opts.group, "group", "g", util.DefaultArtifactGroup, "Artifact group")
--- a/pkg/cmd/registry/artifact/crud/get/get.go
+++ b/pkg/cmd/registry/artifact/crud/get/get.go
@@ -60,7 +60,7 @@ For fetching artifacts using global identifiers please use "service-registry dow
 rhoas service-registry artifact get --artifact-id=my-artifact
 
 ## Get latest artifact with name "my-artifact" from group "my-group" and save it to artifact.json file
-rhoas service-registry artifact get --artifact-id=my-artifact --group=my-group --file=artifact.json
+rhoas service-registry artifact get --artifact-id=my-artifact --group=my-group --output-file=artifact.json
 
 ## Get latest artifact and pipe it to other command 
 rhoas service-registry artifact get --artifact-id=my-artifact | grep -i 'user'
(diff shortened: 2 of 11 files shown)


pkg/cmd/registry/artifact/crud/create/create.go
```diff
@@ -129,7 +129,7 @@ rhoas service-registry artifact create --type=JSON my-artifact.json
 	}
 
 	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
-	cmd.Flags().StringVar(&opts.file, "file", "", "File location of the artifact")
+	cmd.Flags().StringVar(&opts.file, "output-file", "", "File location of the artifact")
```
rkpattnaik780, line 132: This works more like an input file I guess, wdyt?
Same query for `artifact update`.

craicoverflow: Sorry, yes you are right. this is for input 🤦🏻

craicoverflow: Please "Request changes" on this PR to prevent it from being merged.

pkg/cmd/registry/artifact/download/download.go
```diff
@@ -99,7 +99,7 @@ rhoas service-registry artifact download --hash=c71d239df91726fc519c6eb72d318ec6
 	cmd.Flags().Int64VarP(&opts.globalId, "global-id", "", unusedFlagIdValue, "Global ID of the artifact")
 	cmd.Flags().Int64VarP(&opts.contentId, "content-id", "", unusedFlagIdValue, "ContentId of the artifact")
 
-	cmd.Flags().StringVarP(&opts.outputFile, "file-location", "", "", "Location of the output file")
+	cmd.Flags().StringVar(&opts.outputFile, "output-file", "", "Location of the output file")
```
wtrocki, line 102: Download and get might need to have the same flag (they basically do the same)

pkg/cmd/registry/artifact/crud/get/get.go
```diff
@@ -95,7 +95,7 @@ rhoas service-registry artifact get --artifact-id=myartifact --version=4
 	cmd.Flags().StringVarP(&opts.artifact, "artifact-id", "a", "", "Id of the artifact")
 	cmd.Flags().StringVarP(&opts.group, "group", "g", util.DefaultArtifactGroup, "Artifact group")
 	cmd.Flags().StringVar(&opts.registryID, "instance-id", "", "Id of the registry to be used. By default uses currently selected registry")
-	cmd.Flags().StringVar(&opts.outputFile, "file-location", "", "Location of the output file")
+	cmd.Flags().StringVar(&opts.outputFile, "file", "", "Location of the output file")
```
wtrocki, line 98: My assumption is that you meant to have it as others?
```suggestion
	cmd.Flags().StringVar(&opts.outputFile, "output-file", "", "Location of the output file")
```

wtrocki:
Looking

pkg/cmd/serviceaccount/create/create.go
```diff
@@ -94,7 +94,7 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 	cmd.Flags().StringVar(&opts.name, "name", "", opts.localizer.MustLocalize("serviceAccount.create.flag.name.description"))
 	cmd.Flags().StringVar(&opts.description, "description", "", opts.localizer.MustLocalize("serviceAccount.create.flag.description.description"))
 	cmd.Flags().BoolVar(&opts.overwrite, "overwrite", false, opts.localizer.MustLocalize("serviceAccount.common.flag.overwrite.description"))
-	cmd.Flags().StringVar(&opts.filename, "file-location", "", opts.localizer.MustLocalize("serviceAccount.common.flag.fileLocation.description"))
+	cmd.Flags().StringVar(&opts.filename, "output-file", "", opts.localizer.MustLocalize("serviceAccount.common.flag.fileLocation.description"))
```
wtrocki, line 97: In this case we dealing with input file.
I'm not 100% convinced we need output/input as that can be derrived from command:

- get downloads file
- create sends file

It is less typing

wtrocki: Anyway this is and edit command should have different values than rest.
Apart from that all good

craicoverflow: No, we are dealing with an output file here. Yo create a service account and it outputs the credentials to a file.

craicoverflow: > Anyway this is and edit command should have different values than rest.

Sorry, not sure I understand here?

wtrocki: OMG - service account not service registry :).

SHAME! Let's forget about this 📸

wtrocki:
Fun fact. I discovered that flags can have `silent aliases`.  They are still supported but not printed in help. 
Those can be the way to keep backwards compatibility in the future.

craicoverflow:
> Fun fact. I discovered that flags can have `silent aliases`. They are still supported but not printed in help.
> Those can be the way to keep backwards compatibility in the future.

Yeah I considered this to solve the "serviceaccount/service-account" conundrum, but its a pity we cannot get it to appear in completions.

wtrocki (approved):

wtrocki:
@craicoverflow This might be offtopic but since this is breaking change and we have been encouraging people to update CLI maybe we should drop another update (considering that we will stop breaking changes from now ;) 

This way all examples and guides can be corrected (even external ones) as soon as possible. 
In coming weeks we will get much more exposure with external materials etc. so our ability to push breaking changes will shrink

craicoverflow:
> This might be offtopic but since this is breaking change and we have been encouraging people to update CLI maybe we should drop another update (considering that we will stop breaking changes from now ;)

Yes that is why I am doing this now - before customers come.

PR #1010, opened 2021-09-01
style: error strings should not be capitalized

alexal:
Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

### DIFF
--- a/cmd/rhoas/main.go
+++ b/cmd/rhoas/main.go
@@ -112,5 +112,5 @@ func initConfig(f *factory.Factory) error {
 }
 
 func wrapErrorf(err error, localizer localize.Localizer) error {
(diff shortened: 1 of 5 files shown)


cmd/rhoas/main.go
```diff
@@ -112,5 +112,5 @@ func initConfig(f *factory.Factory) error {
 }
 
 func wrapErrorf(err error, localizer localize.Localizer) error {
-	return fmt.Errorf("Error: %w. %v", err, localizer.MustLocalize("common.log.error.verboseModeHint"))
```
craicoverflow, line 115: Since this is the root error, which wraps all others, it is fine to have this capitalized, as it prefixes sub-errors.

alexal: Sure

PR #1014, opened 2021-09-01
chore: variable 'connection' collides with imported package name

alexal:
Variable 'connection' collides with imported package name


wtrocki:
> Variable 'connection' collides with imported package name

Out of curiosity where this comes from. Is that warning of some sorts from golangci lint.

alexal:
> > Variable 'connection' collides with imported package name
> 
> Out of curiosity where this comes from. Is that warning of some sorts from golangci lint.
That is from GoLand Code Inspector.

PR #1018, opened 2021-09-01
chore: variable 'config' collides with imported package name

alexal:
Variable 'config' collides with imported package name


wtrocki:
Do you mind creating single PR with `x` colides with `x`. This generates a loot of buzz for repo and consumes time for verification (each PR needs to be tested etc.)

PR #1022, opened 2021-09-01
chore: variable 'editor' collides with imported package name

alexal:
Variable 'editor' collides with imported package name

### DIFF
--- a/pkg/cmd/registry/artifact/metadata/set.go
+++ b/pkg/cmd/registry/artifact/metadata/set.go
@@ -177,8 +177,8 @@ func runEditor(currentMetadata *registryinstanceclient.EditableMetaData) (*regis
 	if err != nil {
 		return nil, err
 	}
-	edit := editor.New(metadataJson, "metadata.json")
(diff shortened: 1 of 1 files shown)


pkg/cmd/registry/artifact/metadata/set.go
```diff
@@ -177,8 +177,8 @@ func runEditor(currentMetadata *registryinstanceclient.EditableMetaData) (*regis
 	if err != nil {
 		return nil, err
 	}
-	editor := editor.New(metadataJson, "metadata.json")
-	output, err := editor.Run()
+	edit := editor.New(metadataJson, "metadata.json")
+	output, err := edit.Run()
```
wtrocki, line 181:
```suggestion
	systemEditor := editor.New(metadataJson, "metadata.json")
	output, err := systemEditor.Run()
```

wtrocki:
When changing variables please try to be more verbose rather than making things shorter. `edit` is not good variable name :)

PR #1023, opened 2021-09-02
chore: code cleanup, declaration redundancy

alexal:
Code cleanup, declaration redundancy

### DIFF
--- a/pkg/profile/profile.go
+++ b/pkg/profile/profile.go
@@ -8,7 +8,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var DevPreviewEnv = "RHOAS_DEV"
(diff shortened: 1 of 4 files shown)


pkg/profile/profile.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var DevPreviewEnv string = "RHOAS_DEV"
+var DevPreviewEnv = "RHOAS_DEV"
```
craicoverflow, line 11: This could actually be a a `const` while at it.

wtrocki:
```suggestion
const DevPreviewEnv = "RHOAS_DEV"
```

wtrocki (approved):

PR #1027, opened 2021-09-03
feat: added a global timeout to the login command

### LINKED ISSUE #989: Add a timeout when logging in via the web browser

alexal:
Closes #989

### DIFF
--- a/pkg/auth/login/login.go
+++ b/pkg/auth/login/login.go
@@ -44,7 +44,6 @@ func (a *AuthorizationCodeGrant) Execute(ctx context.Context, ssoCfg *SSOConfig,
 	if err := a.loginSSO(ctx, ssoCfg); err != nil {
 		return err
 	}
-	a.Logger.Info(a.Localizer.MustLocalize("login.log.info.loggedIn"))
 
 	masSSOHost := masSSOCfg.AuthURL.Host
 
@@ -53,6 +52,7 @@ func (a *AuthorizationCodeGrant) Execute(ctx context.Context, ssoCfg *SSOConfig,
 	if err := a.loginMAS(ctx, masSSOCfg); err != nil {
 		return err
 	}
+	a.Logger.Info(a.Localizer.MustLocalize("login.log.info.loggedIn"))
 	a.Logger.Info(a.Localizer.MustLocalize("login.log.info.loggedInMAS", localize.NewEntry("Host", masSSOHost)))
 
 	return nil
--- a/pkg/cmd/login/login.go
+++ b/pkg/cmd/login/login.go
@@ -6,11 +6,10 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
-	"net/http"
-	"net/url"
-
 	"github.com/redhat-developer/app-services-cli/internal/build"
 	"golang.org/x/oauth2"
+	"net/http"
+	"net/url"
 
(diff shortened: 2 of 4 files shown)


pkg/auth/login/login.go
```diff
@@ -53,6 +52,7 @@ func (a *AuthorizationCodeGrant) Execute(ctx context.Context, ssoCfg *SSOConfig,
 	if err := a.loginMAS(ctx, masSSOCfg); err != nil {
 		return err
 	}
+	a.Logger.Info(a.Localizer.MustLocalize("login.log.info.loggedIn"))
```
alexal, line 55: @wtrocki , @craicoverflow 

I've moved this here because we may end up in a situation like this

Logging in...

Logged in successfully
Logging in to identity.api.openshift.com...
Error: login operation took too long. Please try again. Run the command in verbose mode using the -v flag to see more information

After I did that, everything looks much better

Logging in...

Logging in to identity.api.openshift.com...
Error: login operation took too long. Please try again. Run the command in verbose mode using the -v flag to see more information

alexal:
Also, it seems like global timeout is enough and createClientContext don't need to be changed.

pkg/cmd/login/login.go
```diff
@@ -163,7 +162,15 @@ func runLogin(opts *Options) (err error) {
 			RedirectPath: build.MASSSORedirectPath,
 		}
 
-		if err = loginExec.Execute(context.Background(), ssoCfg, masSsoCfg); err != nil {
+		// Creating a global context with timeout
+		ctx, cancel := context.WithTimeout(context.Background(), build.DefaultLoginTimeout)
+		defer cancel()
+
+		if err = loginExec.Execute(ctx, ssoCfg, masSsoCfg); err != nil {
```
craicoverflow, line 169:
```suggestion
		if err = loginExec.Execute(ctx, ssoCfg, masSsoCfg); err != nil && errors.Is(err, context.DeadlineExceeded) {
```

This could work, to save a nested if statement.

craicoverflow (approved):
Awesome work!

pkg/auth/login/login.go
```diff
@@ -53,6 +52,7 @@ func (a *AuthorizationCodeGrant) Execute(ctx context.Context, ssoCfg *SSOConfig,
...
 	a.Logger.Info(a.Localizer.MustLocalize("login.log.info.loggedInMAS", localize.NewEntry("Host", masSSOHost)))
```
craicoverflow, line 56:
```suggestion
	a.Logger.Debug(a.Localizer.MustLocalize("login.log.info.loggedInMAS", localize.NewEntry("Host", masSSOHost)))
```

While you are here, would you mind making this a debug message? I think the new ordering might be a little confusing, and the user does not really need to see two log in messages, now that I think about it :)

pkg/auth/login/login.go
craicoverflow:
```go
a.Logger.Info(a.Localizer.MustLocalize("login.log.info.loggingInMAS", localize.NewEntry("Host", masSSOHost)))
```

Could you make this into a debug message instead? The user does not need to see two login statuses whyen it should happen automatically. The new ordering might be a bit confusing but that's just an effect of repeated login messages

PR #1028, opened 2021-09-04
feat: emoji support for both error and success messages

### LINKED ISSUE #1024: Use ✔️ emoji in success messages
Emoji are commonly used in nowadays to enhance the visual output of the CLI.

```shell
$ rhoasd kafka use --name my-instance
? Select Kafka instance to connect: enda
✔️ Kafka instance "enda" has been set as the current instance.
```

It is a small, nice-to-have feature but I feel it would be a great addition.

### LINKED ISSUE #1025: Prefix error messages with "❌"
We could prefix all error messages with an "❌" emoji to replace the "Error" prefix.

```shell
$ rhoas kafka use --name kascfkadsfk
❌ Kafka instance "kascfkadsfk" not found. Run the command in verbose mode using the -v flag to see more information
```

This will enhance the visual output of the CLI.

alexal:
Closes #1025 
Closes #1024

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -2,6 +2,7 @@ package build
 
 import (
 	"context"
+	"runtime"
 	"runtime/debug"
 
 	"github.com/google/go-github/github"
@@ -57,11 +58,21 @@ var (
 	StagingMasAuthURL           = "https://identity.api.stage.openshift.com/auth/realms/rhoas"
 )
 
-// The emoji sequences
-var (
-	EmojiSuccess = "\u2714\ufe0f"
-	EmojiError   = "\u274c"
-)
+// GetEmojiSuccess returns check mark emoji on every platform, but Windows
+func GetEmojiSuccess() string {
+	if runtime.GOOS != "windows" {
+		return "\u2714\ufe0f"
+	}
+	return ""
+}
+
+// GetEmojiError returns cross mark emoji on every platform, but Windows
+func GetEmojiError() string {
+	if runtime.GOOS != "windows" {
+		return "\u274c"
+	}
+	return "Error:"
+}
 
 func init() {
 	if isDevBuild() {
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -57,6 +57,12 @@ var (
 	StagingMasAuthURL           = "https://identity.api.stage.openshift.com/auth/realms/rhoas"
 )
 
+// The emoji sequences
+var (
+	EmojiSuccess = "\u2714\ufe0f"
+	EmojiError   = "\u274c"
+)
+
 func init() {
 	if isDevBuild() {
 		if info, ok := debug.ReadBuildInfo(); ok && info.Main.Version != "(devel)" {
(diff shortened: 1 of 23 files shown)


alexal:
Error messages:
image

alexal:
Logout
image

alexal:
@wtrocki , @craicoverflow unfortunately, I do not have a Windows machine to test the changes, but I will try to see if I can spawn a temporary instance or something. I will try to do that on Wednesday next week, as I will be out on Monday and Tuesday.

wtrocki:
Ignore. We have aws machine for testing

alexal:
> Ignore windows part. We have aws machine for testing. My point was to first test it on windows if that is going to work. We will check as early as possible.

Thanks, @wtrocki. According to this post https://stackoverflow.com/questions/49476326/displaying-unicode-in-powershell/49481797#49481797, it may work in PowerShell, however, I'm not sure whether it will work in a standard command line or whatever it is.

wtrocki:
https://github.com/Microsoft/Terminal/issues/190

We need to check os in the buildEmoji call and ignore windows

wtrocki (changes requested):
We need to disable windows

wtrocki:
I actually got it supported on windows, with some tweaking.

alexal:
> We need to disable windows

Done, please review.

wtrocki (approved):

internal/build/build.go
```diff
@@ -57,6 +58,22 @@ var (
...
+	return ""
+}
+
+// GetEmojiError returns cross mark emoji on every platform, but Windows
+func GetEmojiError() string {
+	if runtime.GOOS != "windows" {
+		return "\u274c"
+	}
+	return "Error:"
+}
```
craicoverflow, line 75: Nice! However I think this should not be in the `build` package.

I see us adding some more emojis in the future so a package called `icon` in `pkg` for this would be ideal.

I also think we may want to add one-off emojis sometimes so add a function called `Emoji()` with a default value as the second argument. `Success()` and `Error()` functions can then use this method and you won't need to repeatedly check if runtime.GOOS == windows.

craicoverflow (changes requested):

craicoverflow:
> Closes #1024

This will not close this particular issue as it is adding the success icon to just one success message.

internal/build/build.go
alexal: @craicoverflow something like this?

`package icon

import "runtime"

func Emoji() bool {
	if runtime.GOOS != "windows" {
		return true
	}
	return false
}

// Success returns check mark emoji on every platform, but Windows
func Success() string {
	if Emoji() {
		return "\u2714\ufe0f"
	}
	return ""
}

// Error returns cross mark emoji on every platform, but Windows
func Error() string {
	if Emoji() {
		return "\u274c"
	}
	return "Error:"
}
`

craicoverflow: Yeah something like that, but make `Emoji()` function private and call it something else like `hasEmojiSupport`

craicoverflow: You should also have a generic function that you would use like this:

```go
prefix := icon.Emoji("\u274c", "Error:")

This function should be used by `Error()` and `Success()`

```

wtrocki:
Rebase needed.

alexal:
@wtrocki , @craicoverflow please see a new PR #1046 . Something is wrong when I do git rebase with main branch as it always bring other commits to my push request.

PR #1030, opened 2021-09-06
chore: remove i18n from use commands

wtrocki:
Let me know if we want that across entire fleet of commands


rkpattnaik780 (approved):

rkpattnaik780:
> Let me know if we want that across entire fleet of commands

I think we should do it across all commands.

PR #1033, opened 2021-09-06
chore(github): add a CODEOWNERS file

craicoverflow:
Added a CODEOWNERS file to assign reviewers automatically. 

We can customize this if needed, but I don't think we should assign individuals to specific areas.


craicoverflow:
If we organize i18n files by service - then we can assign the appropriate members from the docs team, but we are not ready for this quite yet.

PR #1037, opened 2021-09-06
feat(registry-artifacts): import export commands

wtrocki:
Verification:
```
rhoas service-registry artifact list > tmp.list
rhoas service-registry artifact export --output-file=export.zip

rhoas service-registry create --name=new-export
rhoas service-registry artifact import --file=export.zip
rhoas service-registry artifact list > tmp.list2

diff tmp.list tmp.list2 | wc

```


wtrocki:
Related PR(bug): https://github.com/redhat-developer/app-services-sdk-go/pull/272

wtrocki:
Ready for review

rkpattnaik780 (commented):
Import command doesn't seem to be working, imported a zip with 2 artifacts ( group = default ). Can't see the imported artifacts.

Screenshot from 2021-09-07 16-11-55

pkg/cmd/registry/artifact/migrate/import.go
```diff
@@ -0,0 +1,104 @@
...
+				return errors.New(opts.localizer.MustLocalize("artifact.cmd.common.error.noServiceRegistrySelected"))
+			}
+
+			opts.registryID = cfg.Services.ServiceRegistry.InstanceID
+			return runImport(opts)
+		},
+	}
+	cmd.Flags().StringVar(&opts.file, "file", "", opts.localizer.MustLocalize("artifact.common.file.location"))
+	cmd.Flags().StringVar(&opts.registryID, "instance-id", "", opts.localizer.MustLocalize("artifact.common.instance.id"))
+	_ = cmd.MarkFlagRequired("file")
```
rkpattnaik780, line 72: Marking `file` as required doesnt allow us to pass file path as argument

pkg/cmd/registry/artifact/migrate/export.go
```diff
@@ -0,0 +1,103 @@
...
+
+			if !cfg.HasServiceRegistry() {
+				return errors.New(opts.localizer.MustLocalize("artifact.cmd.common.error.noServiceRegistrySelected"))
+			}
+
+			opts.registryID = cfg.Services.ServiceRegistry.InstanceID
+			return runExport(opts)
+		},
+	}
+	cmd.Flags().StringVar(&opts.file, "file", "", opts.localizer.MustLocalize("artifact.common.file.location"))
```
rkpattnaik780, line 67: Should it be `--output-file`?
Context: #1008

rkpattnaik780 (changes requested):

pkg/cmd/registry/artifact/migrate/export.go
wtrocki: @craicoverflow what would be your take on this?

pkg/cmd/registry/artifact/migrate/import.go
wtrocki: Bad copy paste :)

pkg/cmd/registry/artifact/migrate/export.go
craicoverflow: If the contents from the `export` command being saved in the file, it should be `--output-file`.

I notice this is not using ['artifact.common.message.file.location'](https://github.com/redhat-developer/app-services-cli/blob/1165a3fde7562851355f55a1940e1e059a4f4ea4/pkg/localize/locales/en/cmd/artifact.en.toml#L103) like other output file flags are using.

craicoverflow: In saying that it is obvious from the context this time (`service-account create` is less clear). But we should be consistent.

rkpattnaik780 (approved):

wtrocki:
Import were identified as backend problem. Merging so others can try to fix that on their side.

PR #1042, opened 2021-09-07
fix: swap survey editor

wtrocki:
Swapping editor with survey editor. However implementation doesn't seem to work for me also it has couple usability quirks. 
I would recommend to stay with original editor if possible.


wtrocki:
CC @craicoverflow

PR #1046, opened 2021-09-08
feat: emoji support for both error and success messages

alexal:
Closes #1025
Closes #1024

PR #1047, opened 2021-09-08
fix: remove space after prefix from root error

craicoverflow:
Follow up to the great work done in #1046 

Three things are covered here:

- Renamed methods
- Removed the extra space between the Error prefix and message
- Detect if an emoji is used, and if so it capitalizes the first character if the message.

### Verification Steps

Windows:

```shell
$ rhoas kafka topic list
Error: not logged in to identity.api.openshift.com. 
```

Linux:

```shell
$ rhoas kafka topic list
❌ Not logged in to identity.api.openshift.com.
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/login/login.go
+++ b/pkg/cmd/login/login.go
@@ -202,9 +202,9 @@ func runLogin(opts *Options) (err error) {
 	opts.Logger.Info("")
 
 	if !ok {
-		opts.Logger.Info(icon.Success(), opts.localizer.MustLocalize("login.log.info.loginSuccessNoUsername"))
+		opts.Logger.Info(icon.SuccessPrefix(), opts.localizer.MustLocalize("login.log.info.loginSuccessNoUsername"))
 	} else {
-		opts.Logger.Info(icon.Success(), opts.localizer.MustLocalize("login.log.info.loginSuccess", localize.NewEntry("Username", username)))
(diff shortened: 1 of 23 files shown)


pkg/cmd/login/login.go
```diff
@@ -202,9 +202,9 @@ func runLogin(opts *Options) (err error) {
 	opts.Logger.Info("")
 
 	if !ok {
-		opts.Logger.Info(icon.Success(), opts.localizer.MustLocalize("login.log.info.loginSuccessNoUsername"))
+		opts.Logger.Info(icon.SuccessPrefix(), opts.localizer.MustLocalize("login.log.info.loginSuccessNoUsername"))
```
wtrocki, line 205: Just idea:  `opts.Logger.Success`

craicoverflow: Yeah, that might make things nicer actually, though I do like keeping the Logger lean and focused on log-levels only.

wtrocki: our logger is not logger per se. It is an stderr printer. Having more high-level methods will be good idea for future issue.

wtrocki (approved):

pkg/cmd/login/login.go
craicoverflow: Yeah, but the stderr printer is an implementation of a higher-level Logger interface, which could be implemented in different contexts. Though I do not see it being used in ways other than to stderr

PR #1048, opened 2021-09-08
chore: removed i18n from use commands

### LINKED ISSUE #1032: Replace use i18n commands names with explicit strings
Use will not change depending on the language - those are actual documented commands etc. 

See this PR for example

_Originally posted by @rkpattnaik780 in https://github.com/redhat-developer/app-services-cli/issues/1030#issuecomment-913572953_

alexal:
Closes #1032


craicoverflow:
The fact that the generated docs have not changed automatically verifies this one :)

PR #1050, opened 2021-09-08
feat(kafka): add ability to update Kafka instance

### LINKED ISSUE #1012: Add interactive mode for the `kafka update` command
The `kafka update` command needs an interactive mode to let users "select" the user they wish to set as the owner. This was not done in #844 because there are some changes pending on the API gateway to grant access to the rhoas-cli-prod client. 

As a workaround, we can do this by logging in using the token-based login, as the cloud-services token does have access to the required endpoint.

craicoverflow:
Closes #1012 

Verification is only possible by an org admin (@wtrocki).

### Verification Steps

1. Run `export RHOAS_DEV=true` to enable this hidden feature.
1. Log in using an offline token (currently only supported method)
2. Create a Kafka, or skip this if you already have one.
3. Run `rhoas kafka update`. You should get an interactive prompt to select a new owner. Select `sharedaccount_kafka_devexp`.
4. It should update the Kafka instance owner.
5. Run `rhoas kafka update --owner=<TAB>` and it should load the full list of available users to choose from.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed


wtrocki:
rebase needed?

craicoverflow:
rebased.

wtrocki:
Looking into this now..Another rebase needed :/

craicoverflow:
It has been rebased.

wtrocki:
I'm not sure about update as name. 

if we have command like `owner` or `owner-set` it might be more future proof. 
Like in future we might have requirements for project/env/labels/tags/name updates and this will make it really ackward. 
We technically not updating kafka - just changing ownership

craicoverflow:
> I'm not sure about update as name.
> 
> if we have command like `owner` or `owner-set` it might be more future proof.
> Like in future we might have requirements for project/env/labels/tags/name updates and this will make it really ackward.
> We technically not updating kafka - just changing ownership

This is a command which will be used to update various other configuration details in the future, it will not always exclusive to owner update.

wtrocki (approved):

Issue #1034, opened 2021-09-06
`--verbose/-v` flag is not working in commands with one subcommand


craicoverflow:
## Description

When you run a command which has a single subcommand, the `verbose` flag is ineffective.

Example: run `rhoas login -v`

### Expected vs actual behaviour

Expected behaviour:

```shell
$ rhoas login -v
Refreshing tokens
...
Logging in...
```

Actual behaviour:

```shell
$ rhoas login -v
Logging in...
```

rkpattnaik780:
Happens for other sub-commands as well, e.g not able to see `Starting interactive prompt` for topic and service account create operations.

alexal:
Please see my comments in #1051 but basically debutEnabled is always equals to false because we set it before add a verbose flag.

PR #1051, opened 2021-09-08
fix: added a Parse function for the command-line arguments

alexal:
This is a potential fix for #1034  . Seems like that we have to run a Parse() function, at least it started working in my local copy. @wtrocki , @craicoverflow what do you think?

### DIFF
--- a/pkg/cmd/root/root.go
+++ b/pkg/cmd/root/root.go
@@ -2,6 +2,7 @@ package root
 
 import (
 	"flag"
+	"os"
 
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/registry"
 
@@ -41,6 +42,7 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
 	cmd.Version = version
 
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	fs.Parse(os.Args[1:])
 
 	// Child commands
 	cmd.AddCommand(login.NewLoginCmd(f))
--- a/pkg/cmd/root/root.go
+++ b/pkg/cmd/root/root.go
@@ -42,7 +42,7 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
 	cmd.Version = version
 
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	fs.Parse(os.Args[1:])
+	_ = fs.Parse(os.Args[1:])
 
 	// Child commands
 	cmd.AddCommand(login.NewLoginCmd(f))


wtrocki:
Issue number if wrong :)

alexal:
> Issue number if wrong :)

Right, the correct one is #1034

pkg/cmd/root/root.go
```diff
@@ -41,6 +42,7 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
 	cmd.Version = version
 
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	_ = fs.Parse(os.Args[1:])
```
wtrocki, line 45: What this does?

wtrocki:
I cannot reproduce this:

- rhoas status  -v (works)
- rhoas kafka list  -v (works)
- rhoas service-registry artifact list -v (works)

The only command that doesn't work is login.

pkg/cmd/root/root.go
alexal: Parses a command line into the defined flags.

alexal:
> The only command that doesn't work is login.
I see, let me take a look.

wtrocki:
No need to close. I thinkl this is the reason: https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/debug/flag.go#L8:L8

wtrocki:
SOrry wrong link. I meant this: https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/factory/default.go#L34:L34 returning false for login but ok for the rest

alexal:
> https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/factory/default.go#L34:L34
What is interesting, it is also false for a status command. So, most likely it was changed somewhere later in the stack.
image

wtrocki:
Yes.. it is the same. It is never enabled appently. Very strange. My take is that we should not enable this in factory, but later after flag is setup and can be detected.

alexal:
This piece of code

c.logger.Debug("Refreshing tokens")

Gets a logger from KeycloakConnection, not from the status command. And for KeycloakConnection logger, debugEnabled is set to true:

image

wtrocki:
That is it. you cracked system

alexal:
@wtrocki ,

I know where is the problem. This line in default.go

debugEnabled := debug.Enabled()

is called before these lines in root.go

arguments.AddDebugFlag(fs)
...
pflag.CommandLine.AddGoFlagSet(flag.CommandLine)

This is why debugEnabled will always be set to false.

wtrocki:
exactly. Moving this after will fix jt

PR #1055, opened 2021-09-09
feat: added an option to show bootstrap server host of the Kafka instance

### LINKED ISSUE #636: Add option to show the bootstrapServer

Add an option for the command to return the bootstrapServer hostname of the instance in use

### Proposed solution

rhoas kafka cluster describe --bootstrapserver

### Additional context

When getting the information through the CLI it can be easy to just return the value without the need to parse the json returned.

alexal:
Closes #636

### DIFF
--- a/pkg/cmd/kafka/describe/describe.go
+++ b/pkg/cmd/kafka/describe/describe.go
@@ -25,9 +25,10 @@ import (
 )
 
 type Options struct {
-	id           string
-	name         string
-	outputFormat string
+	id              string
+	name            string
+	bootstrapServer bool
+	outputFormat    string
 
 	IO         *iostreams.IOStreams
 	Config     config.IConfig
@@ -88,6 +89,7 @@ func NewDescribeCommand(f *factory.Factory) *cobra.Command {
 	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("kafka.common.flag.output.description"))
 	cmd.Flags().StringVar(&opts.id, "id", "", opts.localizer.MustLocalize("kafka.describe.flag.id"))
 	cmd.Flags().StringVar(&opts.name, "name", "", opts.localizer.MustLocalize("kafka.describe.flag.name"))
+	cmd.Flags().BoolVar(&opts.bootstrapServer, "bootstrapserver", false, opts.localizer.MustLocalize("kafka.describe.flag.bootstrapserver"))
 
 	if err := kafkacmdutil.RegisterNameFlagCompletionFunc(cmd, f); err != nil {
 		opts.Logger.Debug(opts.localizer.MustLocalize("kafka.common.error.load.completions.name.flag"), err)
@@ -125,7 +127,11 @@ func runDescribe(opts *Options) error {
 		}
 	}
 
-	return printKafka(kafkaInstance, opts)
+	if !opts.bootstrapServer {
+		return printKafka(kafkaInstance, opts)
+	}
+	opts.Logger.Info(*kafkaInstance.BootstrapServerHost)
+	return nil
 }
 
 func printKafka(kafka *kafkamgmtclient.KafkaRequest, opts *Options) error {
--- a/pkg/cmd/kafka/describe/describe.go
+++ b/pkg/cmd/kafka/describe/describe.go
@@ -128,7 +128,7 @@ func runDescribe(opts *options) error {
 	}
 
 	if opts.bootstrapServer {
-		fmt.Fprintf(opts.IO.Out, kafkaInstance.GetBootstrapServerHost())
+		fmt.Fprintln(opts.IO.Out, kafkaInstance.GetBootstrapServerHost())
 		return nil
 	}
 
--- a/pkg/cmd/kafka/describe/describe.go
+++ b/pkg/cmd/kafka/describe/describe.go
@@ -128,7 +128,11 @@ func runDescribe(opts *options) error {
 	}
 
 	if opts.bootstrapServer {
-		fmt.Fprintln(opts.IO.Out, kafkaInstance.GetBootstrapServerHost())
+		if host, ok := kafkaInstance.GetBootstrapServerHostOk(); ok {
+			fmt.Fprintln(opts.IO.Out, *host)
+			return nil
+		}
+		opts.Logger.Info(opts.localizer.MustLocalize("kafka.describe.bootstrapserver.not.available"))
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/kafka/describe/describe.go
```diff
@@ -88,6 +89,7 @@ func NewDescribeCommand(f *factory.Factory) *cobra.Command {
 	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("kafka.common.flag.output.description"))
 	cmd.Flags().StringVar(&opts.id, "id", "", opts.localizer.MustLocalize("kafka.describe.flag.id"))
 	cmd.Flags().StringVar(&opts.name, "name", "", opts.localizer.MustLocalize("kafka.describe.flag.name"))
+	cmd.Flags().BoolVar(&opts.bootstrapServer, "bootstrapserver", false, opts.localizer.MustLocalize("kafka.describe.flag.bootstrapserver"))
```
craicoverflow, line 92:
```suggestion
	cmd.Flags().BoolVar(&opts.bootstrapServer, "bootstrap-server", false, opts.localizer.MustLocalize("kafka.describe.flag.bootstrapserver"))
```

pkg/cmd/kafka/describe/describe.go
```diff
@@ -125,7 +127,11 @@ func runDescribe(opts *Options) error {
 		}
 	}
 
-	return printKafka(kafkaInstance, opts)
+	if !opts.bootstrapServer {
+		return printKafka(kafkaInstance, opts)
+	}
+	opts.Logger.Info(*kafkaInstance.BootstrapServerHost)
```
craicoverflow, line 133: It needs to print to stdout.

The logger only prints to stderr, which is not machine-readable and so is not useful for this.

pkg/cmd/kafka/describe/describe.go
```diff
@@ -125,7 +127,11 @@ func runDescribe(opts *Options) error {
 		}
 	}
 
-	return printKafka(kafkaInstance, opts)
+	if !opts.bootstrapServer {
```
craicoverflow, line 130: I'd rather see `if opts.bootstrapServer` and print the url and return, leaving the entire object as the default.

craicoverflow (changes requested):
A couple of changes needed, but nice job 👍🏻

alexal:
> A couple of changes needed, but nice job 👍🏻

Thanks for the feedback. I've made the changes, please take a look.

pkg/cmd/kafka/describe/describe.go
```diff
@@ -125,6 +129,11 @@ func runDescribe(opts *Options) error {
 		}
 	}
 
+	if opts.bootstrapServer {
+		fmt.Fprintf(os.Stdout, "%v", *kafkaInstance.BootstrapServerHost)
```
craicoverflow, line 133:
```suggestion
		fmt.Println(os.Stdout, kafkaInstance.GetBootstrapServerHost())
```

^ Untested, but this is preferred. `IOStreams` is a an abstraction of stderr and stdout which we use instead.

craicoverflow: It is possible for the bootstrap server URL to be nil. We should handle this with a message.

alexal: Done, but we still have to use fmt.Fprintf function
`fmt.Fprintf(opts.IO.Out, kafkaInstance.GetBootstrapServerHost())`

wtrocki (approved):
Thank you so much for contribution.

pkg/cmd/kafka/describe/describe.go
```diff
@@ -125,6 +127,11 @@ func runDescribe(opts *options) error {
 		}
 	}
 
+	if opts.bootstrapServer {
+		fmt.Fprintln(opts.IO.Out, kafkaInstance.GetBootstrapServerHost())
+		return nil
+	}
```
craicoverflow, line 133: I think we should be checking if the bootstrap_server_host exists, and if it does not print some info message letting the user know this.

wtrocki: Agree. We can extract this code into helper: 
https://github.com/redhat-developer/app-services-cli/blob/main/pkg/connection/keycloak_connection.go#L254-L267

alexal: I've added a message, please take a look.

pkg/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -402,6 +402,13 @@ one = 'Unique ID of the Kafka instance you want to view'
 description = 'Description for the --name flag'
 one = 'Name of the Kafka instance you want to view'
 
+[kafka.describe.flag.bootstrapserver]
```
wtrocki, line 405: In long run we want to have ability to get some specific config of the services in unified fassion. 

It is good to have this in describe for now, but we should have similar capability for the status command

wtrocki:
LGTM

wtrocki:
Reverified. Good job!

pkg/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -406,6 +406,9 @@ one = 'Name of the Kafka instance you want to view'
 description = 'Description for the --bootstrap-server flag'
 one = 'If specified, only bootstrap server host of the Kafka instance will be displayed'
 
+[kafka.describe.bootstrapserver.not.available]
+one = 'Unable to get bootstrap server host name'
```
wtrocki, line 410: get Kafka bootstrapServerHost

wtrocki (changes requested):
Small message fix. Works good

pkg/localize/locales/en/cmd/kafka.en.toml
alexal: Looks better now?

pkg/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -402,6 +402,13 @@ one = 'Unique ID of the Kafka instance you want to view'
 description = 'Description for the --name flag'
 one = 'Name of the Kafka instance you want to view'
 
+[kafka.describe.flag.bootstrapserver]
+description = 'Description for the --bootstrap-server flag'
+one = 'If specified, only bootstrap server host of the Kafka instance will be displayed'
+
+[kafka.describe.bootstrapserver.not.available]
+one = 'Unable to get bootstrap server host name'
```
craicoverflow:
```suggestion
one = 'Kafka instance "{{.Name}}" does not have a bootstrap server URL.'
```

usually we give more specific information. It is not that we are unable, it does not exist.

alexal: @craicoverflow done, please take a look.

PR #1059, opened 2021-09-10
chore: made all Options in commands private

### LINKED ISSUE #1058: Make all `Options` in commands private
The Options type in every command is public. This was originally done at the beginning with the anticipation is being used in tests, but since there is no need anymore, we can make all options private now.

alexal:
All of them are private now, except pkg/status/status.go.

Closes #1058

### DIFF
--- a/pkg/cmd/cluster/bind/bind.go
+++ b/pkg/cmd/cluster/bind/bind.go
@@ -15,7 +15,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type Options struct {
+type args struct {
 	Config     config.IConfig
 	Connection func(connectionCfg *connection.Config) (connection.Connection, error)
 	Logger     logging.Logger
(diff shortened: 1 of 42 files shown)


pkg/cmd/cluster/bind/bind.go
```diff
@@ -73,7 +73,7 @@ func NewBindCommand(f *factory.Factory) *cobra.Command {
 	return cmd
 }
 
-func runBind(opts *Options) error {
+func runBind(opts *args) error {
```
craicoverflow, line 76: I know there was a discussion around the new name but I don't think we fully agreed on it (myself and @wtrocki coming from different view points). I think the opts variable name could be confusing with an `args` type. On consideration I think it is best to leave it as `options` for now, we can always revisit later but it causes the least disruption.

craicoverflow (commented):
See comment re: leaving it as `options` for the time being.

alexal:
> See comment re: leaving it as `options` for the time being.

Done.

craicoverflow (approved):
Thank you! 👍🏻 💯

PR #1061, opened 2021-09-10
fix: fixed the problem with a verbose flag

### LINKED ISSUE #1034: `--verbose/-v` flag is not working in commands with one subcommand

alexal:
I've added a new SetDebug() function to the Logger interface and implemented it in the StdLoggerBuilder. Also, I've moved debug.Enabled() out of default.go file to the main.go and placed it after the command line arguments definition, so the verbose flag will be properly read and set.

Closes #1034

### DIFF
--- a/pkg/cmd/root/root.go
+++ b/pkg/cmd/root/root.go
@@ -2,6 +2,7 @@ package root
 
 import (
 	"flag"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/debug"
 
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/registry"
 
@@ -29,6 +30,9 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
 		Short:         f.Localizer.MustLocalize("root.cmd.shortDescription"),
 		Long:          f.Localizer.MustLocalize("root.cmd.longDescription"),
 		Example:       f.Localizer.MustLocalize("root.cmd.example"),
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			f.Logger.SetDebug(debug.Enabled())
+		},
 	}
 	fs := cmd.PersistentFlags()
 	arguments.AddDebugFlag(fs)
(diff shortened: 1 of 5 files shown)


alexal:
Now, the debug is working properly
image

pkg/cmd/root/root.go
```diff
@@ -29,6 +30,9 @@ func NewRootCommand(f *factory.Factory, version string) *cobra.Command {
 		Short:         f.Localizer.MustLocalize("root.cmd.shortDescription"),
 		Long:          f.Localizer.MustLocalize("root.cmd.longDescription"),
 		Example:       f.Localizer.MustLocalize("root.cmd.example"),
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			f.Logger.SetDebug(debug.Enabled())
```
wtrocki, line 34: That will run for some meta commands like __completion

wtrocki: I would set this as part of the run code

wtrocki (approved):
Amazing work! Thank you so much

pkg/cmd/root/root.go
alexal: > I would set this as part of the run code

Do you recommend any particular place for that? I can check if debug is reported as enabled there.

wtrocki: After line 39 in the same file seems more cleaner

wtrocki: After arguments.AddDebugFlag(fs)

alexal: > After arguments.AddDebugFlag(fs)

Unfortunately debug.Enabled() returns false there...

image

wtrocki:
I have tried to read -v flag in root and avoid running things in the command context but without success. This seems to be the best option

Issue #1057, opened 2021-09-10
rhoas kafka create - 429 Too Many Requests


alexal:
Right now, if you have created one Kafka instance and try to create another one, you will get the following error

```
/private/var/folders/s6/dqhzqm7x2gb8vlvskw4jqj380000gn/T/GoLand/___rhoas_kafka_create kafka create
? Name: test1
? Cloud Provider: aws
? Cloud Region: us-east-1
❌ 429 Too Many Requests. Run the command in verbose mode using the -v flag to see more information
```

With -v flag enabled, you will see the actual reason of 429 error

`{"id":"24","kind":"Error","href":"/api/kafkas_mgmt/v1/errors/24","code":"KAFKAS-MGMT-24","reason":"only one eval instance is allowed","operation_id":"c4tal4oc3kjol9oa6a0g"}`

However, my question is, should we improve the error message?

wtrocki:
we should

craicoverflow:
Yes. Myself and @wtrocki were discussing options on how to reliably check the meaning of the error codes from the backend.

I think in the meantime though we should print the error "reason" from the backend.

wtrocki:
@craicoverflow I do that already in registry (using helper to cast to APIError). Plan was to also do the same for kafka. +1 from me

craicoverflow:
Agreed.

craicoverflow:
> @craicoverflow I do that already in registry (using helper to cast to APIError). Plan was to also do the same for kafka. +1 from me

I'm seeing some cases where this is not being done. Maybe it is not applied everywhere?

```shell
❯ ./rhoas service-registry create --name enda-dev
Creating Service Registry with name: enda-dev
Error: 500 Internal Server Error. Run the command in verbose mode using the -v flag to see more information
```

wtrocki:
Yes.. Registry artifacts only for now

alexal:
@craicoverflow, @wtrocki I would love to work on this problem.

craicoverflow:
Sure thing @alexal. It is a good idea to suggest a couple of ways to address this problem as the way to implement it is not trivial.

We can converse on this issue until a collective agreement is made. This can save you time in the long run and help build a better solution :)

wtrocki:
There is couple problems here:

- SDK should have common APIError object that can be used to cast error coming from API. 
This way we can provide generic method to read reason

[...]

PR #1064, opened 2021-09-13
feat: improved error reporting for Kafka create command

alexal:
This is a potential fix for #1057 . @wtrocki , @craicoverflow let me know if there are other ways to implement the error handling for this specific situation.

### DIFF
--- a/pkg/cmd/kafka/kafkaerror/error.go
+++ b/pkg/cmd/kafka/kafkaerror/error.go
@@ -1,0 +1,34 @@
+package kafkaerror
+
+import (
+	"encoding/json"
+	"errors"
+	"github.com/redhat-developer/app-services-cli/pkg/logging"
+	kafkamgmtclient "github.com/redhat-developer/app-services-sdk-go/kafkamgmt/apiv1/client"
+)
+
+// GetAPIError gets a strongly typed error from an error
+func GetAPIError(err error, logger logging.Logger) (e kafkamgmtclient.Error, ok bool) {
+	var apiError kafkamgmtclient.GenericOpenAPIError
+	var kafkaError kafkamgmtclient.Error
+
+	if ok = errors.As(err, &apiError); ok {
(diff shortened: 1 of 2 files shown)


pkg/cmd/kafka/kafkaerror/error.go
```diff
@@ -0,0 +1,34 @@
+package kafkaerror
```
craicoverflow, line 1: This does not belong in pkg/cmd (should be reserved for commands). Check in `pkg/api`

craicoverflow:
This is a problem which requires a global fix as opposed to just one command. let's chat about it in the issue.

pkg/cmd/kafka/kafkaerror/error.go
alexal: I've moved it there.

pkg/api/kafka/error.go
```diff
@@ -0,0 +1,34 @@
...
+// GetAPIError gets a strongly typed error from an error
+func GetAPIError(err error, logger logging.Logger) (e kafkamgmtclient.Error, ok bool) {
+	var apiError kafkamgmtclient.GenericOpenAPIError
+	var kafkaError kafkamgmtclient.Error
+
+	if ok = errors.As(err, &apiError); ok {
+		kafkaError = kafkamgmtclient.Error{}
+		err = json.Unmarshal(apiError.Body(), &kafkaError)
+		if err != nil {
+			logger.Error(err)
```
craicoverflow, line 19: We don't want to log the error here.

wtrocki:
@alexal I think we need to chat with the team about this as we need a good strategy for SDK's and CLI. I will put this PR on hold as we need more time to discuss that.

alexal:
@wtrocki sure, keep me posted.

craicoverflow:
Closing this for now as there is no current solution.

PR #1066, opened 2021-09-14
refactor: added dump print helper to all commands

### LINKED ISSUE #905: Add dump print helper to all commands.

Currently we are replicating a lot of code across the commands to print based on the output format. 
We should use helper in the dump package (available currently in the PR #859)

alexal:
Closes #905

### DIFF
--- a/pkg/cmd/kafka/consumergroup/describe/describe.go
+++ b/pkg/cmd/kafka/consumergroup/describe/describe.go
@@ -143,7 +143,7 @@ func runCmd(opts *options) error {
 	stdout := opts.IO.Out
 
 	switch opts.outputFormat {
-	case "table":
+	case dump.TableFormat:
 		printConsumerGroupDetails(stdout, consumerGroupData, opts.localizer)
 	default:
 		dump.PrintDataInFormat(opts.outputFormat, consumerGroupData, stdout)
--- a/pkg/cmd/kafka/consumergroup/describe/describe.go
+++ b/pkg/cmd/kafka/consumergroup/describe/describe.go
@@ -143,10 +143,10 @@ func runCmd(opts *options) error {
 	stdout := opts.IO.Out
 
 	switch opts.outputFormat {
-	case dump.JSONFormat, dump.YAMLFormat, dump.YMLFormat:
-		dump.PrintDataInFormat(opts.outputFormat, consumerGroupData, stdout)
-	default:
+	case "table":
 		printConsumerGroupDetails(stdout, consumerGroupData, opts.localizer)
+	default:
+		dump.PrintDataInFormat(opts.outputFormat, consumerGroupData, stdout)
 	}
 
 	return nil
--- a/pkg/cmd/kafka/consumergroup/describe/describe.go
+++ b/pkg/cmd/kafka/consumergroup/describe/describe.go
@@ -3,14 +3,20 @@ package describe
 import (
 	"context"
 	"errors"
+	"fmt"
 	"github.com/redhat-developer/app-services-cli/pkg/cmdutil"
+	cgutil "github.com/redhat-developer/app-services-cli/pkg/kafka/consumergroup"
 	"github.com/redhat-developer/app-services-cli/pkg/localize"
+	kafkainstanceclient "github.com/redhat-developer/app-services-sdk-go/kafkainstance/apiv1internal/client"
+	"io"
 	"net/http"
+	"sort"
 
 	"github.com/redhat-developer/app-services-cli/internal/config"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/factory"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/flag"
 	flagutil "github.com/redhat-developer/app-services-cli/pkg/cmdutil/flags"
+	"github.com/redhat-developer/app-services-cli/pkg/color"
 	"github.com/redhat-developer/app-services-cli/pkg/connection"
 	"github.com/redhat-developer/app-services-cli/pkg/dump"
 	"github.com/redhat-developer/app-services-cli/pkg/iostreams"
@@ -27,6 +33,15 @@ type options struct {
 	Connection factory.ConnectionFunc
 	localizer  localize.Localizer
 	Context    context.Context
+}
+
+type consumerRow struct {
+	MemberID      string `json:"memberId,omitempty" header:"Consumer ID"`
+	Partition     int    `json:"partition,omitempty" header:"Partition"`
+	Topic         string `json:"topic,omitempty" header:"Topic"`
+	LogEndOffset  int    `json:"logEndOffset,omitempty" header:"Log end offset"`
+	CurrentOffset int    `json:"offset,omitempty" header:"Current offset"`
+	OffsetLag     int    `json:"lag,omitempty" header:"Offset lag"`
 }
 
 // NewDescribeConsumerGroupCommand gets a new command for describing a consumer group.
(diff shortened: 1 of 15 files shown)


pkg/cmd/serviceaccount/list/list.go
```diff
@@ -90,36 +75,6 @@ func runList(opts *options) (err error) {
...
-			ClientID:  sa.GetClientId(),
-			Owner:     sa.GetOwner(),
-			CreatedAt: sa.GetCreatedAt().String(),
-		}
-
-		rows = append(rows, row)
-	}
-
-	return rows
-}
```
craicoverflow, line 125: What is the reason for deleting this?

craicoverflow (changes requested):
A lot of the `list` commands print in table format, which is not covered by the PrintDataInFormat function. A switch statement is still needed in these cases.

pkg/cmd/serviceaccount/list/list.go
alexal: I thought the only acceptable formats were specified in ValidOutputFormats? I see nothing there about table format.

`ValidOutputFormats       = []string{dump.JSONFormat, dump.YAMLFormat, dump.YMLFormat}`

craicoverflow: Table format is used when there is no format passed in the list commands (default). There is no way so specify it using the output format.

alexal: @craicoverflow done

pkg/cmd/kafka/consumergroup/describe/describe.go
```diff
@@ -144,13 +141,10 @@ func runCmd(opts *options) error {
...
 	stdout := opts.IO.Out
+
 	switch opts.outputFormat {
-	case dump.JSONFormat:
-		data, _ := json.Marshal(consumerGroupData)
-		_ = dump.JSON(stdout, data)
-	case dump.YAMLFormat, dump.YMLFormat:
-		data, _ := yaml.Marshal(consumerGroupData)
-		_ = dump.YAML(stdout, data)
+	case dump.JSONFormat, dump.YAMLFormat, dump.YMLFormat:
```
wtrocki, line 146: Switch with 1 case?

pkg/cmd/kafka/consumergroup/list/list.go
```diff
@@ -162,19 +159,13 @@ func runList(opts *options) (err error) {
...
 	switch opts.output {
-	case dump.JSONFormat:
-		data, _ := json.Marshal(consumerGroupData)
-		_ = dump.JSON(opts.IO.Out, data)
-	case dump.YAMLFormat, dump.YMLFormat:
-		data, _ := yaml.Marshal(consumerGroupData)
-		_ = dump.YAML(opts.IO.Out, data)
+	case dump.JSONFormat, dump.YAMLFormat, dump.YMLFormat:
+		dump.PrintDataInFormat(opts.output, consumerGroupData, opts.IO.Out)
 	default:
```
wtrocki, line 164: This is a very convoluted way of simple
```
if opts.output == "table" {
  rows := mapConsumerGroupResultsToTableFormat(consumerGroups)}
  dump.Table(opts.IO.Out, rows)
} else {
   dump.PrintDataInFormat(opts.output, consumerGroupData, opts.IO.Out)
}
```

wtrocki (changes requested):
Small fix required

craicoverflow:
This is a refactor, not a chore (I recommend reading up on conventional commits).

pkg/cmd/kafka/consumergroup/list/list.go
craicoverflow: Actually in Go, if/else statements are uncommon, and generally switch statements are recommended over if/else: https://golang.org/doc/effective_go#switch

We can still use a switch, but swap the order to mimic the if/else example by @wtrocki

pkg/cmd/kafka/consumergroup/describe/describe.go
craicoverflow: We can still use switch, but make it simpler. See https://github.com/redhat-developer/app-services-cli/pull/1066#discussion_r710011053

alexal: Done

pkg/cmd/kafka/consumergroup/list/list.go
alexal: Done

alexal:
> Small fix required

Done, please review.

pkg/cmd/kafka/list/list.go
```diff
@@ -127,18 +123,13 @@ func runList(opts *options) error {
...
 
 	switch opts.outputFormat {
-	case dump.JSONFormat:
-		data, _ := json.Marshal(response)
-		_ = dump.JSON(opts.IO.Out, data)
-	case dump.YAMLFormat, dump.YMLFormat:
-		data, _ := yaml.Marshal(response)
-		_ = dump.YAML(opts.IO.Out, data)
-	default:
+	case "table":
```
wtrocki, line 126: I would define this as constat in dump: https://github.com/redhat-developer/app-services-cli/blob/roles/pkg/dump/dump.go#L23:L23

wtrocki:
Sounds good

pkg/cmd/kafka/list/list.go
alexal: Sure, give me a sec.

alexal: Done

wtrocki (approved):

pkg/cmd/kafka/list/list.go
craicoverflow: The last time I checked, "table" is	even a possible value for `--output`. Has it been added?

wtrocki: table should be default

craicoverflow: "table" is not currently a valid output format. Currently an empty string maps to the output table format.

https://github.com/redhat-developer/app-services-cli/blob/e34a650f24a6f291ec375a936f9431060f9ce5b6/pkg/cmdutil/flags/flags.go#L10

Which I think is fine, because there will be no circumstance where `--output=table` is ever needed.

craicoverflow (changes requested):
```shell
~/c/gith/redhat-developer/app-services-cli fix/905 *26 ❯ ./rhoas kafka list --limit 1                                                 ✘ INT 13:45:07
{
  "items": [
    {
      "bootstrap_server_host": "ajay-test-c--p-u-iso-cjc-mv-fg.bf2.kafka-stage.rhcloud.com:443",
      "cloud_provider": "aws",
      "created_at": "2021-09-15T06:36:41.547162Z",
      "href": "/api/kafkas_mgmt/v1/kafkas/c50p8u8iso8cjc1mv8fg",
      "id": "c50p8u8iso8cjc1mv8fg",
      "instance_type": "eval",
      "kind": "Kafka",
      "multi_az": true,
      "name": "ajay-test",
      "owner": "apratap_kafka_devexp",
      "region": "us-east-1",
      "status": "ready",
      "updated_at": "2021-09-16T12:44:49.060074Z",
      "version": "2.7.0"
    }
  ],
  "kind": "KafkaRequestList",
  "page": 1,
  "size": 1,
  "total": 8
}
```

With this change, list commands are defaulting to JSON. See my comments regarding the "table" format - there is no table format, it should be "".

pkg/cmd/kafka/list/list.go
alexal: Yes, I've mentioned that in this comment https://github.com/redhat-developer/app-services-cli/pull/1066#discussion_r708358320. Do you want me to change TableFormat to an empty string?

alexal:
@craicoverflow, looks better with an empty string

```
/private/var/folders/zb/5rqszm5s2yj47f8sg4bbxpsw0000gn/T/GoLand/___kafka_list___limit_1 kafka list --limit 1
  ID                     NAME     OWNER                         STATUS   CLOUD PROVIDER   REGION     
 ---------------------- -------- ----------------------------- -------- ---------------- ----------- 
  c51iiefgofi8i86nrupg   alexal   not@for.all   ready    aws              us-east-1  
```

wtrocki:
@alexal As you see not even maintainers sometimes agree on formats sometimes. To prevent you from being a victim of that I would follow what @craicoverflow is suggesting.

alexal:
> @alexal As you see not even maintainers sometimes agree on formats sometimes. To prevent you from being a victim of that I would follow what @craicoverflow is suggesting.

I've renamed dump.TableFormat to dump.EmptyFormat and made it empty string.

craicoverflow:
Sorry for the confusion we might be causing @alexal 😄 The current way it is done is with an empty string. @wtrocki if you think there is good reason to change this would you mind opening an issue?

alexal:
> Sorry for the confusion we might be causing

No problem

craicoverflow (approved):

PR #1068, opened 2021-09-14
Service Registry access rules

wtrocki:
Still WIP but functional. Creating for feedback/binary across teams

## Plan 

 

Then we need specific commands for  registry to manage rbac

```bash
## No difference for us if we create or update roles
rhoas service-registry role add --user=test --role=WRITE
rhoas service-registry role revoke --user=test
rhoas service-registry role list 
```

We might look for the ability to create a new service account with roles 
For simplicity, we can even allow creating SA and role assignment in a single command

```bash
rhoas service-registry role add --new --role=WRITE
```

For simplicity, we can give the user the ability to set a role for himself without specifying the name

```bash
rhoas service-registry role add --me --role=WRITE
```

> NOTE The creation of the service account will give us only read access to topics, consumer groups etc.
Additional actions will be needed for CLI, Operator to supply extra roles for the service account.


pkg/cmd/registry/artifact/crud/create/create.go
```diff
@@ -73,6 +73,12 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 				opts.file = args[0]
 			}
 
+			if opts.artifactType != "" {
+				if _, err := registryinstanceclient.NewArtifactTypeFromValue(opts.artifactType); err != nil {
+					return errors.New(opts.localizer.MustLocalize("artifact.cmd.create.error.invalidArtifactType", localize.NewEntry("AllowedTypes", util.GetAllowedArtifactTypeEnumValuesAsString())))
```
craicoverflow, line 78:
```suggestion
					return errors.New(opts.localizer.MustLocalize("artifact.cmd.create.error.invalidArtifactType", localize.NewEntry("AllowedTypes", util.GetAllowedArtifactTypeEnumValuesAsString())))
```
```suggestion
					return opts.localizer.MustLocalizeError("artifact.cmd.create.error.invalidArtifactType", localize.NewEntry("AllowedTypes", util.GetAllowedArtifactTypeEnumValuesAsString()))
```

wtrocki: Good catch. Will fix that in couple places.

wtrocki:
Blocked because API seems to be not there yet - also we are missing ways to fetch principals and validate them.
Leaving as blocked for now

wtrocki:
Created new issue to cover blocking parts. This way we can merge this changes and resolve issue later phase:
https://github.com/redhat-developer/app-services-cli/issues/1095

wtrocki:
I would leave org list command for the moment. The reason why this was blocked were lack of ability to list user principals. 
Api will improve over the time and we can fix that properly.

pkg/cmd/registry/artifact/role/list/list.go
```diff
@@ -0,0 +1,145 @@
...
+		IO:         f.IOStreams,
+		localizer:  f.Localizer,
+		Context:    f.Context,
+	}
+
+	cmd := &cobra.Command{
+		Use:     "list",
+		Short:   f.Localizer.MustLocalize("registry.role.cmd.list.shortDescription"),
+		Long:    f.Localizer.MustLocalize("registry.role.cmd.list.longDescription"),
+		Example: f.Localizer.MustLocalize("registry.role.cmd.list.example"),
```
rkpattnaik780, line 59:
```suggestion
		Example: f.Localizer.MustLocalize("registry.role.cmd.list.example"),
		Args:    cobra.NoArgs,
```

pkg/localize/locales/en/cmd/role.toml
```diff
@@ -0,0 +1,104 @@
...
+When revoking role of the owner of the instance user will still have ADMIN rights.
+'''
+ 
+[registry.role.cmd.revoke.example] 
+one = '''
+## Revoke role for user
+rhoas service-registry role revoke --username=janedough
+''' 
+
+[registry.role.cmd.revoke.success]
```
rkpattnaik780, line 101: Nit: should we elaborate this message?

rkpattnaik780 (changes requested):
Role list command doesnt work for json and yaml output formats.
Screenshot from 2021-09-22 14-26-11

pkg/localize/locales/en/cmd/role.toml
```diff
@@ -0,0 +1,104 @@
...
+'''
+ 
+[registry.role.cmd.revoke.example] 
+one = '''
+## Revoke role for user
+rhoas service-registry role revoke --username=janedough
+''' 
+
+[registry.role.cmd.revoke.success] 
+one = 'Success'
```
wtrocki, line 102:
```suggestion
one = 'Successfully revoked access for specified principal' 
```

wtrocki:
> Role list command doesnt work for json and yaml output formats.

This must have been regression in other PR. I will also check how other lists commands work now.

wtrocki:
Actually other commands work. it is that I have old version of the code that weren't changed

wtrocki:
Merging to create srs release for tommorow

PR #1077, opened 2021-09-16
Service binding supporting deployment config

wtrocki:
## What this does

We are adding deploymentConfig support by extra flag.

Why flag instead of showing both?
Deployment config is typically problematic with binding (extra rollouts are needed and we cannot make sure that binding will be valid). We needed to provide extra documentation etc. to support that - adding it by default will cause confusion and problems. 
It will also break our existing guides :/

## Notes

This change doesn't introduce any multiservice aritecture. 
This is going to be done in separate feature branch done by @rkpattnaik780 
I have prepared layer for that refactoring but we need to work on top of the changes that Rama has right now.

## Verification

```
[wtrocki@graphapi app-services-cli (support-bind ✗)]$ rhoas cluster bind --deployment-config 
? Select Kafka instance to connect: pete
Namespace not provided. Using wtrocki namespace
? Please select application you want to connect with example
Binding "pete" with "example" app
? Do you want to continue? Yes
Using ServiceBinding Operator to perform binding
✔️ Binding pete with example app succeeded
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/redhat-developer/app-services-sdk-go/kafkamgmt v0.4.0
 	github.com/redhat-developer/app-services-sdk-go/registryinstance v0.3.1
 	github.com/redhat-developer/app-services-sdk-go/registrymgmt v0.2.0
-	github.com/redhat-developer/service-binding-operator v0.8.0
+	github.com/redhat-developer/service-binding-operator v0.9.0
 	github.com/spf13/cobra v1.2.1
 	github.com/spf13/pflag v1.0.5
 	gitlab.com/c0b/go-ordered-json v0.0.0-20201030195603-febf46534d5a
@@ -37,5 +37,5 @@ require (
 	k8s.io/api v0.22.1
 	k8s.io/apimachinery v0.22.1
 	k8s.io/client-go v0.22.1
-	sigs.k8s.io/controller-runtime v0.10.0
+	sigs.k8s.io/controller-runtime v0.10.0 // indirect
 )
--- a/pkg/cluster/serviceBinding.go
+++ b/pkg/cluster/serviceBinding.go
@@ -3,19 +3,17 @@ package cluster
 import (
 	"context"
 	"crypto/rand"
-	"errors"
 	"fmt"
-	"github.com/redhat-developer/app-services-cli/pkg/icon"
 	"os"
 	"path/filepath"
+
+	"github.com/redhat-developer/app-services-cli/pkg/icon"
+	"github.com/redhat-developer/service-binding-operator/apis/binding/v1alpha1"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/redhat-developer/app-services-cli/pkg/color"
 	"github.com/redhat-developer/app-services-cli/pkg/localize"
 	"github.com/redhat-developer/app-services-cli/pkg/logging"
-	"github.com/redhat-developer/service-binding-operator/api/v1alpha1"
-	"github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline/builder"
-	sboContext "github.com/redhat-developer/service-binding-operator/pkg/reconcile/pipeline/context"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -24,7 +22,6 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
-	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 type KubernetesClients struct {
@@ -33,17 +30,19 @@ type KubernetesClients struct {
 	clientConfig  *clientcmd.ClientConfig
 }
 
-var deploymentResource = schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}
+var (
+	deploymentResource       = schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}
+	deploymentConfigResource = schema.GroupVersionResource{Group: "apps.openshift.io", Version: "v1", Resource: "deploymentconfigs"}
+)
 
 type ServiceBindingOptions struct {
 	ServiceName             string
 	Namespace               string
 	AppName                 string
 	ForceCreationWithoutAsk bool
-	ForceUseOperator        bool
-	ForceUseSDK             bool
 	BindingName             string
 	BindAsFiles             bool
+	DeploymentConfig        bool
 }
 
 func ExecuteServiceBinding(ctx context.Context, logger logging.Logger, localizer localize.Localizer, options *ServiceBindingOptions) error {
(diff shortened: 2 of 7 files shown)


wtrocki:
WIP. FYI @rkpattnaik780

wtrocki:
@craicoverflow just review needed. Verification not possible - SBO operator seems to be broken

go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/redhat-developer/app-services-sdk-go/kafkamgmt v0.4.0
 	github.com/redhat-developer/app-services-sdk-go/registryinstance v0.3.1
 	github.com/redhat-developer/app-services-sdk-go/registrymgmt v0.2.0
-	github.com/redhat-developer/service-binding-operator v0.8.0
+	github.com/redhat-developer/service-binding-operator v0.9.0
```
craicoverflow, line 29: Wasn't v0.10.0 announced? Maybe that could resolve it.

pkg/cluster/serviceBinding.go
```diff
@@ -214,7 +193,8 @@ func fetchAppNameFromCluster(ctx context.Context, clients *KubernetesClients, lo
 	}
 
 	if len(appNames) == 0 {
-		return "", errors.New("selected namespace has no deployments")
+		// TODO add key
+		return "", errors.New(localizer.MustLocalize("cluster.serviceBinding.error.noapps"))
```
craicoverflow, line 197: localizer.MustLocalizeError :)

pkg/localize/locales/en/cmd/cluster_bind.en.toml
```diff
@@ -25,19 +28,18 @@ $ rhoas cluster bind
...
 [cluster.bind.flag.bindName]
 one = 'Name of the Service binding object to create when using operator'
 
-[cluster.bind.flag.forceOperator.description]
-one = '''Use ServiceBindingOperator only and fail if Operator is not installed'''
-
-[cluster.bind.flag.forceSDK.description]
-one = '''Use Service Binding SDK and skip ServiceBindingOperator even if installed on the cluster'''
+[cluster.bind.flag.useDeploymentConfig.description]
+one = '''Use DeploymentConfig to detect your application'''
```
craicoverflow, line 38: Just so you know, the ''' ticks is only necessary for multiline texts.

pkg/localize/locales/en/cmd/cluster_bind.en.toml
```diff
@@ -3,16 +3,19 @@ one = 'Connect your RHOAS services to Kubernetes or OpenShift applications'
...
-services with.
+services with. 
+
+If your application uses DeploymentConfig --deployment-config flag needs to be supplied in order to detect it.
+When deployment-config flag is used binding will not take effect until application is manually redeployed by user.
 
 Bind command detects Kubernetes deployments and inject managed services credentials to them.
 For information about what credentials are injected please refer to individual services
-Command will inject credentials as files into `/bindings` folder inside your application.
+By default command will inject credentials as files into `/bindings` folder inside your application.
```
craicoverflow, line 16:
```suggestion
By default the command will inject credentials as files into `/bindings` folder inside your application.
```

pkg/localize/locales/en/cmd/cluster_bind.en.toml
```diff
@@ -3,16 +3,19 @@ one = 'Connect your RHOAS services to Kubernetes or OpenShift applications'
 
 [cluster.bind.cmd.longDescription]
 one = '''
-Command allows you to connect services created by connect command to your application
+Command allows you to connect services created by rhoas cluster connect command to your application
```
craicoverflow, line 6:
```suggestion
Command allows you to connect services created by "rhoas cluster connect" command to your application
```

pkg/cmd/cluster/bind/bind.go
```diff
@@ -66,8 +67,8 @@ func NewBindCommand(f *factory.Factory) *cobra.Command {
 	cmd.Flags().BoolVarP(&opts.forceCreationWithoutAsk, "yes", "y", false, opts.localizer.MustLocalize("cluster.common.flag.yes.description"))
 	cmd.Flags().StringVarP(&opts.namespace, "namespace", "n", "", opts.localizer.MustLocalize("cluster.common.flag.namespace.description"))
 	cmd.Flags().BoolVar(&opts.ignoreContext, "ignore-context", false, opts.localizer.MustLocalize("cluster.common.flag.ignoreContext.description"))
-	cmd.Flags().BoolVar(&opts.forceOperator, "force-operator", false, opts.localizer.MustLocalize("cluster.bind.flag.forceOperator.description"))
-	cmd.Flags().BoolVar(&opts.forceSDK, "force-sdk", false, opts.localizer.MustLocalize("cluster.bind.flag.forceSDK.description"))
+	cmd.Flags().BoolVar(&opts.deploymentConfig, "deployment-config", false, opts.localizer.MustLocalize("cluster.bind.flag.useDeploymentConfig.description"))
```
craicoverflow, line 70: What about a flag which is generic and accepts multiple different types:

```shell
rhoas cluster connect --type=deployment|deploymentconfig
```

craicoverflow (commented):
LGTM

go.mod
wtrocki: 0.10.0 is actually the version of the cluster that breaks things. 
0.10.0 also doesn't seem to compile https://github.com/redhat-developer/service-binding-operator/issues/1022

I have verified this for deployment config. So service binding is created and command works. 
but we cannot verify this end to end. I think we should merge this knowing that command works as it should

pkg/cluster/serviceBinding.go
```diff
@@ -145,24 +149,16 @@ func performBinding(ctx context.Context, options *ServiceBindingOptions, ns stri
...
 	// Check of operator is installed
 	_, err := clients.dynamicClient.Resource(v1alpha1.GroupVersionResource).Namespace(ns).
 		List(ctx, metav1.ListOptions{Limit: 1})
 	if err != nil {
-		if options.ForceUseOperator {
-			return localizer.MustLocalizeError("cluster.serviceBinding.operatorMissing", localize.NewEntry("Error", err))
-		}
-		logger.Debug("Service binding Operator not available. Will use SDK option for binding")
-		return useSDKForBinding(clients, sb)
+		return localizer.MustLocalizeError("cluster.serviceBinding.operatorMissing", localize.NewEntry("Error", err))
```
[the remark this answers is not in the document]
wtrocki: we discussed that yesteday that we going to map errors specifically for providing more reliable info. 
So we should probably:

- Log error that we recognize 
- Return error to be printed again with more generic/non i18n content

pkg/localize/locales/en/cmd/cluster_bind.en.toml
wtrocki: I hate to have mental blockage so keep adding them for single strings just in case we will need multiline :D

pkg/cmd/cluster/bind/bind.go
wtrocki: Love it! Thank you for amazing suggestion. This is more future proof. 
deployment is default, DeploymentConfig is extra case that we were asked to support. 
We will not support both so app-type sounds like better flag

wtrocki: Actually when implementing it I see that needs to be typed which requires much more than boolean enablement. 
Ideally we should be enabling both later but we cannot now due to bug in SBO. Leaving it as it is. Hope you do not mind.

pkg/cluster/serviceBinding.go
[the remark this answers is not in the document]
craicoverflow: Yes - in the commands it makes sense to map errors and preserving the error object does not matter too much. However in an SDK scenario (like this one) I think it is important to preserve it because the place it is being used is able act different based on the error type.

rkpattnaik780:
The error messages need formatting at the last line.

Screenshot from 2021-09-17 18-15-52

Screenshot from 2021-09-17 18-28-31

Also for the latter error message, not sure if it was expected to fail, I could do a bind after it.

pkg/cluster/serviceBinding.go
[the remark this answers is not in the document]
wtrocki: In this case error will be cryptic - 404 kafka API doesn't exist or missing auth token. Do you suggest to return original error only?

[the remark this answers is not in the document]
craicoverflow: No it can be wrapped like this:

```go
return fmt.Errorf("your reason: %w", err)
```

`%w` means wrap the error.

pkg/cmd/cluster/bind/bind.go
craicoverflow: Don't mind no :) Merely a suggestion, but I don't have the domain knowledge to justify if it makes sense or not.

wtrocki:
@rkpattnaik780 That is mostly because of logic - it lets you to select kafka that you do not have on the cluster. 
I have left comment to solve that problem - once we move to binding we should read Kafka and Registry connections from cluster and specify their names instead of reading them from backend

wtrocki:
Blocked by: 
https://github.com/redhat-developer/service-binding-operator/issues/1025

My quess is that we going to merge and release it so we can rebase and continue refactoring it for multiservice architecture

PR #1078, opened 2021-09-16
refactor: made improvements in the dump package

### LINKED ISSUE #1067: Improvements in the `dump` package
The `PrintDataInFormat` method could do with some minor changes to improve it.

- [ ] Change the name to something cleaner, and shorter. `DumpFormatted` is my preferred (fewer words used), I am open to other suggestions.
- [ ] The `io.Writer` parameter always comes first in a Go method which uses that writer. Let's put it first here too.
- [ ] Return an error if any from this method. We should not be ignoring errors here. It is up to the method usage to decide whether to use the error or not, it should be returned.
- [ ] Create a special type `DumpFormat` and set the constants `XFormat` constants to use this format (nice to have, not needed though). This means we can limit the valid values for `format` used in `PrintDataInFormat`.

alexal:
Closes #1067

### DIFF
--- a/pkg/cmd/kafka/consumergroup/describe/describe.go
+++ b/pkg/cmd/kafka/consumergroup/describe/describe.go
@@ -145,7 +145,7 @@ func runCmd(opts *options) error {
 	case dump.EmptyFormat:
 		printConsumerGroupDetails(stdout, consumerGroupData, opts.localizer)
 	default:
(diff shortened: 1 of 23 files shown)


wtrocki (approved):
Leaving @craicoverflow to confirm method name

craicoverflow (approved):

PR #1081, opened 2021-09-17
refactor: made improvements in kafka.InteractiveSelect() function

### LINKED ISSUE #1074: Incorrect prompt text for `rhoas kafka use`
```shell
$ rhoas kafka use
? Select Kafka instance to connect: my-kafka
```

Select Kafka instance to "connect" is inacxcurate. I believe this is being used also in the `cluster connect` command. We need a distinct prompt title which relates to "using" an instance.

alexal:
Closes #1074

### DIFF
--- a/pkg/kafka/interactive.go
+++ b/pkg/kafka/interactive.go
@@ -2,10 +2,9 @@ package kafka
 
 import (
 	"context"
-	"fmt"
-
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/redhat-developer/app-services-cli/pkg/connection"
+	"github.com/redhat-developer/app-services-cli/pkg/localize"
 	"github.com/redhat-developer/app-services-cli/pkg/logging"
 	kafkamgmtclient "github.com/redhat-developer/app-services-sdk-go/kafkamgmt/apiv1/client"
 )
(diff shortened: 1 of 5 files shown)


pkg/kafka/interactive.go
```diff
@@ -14,16 +13,16 @@ const (
...
 )
 
-func InteractiveSelect(ctx context.Context, connection connection.Connection, logger logging.Logger) (*kafkamgmtclient.KafkaRequest, error) {
+func InteractiveSelect(ctx context.Context, connection connection.Connection, logger logging.Logger, localizer localize.Localizer) (*kafkamgmtclient.KafkaRequest, error) {
 	api := connection.API()
 
 	response, _, err := api.Kafka().GetKafkas(ctx).Size(queryLimit).Execute()
 	if err != nil {
-		return nil, fmt.Errorf("unable to list Kafka instances: %w", err)
+		return nil, localizer.MustLocalizeError("kafka.common.error.couldNotFetchKafkas", localize.NewEntry("Error", err))
```
craicoverflow, line 21: We should try to avoid internationalizing the error like this. Returning the `error` object means it can be type checked and cast to a type. What's happening here is it is being converted to a string, meaning it can no longer be cast.

craicoverflow (changes requested):

pkg/kafka/interactive.go
alexal: Sure, done.

pkg/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -46,7 +46,7 @@ one = 'Format in which to display the Kafka instances (choose from: "json", "yml
 
 [kafka.common.error.couldNotFetchKafkas]
 description = 'Error message when list of Kafka instances could not be fetched'
-one = 'unable to list Kafka instances: {{.Error}}'
+one = 'unable to list Kafka instances'
```
craicoverflow: Is this used anywhere anymore?

craicoverflow (approved):

pkg/localize/locales/en/cmd/kafka.en.toml
alexal: It is not used, but it was there, so not sure whether I should delete it or not.

wtrocki: Kill it with fire. Then we merge PR.

alexal: Done.

PR #1082, opened 2021-09-17
refactor: use make to create fixed size list

### LINKED ISSUE #911: Change artifactrows in list commands
```suggestion
	var rows []artifactRow
```

This will create the same value - in Go it is generally preferred to create variables which have not been given a value using `var`.

As an aside, this slice does not have a fixed value - but since we know what the size will be we could set a fixed size.

See https://github.com/redhat-developer/app-services-cli/blob/reset_offset/pkg/cmd/kafka/consumergroup/resetoffset/reset_offset.go#L241

This will have no noticeable impact on performance for the CLI, so i's totally optional, just letting you know that it is possible.

_Originally posted by @craicoverflow in https://github.com/redhat-developer/app-services-cli/pull/859#discussion_r687897771_

alexal:
Closes #911 

**Verification Steps**

- [ ] Run export RHOAS_DEV=true to enable this hidden feature.
- [ ] Create a Service Registry instance, or skip this if you already have one.
- [ ] Run rhoas `service-registry artifact list` it should provide list of Service Registry artifacts

Type of change

- [ ] Other: code refactoring

### DIFF
--- a/pkg/cmd/registry/artifact/crud/list/list.go
+++ b/pkg/cmd/registry/artifact/crud/list/list.go
@@ -157,7 +157,7 @@ func runList(opts *options) error {
 }
 
 func mapResponseItemsToRows(artifacts []registryinstanceclient.SearchedArtifact) []artifactRow {
-	rows := []artifactRow{}
+	rows := make([]artifactRow, len(artifacts))
 
 	for i := range artifacts {
 		k := (artifacts)[i]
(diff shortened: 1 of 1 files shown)


craicoverflow:
This is a `refactor`.

craicoverflow:
@alexal can you please provide verification steps with every pull request?

alexal:
> @alexal can you please provide verification steps with every pull request?

Like what command to run to make sure it works?

craicoverflow:
Yes, most pull requests should have one. Here's an example: https://github.com/redhat-developer/app-services-cli/pull/1050

alexal:
> Yes, most pull requests should have one. Here's an example: #1050

Sure, not a problem.

craicoverflow (approved):

PR #1096, opened 2021-09-20
Make rhoas kafka topic update output less verbose

### LINKED ISSUE #1092: For the `update` commands, do not display the updated object as JSON by default.
Running `rhoas kafka topic update` will by default print the topic configuration to stdout. This is a massive object and it is unlikely that the user needs to see it. We should default to printing a simple success message by default, and allow users to use the output flag to output the object.

SumLare:
Closes #1092

### Type of change

- [ ] Documentation change
- [x] Other (Chore, command output change)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/kafka/topic/update/update.go
+++ b/pkg/cmd/kafka/topic/update/update.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/redhat-developer/app-services-cli/internal/config"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/factory"
+	"github.com/redhat-developer/app-services-cli/pkg/dump"
 	"github.com/redhat-developer/app-services-cli/pkg/iostreams"
 	"github.com/redhat-developer/app-services-cli/pkg/logging"
 	kafkainstanceclient "github.com/redhat-developer/app-services-sdk-go/kafkainstance/apiv1internal/client"
@@ -98,8 +99,10 @@ func NewUpdateTopicCommand(f *factory.Factory) *cobra.Command {
 
 			}
 
-			if err = flag.ValidateOutput(opts.outputFormat); err != nil {
-				return err
+			if opts.outputFormat != "" {
+				if err = flag.ValidateOutput(opts.outputFormat); err != nil {
+					return err
+				}
 			}
 
 			// check if the partition flag is set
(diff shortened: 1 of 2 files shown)


pkg/cmd/kafka/topic/update/update.go
```diff
@@ -302,7 +301,8 @@ func runCmd(opts *options) error {
 	}
 
 	opts.Logger.Info(opts.localizer.MustLocalize("kafka.topic.update.log.info.topicUpdated", topicNameTmplPair, kafkaNameTmplPair))
-	return dump.Formatted(opts.IO.Out, opts.outputFormat, response)
+
+	return nil
```
craicoverflow, line 305: The value of `--output` should be respected. The current default is JSON and we should still print to JSON when it is specified. We need to change the default to nothing, and only then do you not print the object.

wtrocki: I have removed output from update commands in service registry.
Return status should be enough for scripts

craicoverflow: > I have removed output from update commands in service registry.
> Return status should be enough for scripts

So you are saying maybe we should not return any data, ever?

wtrocki: I'm saying what I have done for registry is:
- Removed output flag for update (and other commands that do not expect to return anything)
- Always print some message out to user and scripts use return code

To formalize this as guideline:
```
Commands that are not expected to print out anything else than status messages do not need an output flag.
```

If guideline is to always have output flag then we would print some output as 
```json
{
"message":"successfully done something"
}
```

wtrocki:
@SumLare Rebase needed. Changes look good :) I can help with the rest of the issues later.

SumLare:
> @SumLare Rebase needed. Changes look good :) I can help with the rest of the issues later.

Will do!

pkg/cmd/kafka/topic/update/update.go
craicoverflow: registry does have output flag for update in places: https://github.com/redhat-developer/app-services-cli/blob/d7268802175750df88e3942889b5cd376ec91c21/pkg/cmd/registry/artifact/crud/update/update.go#L95

I think we may not need output at all. I will create a PR on top of this one to apply it globally.

craicoverflow:
Closing. Your commit will be included in #1138 @SumLare :)

PR #1097, opened 2021-09-20
refactor: use make to create fixed size list

alexal:
I've found a few more cases similar to what was described in #911, so I've fixed them.

### Verification Steps
Run commands affected by this change, such as
1. rhoas kafka use
2. rhoas kafka consumer-group describe
3. rhoas kafka list
4. rhoas kafka topic list
5. rhoas kafka consumer-group list

### Type of change
- [x] Code refactoring


rkpattnaik780 (approved):
Great work @alexal 
Thanks or this PR!

PR #1099, opened 2021-09-21
feat: enable kafka update command

craicoverflow:
The `rhoas kafka update` command was hidden as there needed to be some changes in the backend. This has been enabled now, so we can make this command available to users.

### Verification Steps

1. Log in as an org admin.
1. Ensure `RHOAS_DEV=false` or unset.
1. Run `rhoas kafka update`
2. You will be shown a list of users to choose from.
3. Log in as a non-admin.
4. Run `rhoas kafka update`, you should see:

```shell
$ rhoas kafka update
Only organization administrators have the ability to update a Kafka instance.
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/docs/commands/rhoas_kafka_update.adoc
+++ b/docs/commands/rhoas_kafka_update.adoc
@@ -1,0 +1,61 @@
+ifdef::env-github,env-browser[:context: cmd]
+[id='ref-rhoas-kafka-update_{context}']
+= rhoas kafka update
+
+[role="_abstract"]
+Update configuration details of a Kafka instance.
+
+[discrete]
+== Synopsis
+
+Update certain configuration details of a Kafka instance.
+
+Currently it is possible to update the "owner" field. The new owner 
+will have super-user privileges and will be authorized to manage this instance.
+
+
+....
+rhoas kafka update [flags]
+....
+
+[discrete]
+== Examples
+
+....
+# update the Kafka instance owner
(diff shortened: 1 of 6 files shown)


docs/commands/rhoas_kafka_update.adoc
```diff
@@ -0,0 +1,61 @@
...
+[role="_abstract"]
+Update configuration details of a Kafka instance.
+
+[discrete]
+== Synopsis
+
+Update certain configuration details of a Kafka instance.
+
+Currently it is possible to update the "owner" field. The new owner 
+will have super-user privileges and will be authorized to manage this instance.
```
wtrocki, line 14: `super-user` is not clear without explanation. I will remove that from original statement.

wtrocki:
@craicoverflow 

We should be able to know if user is admin from JWT token.
Maybe as form of workaround we can check user claim before performing update - handling error:

"Only organization admin can use interactive mode" or something like that.

wtrocki (approved):

wtrocki:
Alternatively we can also handle error from backend without even checking permissions (Even better as it would work after we fix that on backend)

craicoverflow:
Updated verification steps to include the flow for a non-org admin. Basically, since there is nothing a non-admin can update we can just print a message saying this.

rkpattnaik780 (commented):
LGTM!

PR #1128, opened 2021-09-23
Kafka limits for eval instances are not properly handled in the CLI 4a1bcff

wtrocki:
# Verification 
`rhoas kafka create --name` on account that has already one eval instance.

Fix for bug logged by @DuncanDoyle

### DIFF
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -3,11 +3,11 @@ package create
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"os"
 	"os/signal"
 	"time"
 
+	"github.com/redhat-developer/app-services-cli/pkg/api/kas"
 	"github.com/redhat-developer/app-services-cli/pkg/icon"
 	"github.com/redhat-developer/app-services-cli/pkg/ioutil/spinner"
 
@@ -185,12 +185,13 @@ func runCreate(opts *options) error {
 		defer httpRes.Body.Close()
 	}
 
-	if httpRes.StatusCode == http.StatusBadRequest {
-		return opts.localizer.MustLocalizeError("kafka.create.error.conflictError", localize.NewEntry("Name", payload.Name))
-	}
-
-	if httpRes.StatusCode == http.StatusTooManyRequests {
-		return opts.localizer.MustLocalizeError("kafka.create.error.oneinstance")
+	if apiErr := kas.GetAPIError(err); apiErr != nil {
+		switch apiErr.GetCode() {
+		case kas.ErrorCode24:
+			return opts.localizer.MustLocalizeError("kafka.create.error.oneinstance")
+		case kas.ErrorCode36:
+			return opts.localizer.MustLocalizeError("kafka.create.error.conflictError", localize.NewEntry("Name", payload.Name))
+		}
 	}
 
 	if err != nil {
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -8,6 +8,7 @@ import (
 	"os/signal"
 	"time"
 
+	"github.com/redhat-developer/app-services-cli/pkg/api/kas"
 	"github.com/redhat-developer/app-services-cli/pkg/icon"
 	"github.com/redhat-developer/app-services-cli/pkg/ioutil/spinner"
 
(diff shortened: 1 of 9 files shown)


pkg/cmd/kafka/create/create.go
```diff
@@ -189,6 +190,10 @@ func runCreate(opts *options) error {
 		return opts.localizer.MustLocalizeError("kafka.create.error.conflictError", localize.NewEntry("Name", payload.Name))
 	}
 
+	if kas.IsErr(err, kas.OneEvalInstanceAllowed) {
```
rkpattnaik780, line 193: Getting erroneously evaluated as false.
Screenshot from 2021-09-24 12-19-56

craicoverflow: I looked into this and it appears that the error dooes not properly encode the metadata within the error, it is stringified:

image

I checked another error on the same endpoint but it acts as expected:

image

wtrocki: Sorry for merge without review.
Feel free to reopen/revert this change

craicoverflow: Hmm but this is not merged?

wtrocki: Got notification from other PR and got confused. @craicoverflow This looks like sdk bug related to using very unusual http status code on backend. I would recomend us to change status code in sdk to confirm that this is the case then ping backend to correct status code to 400 or something

wtrocki: 429 is ratelimiting status code and it never carries payload etc. @machi1990

machi1990: > 429 is ratelimiting status code and it never carries payload etc. @machi1990

Are you not mixing this status code with `204`? 

> Seems like not valid use for this error

What error would you expect for this case? Does the ui face the same problem? I think they integrated it already. 

pinging @ziccardi for additional input

craicoverflow: 429 is specifically for rate-limiting, which is not what is happening in this scenario.

wtrocki: Same problem happens for UI and this code is not handled  in UI right now:
https://github.com/bf2fc6cc711aee1a0c2a/kas-ui/blob/1c98ca3820e7574529e8251a6c146a49bf6484a8/src/app/modules/OpenshiftStreams/dialogs/CreateInstance/CreateInstance.tsx#L259
ErrorCodes.INSUFFICIENT_QUOTA is still used. Not sure if that code is returned from backend/not dead?

To avoid team introducing new codes and then passing them to indivudual cli,ui devs we need to come up with strategy for this.

What we should do now
- Correct error code on backend to 400 (control plane team) @machi1990 to confirm
- Adjust UI and CLI to new code once it is there (devexp)
- Raise topic on architecture call to figure out how to handle this problem in future. (devexp done some spikes for this to avoid issues like this one) (Devexp)
-

wtrocki: To confirm:

- No one uses this error right now
- we cannot make clients to use this type of error as we will be introducing security issue

ziccardi: When we implemented the feature in the control plane, we had a discussion about the return code to be used and in the end we opted for 429. In our case, it can be seen as 'exceeding the rate of 1 instance every 48 hours' (after 48 hours the eval instance is deleted). 

According to [RFC-6585](https://datatracker.ietf.org/doc/html/rfc6585#section-4) 429 errors should carry the details of what has happened.

wtrocki: Lets hear our arguments on this. Can we can have short call next week to discuss this?

craicoverflow: As a workaround until then, I have added a check based on the status code: 0ff83ee76c3b90381b551aedcaf28d8365ca73e1

craicoverflow:
@wtrocki I've added 0ff83ee76c3b90381b551aedcaf28d8365ca73e1 to workaround this issue for the time being. Please check.

wtrocki (commented):
Lets hold with merge for 2 days until we get sense on how to handle api errors

wtrocki:
@craicoverflow This is unblocked. Api returns 403 now.
Can you adjust changes for us. Thank you

Issue #1114, opened 2021-09-23
rhoas service-registry artifact update - Add a note that this should conform to validit and compatability rules


wtrocki:
## Description
 
rhoas service-registry artifact update - Add a note that this should conform to validity and compatibility rules

PR #1133, opened 2021-09-23
docs: add note for validity rules for artifact command

wtrocki:


wtrocki:
https://github.com/redhat-developer/app-services-cli/issues/1114

PR #1141, opened 2021-09-27
fix: update devpreview to beta

wtrocki:
Change alpha to beta that will happen in the next release (with announcement)


wtrocki:
FYI - documentation will change - we move to beta as the other services/api states

rkpattnaik780 (approved):

PR #1146, opened 2021-09-27
chore: add example avro schema

wtrocki:
Add example avro schema that we can reference in the documentation of the registry commands


wtrocki:
we need this on master to reference in CLI

PR #1187, opened 2021-10-06
feat: service registry integration with cluster commands

wtrocki:
Posting for early review/feedback. Still working on extra features like disconnect. 
I haven't verified changes yet and still working on resolving other issues like service account creation

### DIFF
--- a/docs/commands/rhoas_cluster.adoc
+++ b/docs/commands/rhoas_cluster.adoc
@@ -56,6 +56,13 @@ ifdef::pantheonenv[]
 endif::[]
 
 ifdef::env-github,env-browser[]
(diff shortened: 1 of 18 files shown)


wtrocki:
That is chicken or egg problem.

All we get from api is 403.
And this can happen for
- If operator is not there
- if user is not logged in
- if namespace is wrong
- if resource name is wrong

So in fact this message is already too specific in my taste. I was trying to do extra checks for status and maybe do some check for namespace before making other calls. That will work but be little bit extensive IMHO

wtrocki:
I have released operator with service registry support internally first. We want to provide ability to collect feedback internally before we release it to the end users. Merging

PR #1202, opened 2021-10-11
fix: fix list order

wtrocki:
Small issue that semi fixes ordering issue for acl list. (still random but at least it is ordered by principals


wtrocki:
Actually will merge into convinience commands

PR #1213, opened 2021-10-12
chore: tests cases for acl

wtrocki:
Adding test cases for acl commands as separate folder

### DIFF
--- a/.cases/kafka-acl-grant-permissions.sh
+++ b/.cases/kafka-acl-grant-permissions.sh
@@ -11,6 +11,7 @@
 ## Framework
 
 source ./asserts.sh
+alias rhoas=$(go env GOPATH)/bin/rhoas
 
 ## Cases
 rhoas kafka acl grant-permissions 


craicoverflow:
> Great point. Couple options I have investigated before I went with this
> 
> * Use readme.md files instead of bash
> * use bash but no simple assertions
> 
> I stick with bash with simple assertions assuming that actual verification is still manual. I like how commands are organized so I can execute all validation commands and quicly inspect output messages - something that were taking me 10 min before.
> 
> If you feel that aserts are problem I'm happy to remove them - added them for cool factor only as they arent reliable or even checking anything that status code

I think the are fine to leave in, but would you consider rewriting in Go using the Go testing framework?

These can be isolated from unit tests and only executable via `make verify-cmd` or something. Go tests can actually be debugged easily which will help with problem diagnosis.

wtrocki:
@craicoverflow done. We need more cases for each Pr so merging and will refine this as we go

PR #1235, opened 2021-10-19
refactor:change (*Config).HasServiceRegistry() return value and boolean if value exists

### LINKED ISSUE #1233: Change `(*Config).HasServiceRegistry()` method to return two values: boolean if exists and the ID value.

We use the `HasServiceRegistry` method to check if a registry instance is currently set, and then must directly access the nested Kafka instance ID if it does exist.

Example:

```go
if !cfg.HasServiceRegistry() {
  return err
}

opts.instanceID = cfg.Services.ServiceRegistry.ClusterID
```

### Proposed solution

We can return if the value exists, as well as the value itself in a single method:

```go
instanceID, ok := cfg.GetServiceRegistryIdOk()
if !ok {
  return err
}
opts.instanceID = instanceID
```

This is both safer (less chance of a nil-pointer exception) and cleaner.

KarolosLykos:
Change (*Config).HasServiceRegistry() method to return two values: boolean if exists and the ID value.

Closes #1233

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/internal/config/type.go
+++ b/internal/config/type.go
@@ -46,7 +46,10 @@ func (c *Config) HasKafka() bool {
 		c.Services.Kafka.ClusterID != ""
 }
 
-func (c *Config) HasServiceRegistry() bool {
-	return c.Services.ServiceRegistry != nil &&
-		c.Services.ServiceRegistry.InstanceID != ""
+func (c *Config) HasServiceRegistry() (string, bool) {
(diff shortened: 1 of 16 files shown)


craicoverflow:
Changes included in #1244 along with some other changes

PR #1240, opened 2021-10-19
fix: remove indentation in long descriptions

### LINKED ISSUE #1238: Incorrect formatting for rhoas kafka acl * command long descriptions
## Description

**Version**: main

The long description of some of the rhoas kafka acl * commands is formatted incorrectly.

### Steps to reproduce

Check the "Synopsis" section for the following commands:

* https://github.com/redhat-developer/app-services-cli/blob/main/docs/commands/rhoas_kafka_acl.adoc#synopsis
* https://github.com/redhat-developer/app-services-cli/blob/main/docs/commands/rhoas_kafka_acl_delete.adoc#synopsis
* https://github.com/redhat-developer/app-services-cli/blob/main/docs/commands/rhoas_kafka_acl_list.adoc#synopsis

The long description should be formatted as regular text (for example, https://github.com/redhat-developer/app-services-cli/blob/main/docs/commands/rhoas_kafka_acl_grant-access.adoc#synopsis).

I thought this might be an issue with how the long descriptions are written in https://github.com/redhat-developer/app-services-cli/blob/main/pkg/localize/locales/en/cmd/acl.en.toml, but I don't see any issues. The problem only seems to be occurring with the acl commands that use the multi-line `one = ''' ... '''` long description format (as opposed to the `one = ' ... '` single-line format), but other commands that use the multi-line long description format don't have the same formatting issue.

craicoverflow:
fixes #1238

### DIFF
--- a/docs/commands/rhoas_kafka_acl.adoc
+++ b/docs/commands/rhoas_kafka_acl.adoc
@@ -8,23 +8,23 @@ Kafka ACL management for users and service accounts
 [discrete]
 == Synopsis
 
-	Set of commands that will let you manage Kafka ACLs.
-	By default, every users and service account have limited access to their Kafka instance (Only DESCRIBE permission is enabled for TOPIC, ACL, and GROUP).
-	
+Set of commands that will let you manage Kafka ACLs.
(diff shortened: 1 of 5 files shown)


craicoverflow:
Needs merging to fix docs.

PR #1273, opened 2021-11-02
fix: add arm64 support

wtrocki:


craicoverflow:
Is there a supporting issue/discussion for this? The reason I ask is I want to know if we need to also support this downstream.

wtrocki:
Yes. We should
https://9to5mac.com/2021/01/06/m1-macs-market-share/

Issue #1275, opened 2021-11-03
Service registry creation is broken due to old terms and conditions.


wtrocki:
## Description

We need to check old terms and conditions for service registry.

craicoverflow:
Should it be that the service registry should also support the new T&Cs? Why are they diverging?

wtrocki:
Service registry will support old TC

PR #1276, opened 2021-11-03
Use remote spec for terms and conditions

wtrocki:
Requires extra work and untested but code should be ok to review.

Fixes https://github.com/redhat-developer/app-services-cli/issues/1275

### DIFF
--- a/.cases/termsandconditions.sh
+++ b/.cases/termsandconditions.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+## Terms and Conditions acceptance
+
+## Requirements
+## Create new account without any terms and conditions accepted.
+
+## Framework
+
+alias rhoas=$(go env GOPATH)/bin/rhoas
+
+## Cases
+
+rhoas service-registry create --name=test -v
+## In order to be able to create a new instance, you must first review and accept the terms and conditions:
+## https://www.redhat.com/wapps/tnc/ackrequired?site=ocm&event=register
+
+rhoas kafka create --name=test --provider=aws --region=eu-east1 -v
+## In order to be able to create a new instance, you must first review and accept the terms and conditions:
+## https://www.redhat.com/wapps/tnc/ackrequired?site=ocm&event=onlineService
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -25,11 +25,8 @@ var (
 	// RepositoryName is the remote GitHub repository for the releases
 	RepositoryName = "app-services-cli"
 
-	// TermsReviewEventCode is the event code used when checking the terms review
-	TermsReviewEventCode = "register"
-
-	// TermsReviewSiteCode is the site code used when checking the terms review
-	TermsReviewSiteCode = "ocm"
+	// TermsReviewSpecURL Url used to download terms and conditions specification
+	TermsReviewSpecURL = "https://raw.githubusercontent.com/redhat-developer/app-services-ui/6699c93c9d5f123a2c898d1a1d72a9c290a20820/static/configs/terms-conditions-spec.json"
 
 	// DefaultPageSize is the default number of items per page when using list commands
 	DefaultPageSize = "10"
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -26,7 +26,7 @@ var (
 	RepositoryName = "app-services-cli"
 
 	// TermsReviewSpecURL Url used to download terms and conditions specification
-	TermsReviewSpecURL = "https://raw.githubusercontent.com/redhat-developer/app-services-ui/6699c93c9d5f123a2c898d1a1d72a9c290a20820/static/configs/terms-conditions-spec.json"
+	TermsReviewSpecURL = "https://raw.githubusercontent.com/redhat-developer/app-services-ui/main/static/configs/terms-conditions-spec.json"
 
 	// DefaultPageSize is the default number of items per page when using list commands
 	DefaultPageSize = "10"
--- a/pkg/ams/terms.go
+++ b/pkg/ams/terms.go
@@ -47,6 +47,8 @@ func GetRemoteTermsSpec(context *context.Context, logger logging.Logger) TermsAn
 		return fallbackTocSpec
 	}
 
+	logger.Debug("Terms spec: ", specJson)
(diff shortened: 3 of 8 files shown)


wtrocki:
Once this PR is deployed to production: https://github.com/redhat-developer/app-services-ui/pull/153

we should change URL from github to the console.redhat.com

internal/build/build.go
```diff
@@ -25,11 +25,8 @@ var (
 	// RepositoryName is the remote GitHub repository for the releases
 	RepositoryName = "app-services-cli"
 
-	// TermsReviewEventCode is the event code used when checking the terms review
-	TermsReviewEventCode = "register"
-
-	// TermsReviewSiteCode is the site code used when checking the terms review
-	TermsReviewSiteCode = "ocm"
+	// TermsReviewSpecURL Url used to download terms and conditions specification
+	TermsReviewSpecURL = "https://raw.githubusercontent.com/redhat-developer/app-services-ui/main/static/configs/terms-conditions-spec.json"
```
wtrocki, line 29: Considering this is hotfix I think we could merge it with that location and switch over to cdn version after?

pkg/ams/terms.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/redhat-developer/app-services-cli/internal/build"
+)
+
+var fallbackTocSpec = []byte(`{
```
wtrocki, line 11: TODO. Need to document where is the source of this file

pkg/ams/terms.go
```diff
@@ -0,0 +1,43 @@
...
+       "StopOnTermsChange": true
+    }
+  }
+`)
+
+func GetRemoteTermsSpec() TermsAndConditionsSpec {
+	response, err := http.Get(build.TermsReviewSpecURL)
+
+	var specJson []byte
+	if err != nil {
```
wtrocki, line 29: I was considering 3 use cases when we get a network error:

1) Fail entire command with a message: Error: downloading terms specification failed
2) Print error but let users continue with creating service - accepting that they would 
3) Provide failover for download by having spec in the CLI. 

I have done some pros and cons and went with option nr3.

Nr1 will be too internal for end-users to understand and result in a very bad experience. There is no way we can troubleshoot this etc.

Nr2 will be the same and also will require handling of the error from backend but without terms URL.

internal/build/build.go
craicoverflow: I think we could have three levels of fallback:

1. Try to load from CDN, if it fails go to number 2:
2. Try to load from GitHub, if it fails to go number 3:
3. Load the local spec file.

If it is cached, doing 3 checks should not be a problem.

craicoverflow: I think we could have three levels of fallback:

1. Try to load from CDN, if it fails go to number 2:
2. Try to load from GitHub, if it fails to go number 3:
3. Load the local spec file.

If it is cached, doing 3 checks should not be a problem.

wtrocki: Having two sources (one of which we do not control - github) is going to be tricky from compliance/firewal rules point of view. We use Akamai as CDN for console so this should be very reliable way. Having second URL would be problematic from debugging point of view (in case issues happens). We have fallback in code right now

pkg/ams/terms.go
```diff
@@ -0,0 +1,43 @@
...
+    "service-registry":{
+       "EventCode":"onlineService",
+       "SiteCode":"ocm",
+       "StopOnTermsChange": true
+    }
+  }
+`)
+
+func GetRemoteTermsSpec() TermsAndConditionsSpec {
+	response, err := http.Get(build.TermsReviewSpecURL)
```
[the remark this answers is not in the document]
wtrocki: My understanding is that considering how frequent you would create instances using cache might result in potential. 

We do not cache AMS terms check so caching this would be overkill. 
I did added cache initially in login but it would be too complex and hard to verify/handle all the cases.

> Or perhaps have this as a JSON file which is embedded into the binary

I had that json as fallback but moved to structure and provided documentation on where is the source for the cdn file.

wtrocki:
@craicoverflow added verification steps in .cases/termsandconditions.sh

wtrocki:
Investigating unrelated issues with missing debug level logging

.cases/termsandconditions.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
craicoverflow, line 1: Getting permission denied on this file. Should it be automatically executable? On my Linux machine it was, could be different on MacOS though..

pkg/ams/terms.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// GetRemoteTermsSpec fetch event and site code information associated with the services
+// Function is used to dynamically download new terms and conditions specifications
+// without forcing end users to update their CLI.
+func GetRemoteTermsSpec(context *context.Context, logger logging.Logger) TermsAndConditionsSpec {
+	client := &http.Client{}
+	req, err := http.NewRequestWithContext(*context, http.MethodGet, build.TermsReviewSpecURL, nil)
+	if err != nil {
+		logger.Debug("Fetching remote terms failed with error ", err)
```
craicoverflow, line 34:
```suggestion
		logger.Debug("Fetching remote terms failed with error", err)
```

Spaces are added automatically

pkg/ams/terms.go
```diff
@@ -0,0 +1,59 @@
...
+		return fallbackTocSpec
+	}
+	response, err := client.Do(req)
+	if err != nil || response.Body == nil {
```
craicoverflow, line 38: `response` could be nil, but if response is not nil, response.Body will never be nil
```suggestion
	if err != nil || response == nil {
```

craicoverflow (approved):
One comment about checking for nil response, other than that all good.

.cases/termsandconditions.sh
wtrocki: Amazing feedback. I will move those to md files as sh doesnt make sense when we copy commands to execute
We can have template for new verification

PR #1280, opened 2021-11-04
fix: enable debug logging for cli

wtrocki:
Analytics implementation broke debug logging. 
This PR fixes it


wtrocki:
@craicoverflow Merging this as I want it to be enabled in other PR to help you with verification

PR #1291, opened 2021-11-05
fix: update terms spec url

wtrocki:
Use production based json file.
use termsandconditions file for verification

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ SHELL = bash
 RHOAS_VERSION ?= "dev"
 REPOSITORY_OWNER ?= "redhat-developer"
 REPOSITORY_NAME ?= "app-services-cli"
-TERMS_SPEC_URL ?= "https://raw.githubusercontent.com/redhat-developer/app-services-ui/main/static/configs/terms-conditions-spec.json"
+TERMS_SPEC_URL ?= "https://console.redhat.com/apps/application-services/terms-conditions-spec.json"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -5,7 +5,7 @@ SHELL = bash
 RHOAS_VERSION ?= "dev"
 REPOSITORY_OWNER ?= "redhat-developer"
 REPOSITORY_NAME ?= "app-services-cli"
-TERMS_SPEC_URL ?= "https://raw.githubusercontent.com/redhat-developer/app-services-ui/main/static/configs/terms-conditions-spec.json"
+TERMS_SPEC_URL ?= "https://console.redhat.com/apps/application-services/terms-conditions-spec.json"
```
craicoverflow, line 8: We will probably want a different one for stage and prod?

wtrocki: Not sure why we would need it?

craicoverflow: Spec file could be different between stage and prod versions

wtrocki: We do not have ability to detect env and we need more sophisticated remote config for massso. I think this requirement would need more design etc. 

If we need config agains stage I think we should build different version of cli for noe

wtrocki: I think this tc config should be refactored to more generic config. Created issue to cover this work but this is out of scope of current PR

craicoverflow (approved):

PR #1299, opened 2021-11-08
feat(acl list): add flags to filter by resource

rkpattnaik780:
User should be able to list the ACLs per Cluster, Topic or Group.
A small bug fix to ACL list

### Verification Steps

1. Run the command to list permissions for all accounts for topic topic_name
```
./rhoas kafka acl list --all-accounts --topic kakashi 
```
2. Run the command to show all permssions pertaining to group "console-consumer"
```
./rhoas kafka acl list --group console-consumer
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed


wtrocki:
Verified/Reviewed. Changes look good. Pinging @craicoverflow as another SME for correctness

pkg/cmd/kafka/acl/list/list.go
```diff
@@ -141,6 +149,40 @@ func runList(opts *options) (err error) {
...
+	}
+
+	if opts.cluster {
+		selectedResourceTypeCount++
+		resourceType = aclutil.ResourceTypeCLUSTER
+		resourceName = aclutil.KafkaCluster
+	}
+
+	if selectedResourceTypeCount > 1 {
+		return opts.localizer.MustLocalizeError("kafka.acl.common.error.oneResourceTypeAllowed", flagutil.ResourceTypeFlagEntries...)
```
craicoverflow, line 175: This error message specifies transactional-id as an option, but it is not valid here.

```shell
❯ ./rhoas kafka acl list --group "enda" --topic "enda"
❌ Provide only one resource type from "--cluster", "--topic", "--group", or "--transactional-id". Run the command in verbose mode using the -v flag to see more information
```

pkg/cmd/kafka/acl/list/list.go
```diff
@@ -116,6 +121,9 @@ func NewListACLCommand(f *factory.Factory) *cobra.Command {
 	flags.AddUser(&userID)
 	flags.AddServiceAccount(&serviceAccount)
 	flags.AddAllAccounts(&allAccounts)
+	flags.AddCluster(&opts.cluster)
```
craicoverflow, line 124: I re-read the requirements and I could be wrong but I don't think it meant to show only cluster-scoped ACLs, rather be able to list ACLs per cluster:

```shell
rhoas kafka acl list --instance-id=xxxx
```

From the requirements:

> When I am using the `rhoas` CLI,

> I want to list the ACLs per Cluster, Topic and/or Group,

> So I can see which principals are allowed to perform which operations on a given Cluster, Topic and/or Group.

That to me means list the ACLs of a cluster.

Do you agree @rkpattnaik780 @wtrocki? If we are unsure we should clarify with PM.

craicoverflow (changes requested):
Some changes needed. Left a comment regarding whether `--cluster` is needed.

pkg/cmd/kafka/acl/list/list.go
```diff
@@ -116,6 +121,9 @@ func NewListACLCommand(f *factory.Factory) *cobra.Command {
...
+	flags.AddTopic(&opts.topic)
+	flags.AddConsumerGroup(&opts.group)
```
craicoverflow, line 126: We need to add dynamic completions for these.

pkg/cmd/kafka/acl/list/list.go
wtrocki: When you brought that it makes sense but initially I also got it as "cluster" type. 
My take is that more likely to be able to do it by cluster so we might remove it, but would wait for Duncan to confirm.

pkg/cmd/kafka/acl/list/list.go
```diff
@@ -141,6 +149,40 @@ func runList(opts *options) (err error) {
...
+		selectedResourceTypeCount++
+		resourceType = aclutil.ResourceTypeGROUP
+		resourceName = opts.group
+	}
+
+	if opts.cluster {
+		selectedResourceTypeCount++
+		resourceType = aclutil.ResourceTypeCLUSTER
+		resourceName = aclutil.KafkaCluster
+	}
```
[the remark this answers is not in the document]
rkpattnaik780: ValidateAnsSetResources differs a bit - doesnt allow no resource to be passed and has an additional transactional ID. while we can tweak the method a bit, I think we should keep it separate.

pkg/cmd/kafka/acl/list/list.go
```diff
@@ -150,6 +192,12 @@ func runList(opts *options) (err error) {
 		return err
 	}
 
+	if permissionsData.GetTotal() == 0 && opts.output == "" {
+		opts.logger.Info(opts.localizer.MustLocalize("kafka.acl.list.log.info.noACLs", localize.NewEntry("InstanceName", kafkaInstance.GetName())))
```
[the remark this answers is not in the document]
rkpattnaik780: Kept it similar to what we return for topic list. We should think of an appropriate message and update it across suitable list commands.

pkg/cmd/kafka/acl/list/list.go
rkpattnaik780: Going through requirements I think the same, thanks for the catch!

pkg/cmd/kafka/acl/list/list.go
rkpattnaik780: Will it be good to add an additional field `factory` in the [flagSet struct](https://github.com/redhat-developer/app-services-cli/blob/eb27c0f575a816945cd75d4f9ce777c2dea28fee/pkg/cmd/kafka/acl/flagutil/flagset.go#L25)? That will make it easier to reuse completions for topic and group from cmdutils?

craicoverflow: Yeah, that would be good. You can remove the localizer and conn fields so and just use this :)

I'd recommend refactoring these completions APIs also to look like this one: https://github.com/redhat-developer/app-services-cli/blob/8edfc44c74d1b21974ad97ae0d5c096c61818587/pkg/kafka/cmdutil/util.go#L11

They also probably should not be in a generic cmdutil package since they are kafka specific (I did that 😂)

craicoverflow:
This is ready for review @rkpattnaik780?

rkpattnaik780:
> This is ready for review @rkpattnaik780?

Yes.

craicoverflow:
> > This is ready for review @rkpattnaik780?
> 
> Yes.

Thanks.

wtrocki:
Documentation needs little bit refinement:
I would use some common prefix to say what flag does - mainly filtering. 

Set filter to all-accounts
Set filter to only cluster type

etc.

Current docs are less refined


```
Flags:
      --all-accounts             Set the ACL principal to match all principals (users and service accounts)
      --cluster                  Set the resource type to cluster
      --group string             Set the consumer group resource. When the --prefix option is also passed, this is used..
      --service-account string   Service account client ID used as principal for this operation
```

wtrocki (approved):

wtrocki:
Documentation needs improvement. Otherwise ok.

rkpattnaik780:
> Documentation needs little bit refinement: I would use some common prefix to say what flag does - mainly filtering.
> 
> Set filter to all-accounts Set filter to only cluster type
> 
> etc.
> 
> Current docs are less refined
> 
> ```
> Flags:
>       --all-accounts             Set the ACL principal to match all principals (users and service accounts)
>       --cluster                  Set the resource type to cluster
>       --group string             Set the consumer group resource. When the --prefix option is also passed, this is used..
>       --service-account string   Service account client ID used as principal for this operation
> ```

Falling back to not using flag abstraction here as the descriptions differ a bit owing to `--prefix` flag.

rkpattnaik780:
Having second thoughts adding a prefix regarding filter operation, wdyt @craicoverflow ?

pkg/kafka/cmdutil/util.go
```diff
@@ -90,3 +90,88 @@ func GetEnabledCloudRegionIDs(regions []kafkamgmtclient.CloudRegion) []string {
 	}
 	return regionIDs
 }
+
+// FilterValidTopicNameArgs filters topics from the API and returns the names
+// This is used in for dynamic completion of topic names
+func FilterValidTopicNameArgs(f *factory.Factory, toComplete string) (validNames []string, directive cobra.ShellCompDirective) {
```
craicoverflow, line 96: nit: these could go in kafka flag util package?

pkg/localize/locales/en/cmd/acl.en.toml
```diff
@@ -154,6 +163,15 @@ The ACLs are displayed in a table by default. Alternatively, you can display the
 [kafka.acl.list.flag.output.description]
 one = 'Format in which to display the Kafka ACL rules (choose from: "json", "yml", "yaml")'
```
craicoverflow, line 164: We have a common out flag now with its own description which could be used.

pkg/cmd/kafka/acl/list/list.go
```diff
@@ -116,6 +122,17 @@ func NewListACLCommand(f *factory.Factory) *cobra.Command {
 	flags.AddUser(&userID)
 	flags.AddServiceAccount(&serviceAccount)
 	flags.AddAllAccounts(&allAccounts)
+	flags.BoolVar(&opts.cluster, "cluster", false, opts.localizer.MustLocalize("kafka.acl.list.flag.cluster.description"))
```
craicoverflow, line 125: Cluster flag should be gone?

craicoverflow (commented):
All looks good and works. However I noticed that the cluster flag still exists and I think we agreed it is not a requirement.

craicoverflow (approved):

Issue #1290, opened 2021-11-05
Add abstraction of download source


craicoverflow:
## The problem

Currently, the CLI assumes the download source to be from upstream (GitHub release).

We now have two possible download sources: GitHub and a productised version which will be downloadable from developers.redhat.com. In the future there could be others such as DNF repositories or Brew.

There is a version checking system in place which checks only GitHub for a new version. This will produce invalid info if the user has downloaded the CLI from a different source.

## The solution

First, we need to add a build variable to indicate the build/download source. Values can be `github`, `redhat` etc.

Then, we should have an interface defining some common methods: `CheckForUpdate` initially to replicate what we have at present. There would need to be an implementation for every download source.
It could be expanded in the future to fully automate version updates from within the CLI.

wtrocki:
My preference would be to not mix upstream releases with product. Product team should have their own answers to brew, dnf etc. We should not do it on our own and follow what other CLI from product use.

Personal opinion
 Install script is better remain github/upstream especially since we are still not 1.0 and we require people to update very often from github. Doing so many downstream releases will be unpractical and cumbersome

wtrocki:
I do not like product leaking to upstream and also want to push to get value of productization process. Setting up call to discuss how others planning to do it as I'm aware there are discussions about generic cli publishing space thru product

craicoverflow:
I think you are missing the point of this. This is nothing to do with dnf or yum. We have an upstream and product release, but the CLI is internally checking for new versions against upstream..t hat check should be different for the productised CLI.

PR #1303, opened 2021-11-09
refactor: check for new version in upstream build only

craicoverflow:
This is a stop-gap solution to #1290 by ensuring that version checking and printiing will only occur for the upstream CLI, when the `BuildSource` build variable is "github".

### DIFF
--- a/internal/build/build.go
+++ b/internal/build/build.go
@@ -2,6 +2,7 @@ package build
 
 import (
 	"context"
+	"fmt"
 	"runtime/debug"
 	"time"
 
@@ -9,6 +10,12 @@ import (
 	"github.com/redhat-developer/app-services-cli/pkg/color"
 	"github.com/redhat-developer/app-services-cli/pkg/localize"
 	"github.com/redhat-developer/app-services-cli/pkg/logging"
+)
+
+type buildSource string
+
+const (
+	githubBuildSource buildSource = "github"
 )
 
 // Define public variables here which you wish to be configurable at build time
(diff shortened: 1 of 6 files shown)


wtrocki:
That change gives nice context for what #1290 was needed. 
Yes. Automatic updates should be github oriented for now as downstream will use package managers like brew for that purpose.

wtrocki:
Build it by default as dev. Build it as github and with some other random flag. It seems to work

internal/build/build.go
```diff
@@ -68,6 +78,11 @@ func init() {
 // the version currently being used. If so, it logs this information
 // to the console.
 func CheckForUpdate(ctx context.Context, logger logging.Logger, localizer localize.Localizer) {
+	if BuildSource != string(githubBuildSource) {
+		return
+	}
+	fmt.Println("Checking for a new version")
```
wtrocki, line 84: i18n?

craicoverflow: I had that in for testing, needs to go.

internal/build/build.go
```diff
@@ -67,7 +77,15 @@ func init() {
...
 // to the console.
-func CheckForUpdate(ctx context.Context, logger logging.Logger, localizer localize.Localizer) {
+func CheckForUpdate(ctx context.Context, version string, logger logging.Logger, localizer localize.Localizer) {
+	if BuildSource != string(githubBuildSource) {
+		return
+	}
+	// prefix version with a v to correspond with Git tag
+	if !strings.HasPrefix(version, "v") {
+		version = "v" + version
+	}
```
craicoverflow, line 87: Turns out version checks had not been working because of differences in Git tag and version.

wtrocki (approved):

internal/build/build.go
wtrocki: Any way I can verify this locally? 

It looks like good fix

craicoverflow:
```shell
BUILD_SOURCE=github RHOAS_VERSION=0.30.0 make binary
rhoas version
```

wtrocki:
All good!

Issue #1224, opened 2021-10-14
Error message should make it clear `--topic-prefix` or `--topic` can be used


craicoverflow:
In the `kafka acl grant-access` command:

`--topic` OR `--topic-prefix` can be passed, but error messages do not make this known - they say "--topic" is required.

The error message must reflect the allowed values.

```shell
$ rhoas kafka acl grant-access --producer --user ephelan_kafka_devexp
❌ "--topic" flag is required for consumer and producer operations. Run the command in verbose mode using the -v flag to see more information
```

gkpani97:
Hello @craicoverflow , I'm interested in solving this issue, can you assign it to me.

PR #1311, opened 2021-11-11
fix(kafka update): broken preview

### LINKED ISSUE #1302: Broken update summary for Kafka update operation
## Description

Update summary while doing owner update seems to be broken.

**Version**: 0.36.2

Screenshot from 2021-11-09 13-35-36

### Steps to reproduce

Change owner for a given kafka instance
```
rhoas kafka update --owner=foo_user
```
It should display the update summary.

rkpattnaik780:
Closes #1302, #1224

### Verification Steps
1. Run `rhoas kafka update --owner foo_user`
2. This should display the existing user and the new user to be set

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/kafka/update/update.go
+++ b/pkg/cmd/kafka/update/update.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	kafkamgmtclient "github.com/redhat-developer/app-services-sdk-go/kafkamgmt/apiv1/client"
@@ -281,20 +282,23 @@ func promptConfirmUpdate(opts *options) (bool, error) {
 // creates a summary of what values will be changed in this update
 // returns a formatted string. Example:
 // owner: foo_user	➡️	bar_user
+// to do: Needs update once reauthentication_enabled is incorporated
 func generateUpdateSummary(new reflect.Value, current reflect.Value) string {
 	var summary string
 
 	newT := new.Type()
 
-	for i := 0; i < new.NumField(); i++ {
+	for i := 0; i < 1; i++ {
 		field := newT.Field(i)
 		fieldTag := field.Tag.Get("json")
 		fieldName := field.Name
 
 		oldVal := getElementValue(current.FieldByName(fieldName)).String()
-		newVal := getElementValue(new.FieldByName(fieldName)).String()
+		newVal := getUpdateObjValue(new.FieldByName(fieldName)).String()
 
-		summary += fmt.Sprintf("%v: %v    %v    %v\n", color.Bold(fieldTag), oldVal, icon.Emoji("\u27A1", "=>"), newVal)
+		fieldTagDisp := strings.Split(fieldTag, ",")[0]
+
+		summary += fmt.Sprintf("%v: %v    %v    %v\n", color.Bold(fieldTagDisp), oldVal, icon.Emoji("\u27A1", "=>"), newVal)
 	}
 
 	return summary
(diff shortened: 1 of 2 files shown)


rkpattnaik780:
Added a few todos for when cli will work with `ReauthenticationEnabled`. Currently handles only string values, needs modification to support other types later.

pkg/cmd/kafka/update/update.go
```diff
@@ -281,20 +282,23 @@ func promptConfirmUpdate(opts *options) (bool, error) {
 // creates a summary of what values will be changed in this update
 // returns a formatted string. Example:
 // owner: foo_user	➡️	bar_user
+// to do: Needs update once reauthentication_enabled is incorporated
```
wtrocki, line 285: How reauthentication_enabled affects this preview?

rkpattnaik780: The methods generating preview iterate through the entire update struct, the broken preview did display `reauthentication_enabled` too.

pkg/cmd/kafka/update/update.go
```diff
@@ -281,20 +282,23 @@ func promptConfirmUpdate(opts *options) (bool, error) {
...
 func generateUpdateSummary(new reflect.Value, current reflect.Value) string {
 	var summary string
 
 	newT := new.Type()
 
-	for i := 0; i < new.NumField(); i++ {
+	for i := 0; i < 1; i++ {
```
craicoverflow, line 291: I'm just curious what this does, how it fixes it?

craicoverflow (dismissed):

pkg/cmd/kafka/update/update.go
rkpattnaik780: This is to hide the `reauthentication_nabled` from preview.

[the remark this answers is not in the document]
wtrocki: It looks like in is not stable/subject to break when api return new data

craicoverflow: Yes I understand _why_ it was done, I just did not yet understand _how_. I've debugged locally to understand it now.

This is a "quick fix" but once reauth update is possible we won't be able to just revert this change. If the update does not include and change to `reauthentication_enabled`, reverting this will still include it in the update summary.

I see that the value of ReauthenticationEnabled is nil, so perhaps we can check for nil before including it in the update summary?

This also assumes that the owner field will always be the first property of the update object, we should not rely on this.

pkg/cmd/kafka/update/update.go
wtrocki: Yes. That iteration thru all elements of struct is core issue IMHO

craicoverflow: We need to filter out the nil properties of the update object.

pkg/cmd/kafka/update/update.go
```diff
@@ -308,3 +314,13 @@ func getElementValue(v reflect.Value) reflect.Value {
 	}
 	return v
 }
+
+// get the true value from a reflect.Value for KafkaUpdateRequest struct
+// to do: Needs update once reauthentication_enabled is incorporated
+func getUpdateObjValue(v reflect.Value) reflect.Value {
+	if v.Kind() == reflect.Struct {
+		vstruct := v.Field(0)
```
rkpattnaik780, line 322: A better waywould be to get the `value` field instead of index. Do we need to iterate through this struct as well?

craicoverflow: This should work:

```suggestion
		vstruct := v.FieldByName("value")
```

This should suffice, no need to iterate.

Though I'm starting to think going forward with reflection on this is a bad idea, as we are making assumptions that are tightly coupled with the SDK, which could introduce breaking changes in the future by changing the name from "value" and since this is not strongly typed, we would not catch it..

For now, once you update with the suggestion above, we can merge as it is blocking any releases. Could you create a follow up issue to address this? We can simplify it by operating directly on the SDK model properties instead.

rkpattnaik780: Done!
Ready for review.

craicoverflow (approved):

Issue #1318, opened 2021-11-16
Wrong help test for `--output` in `service-registry artifact create`


rkpattnaik780:
## Description

**Version**: 0.36.3

The `--output` flag for `service-registry artifact create` has description - Format in which to display the Service Registry instance.
It should be changed to : `Format in which to display the artifact`

![Screenshot from 2021-11-16 00-43-01](https://user-images.githubusercontent.com/23582438/141949703-7dc1ef8e-7bfb-4335-b5e2-424f9ac2c4aa.png)

### Steps to reproduce

1. Run `rhoas service-registry artifact create -h`

craicoverflow:
We should use the common flag set for all `output` and `yes` flags:

https://github.com/redhat-developer/app-services-cli/blob/060f870988c2bc77fabc4ea1d3e9c5a95d2b7fce/pkg/cmdutil/flagutil/flagset.go#L35-L66

The approach can be pretty much cloned from here: https://github.com/redhat-developer/app-services-cli/blob/e56d44356618fab98b3db1c8c76b22d58232798e/pkg/cmd/kafka/topic/create/create.go#L130-L131

While these common flags become less unique in their individual descriptions, it should be much more sustainable.

rkpattnaik780:
> We should use the common flag set for all `output` and `yes` flags:

Would you like to create a separate issue for this task?

PR #1319, opened 2021-11-16
fix: output format for create artifact cmd

wtrocki:
Fixes https://github.com/redhat-developer/app-services-cli/issues/1318

### DIFF
--- a/pkg/cmd/registry/artifact/crud/create/create.go
+++ b/pkg/cmd/registry/artifact/crud/create/create.go
@@ -98,7 +98,7 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cmd/registry/artifact/crud/create/create.go
```diff
@@ -98,7 +98,7 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("registry.cmd.flag.output.description"))
+	cmd.Flags().StringVarP(&opts.outputFormat, "output", "o", "json", opts.localizer.MustLocalize("artifact.common.message.output.format"))
```
wtrocki, line 101: This uses the same flag as rest of the artefact commands

PR #1326, opened 2021-11-17
RHOAS Profile configiuration

wtrocki:
Creating for visibility. This is early POC! Not to be reviewed or used.

Instructions and usage: https://github.com/rhoas-examples/rhoas-cli-config-example


wtrocki:
That was POC. I'm not sure if we should have it as PR :)

PR #1327, opened 2021-11-17
fix(cluster connect): print access commands separately for services

rkpattnaik780:
Display commands required to grant access to the service-account.

### Verification Steps

1. Connect a service registry instance with the cluster:
```
rhoas cluster connect --service-type service-registry --namespace default -y
```
It should print command to give access to the the service registry

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cluster/connect.go
+++ b/pkg/cluster/connect.go
@@ -89,7 +89,7 @@ func (api *KubernetesClusterAPIImpl) ExecuteConnect(connectOpts *v1alpha.Connect
 		return kubeclient.TranslatedKubernetesErrors(api.CommandEnvironment, err)
 	}
 
-	err = api.createServiceAccountSecretIfNeeded(currentNamespace)
+	err = api.createServiceAccountSecretIfNeeded(currentNamespace, currentService)
 	if err != nil {
 		return kubeclient.TranslatedKubernetesErrors(api.CommandEnvironment, err)
 	}
@@ -181,7 +181,7 @@ func (c *KubernetesClusterAPIImpl) createTokenSecretIfNeeded(namespace string, a
 }
 
 // createSecret creates a new secret to store the SASL/PLAIN credentials from the service account
-func (c *KubernetesClusterAPIImpl) createServiceAccountSecretIfNeeded(namespace string) error {
+func (c *KubernetesClusterAPIImpl) createServiceAccountSecretIfNeeded(namespace string, currentService services.RHOASKubernetesService) error {
 	cliOpts := c.CommandEnvironment
 	kClients := c.KubernetesClients
 	ctx := cliOpts.Context
@@ -217,6 +217,8 @@ func (c *KubernetesClusterAPIImpl) createServiceAccountSecretIfNeeded(namespace 
 		localize.NewEntry("Name", createdSecret.Name),
 		localize.NewEntry("ClientID", serviceAcct.GetClientId()),
 	))
+
+	currentService.PrintAccessCommands(serviceAcct.GetClientId())
 
 	return nil
 }
--- a/pkg/cluster/connect.go
+++ b/pkg/cluster/connect.go
@@ -89,9 +89,13 @@ func (api *KubernetesClusterAPIImpl) ExecuteConnect(connectOpts *v1alpha.Connect
 		return kubeclient.TranslatedKubernetesErrors(api.CommandEnvironment, err)
 	}
 
-	err = api.createServiceAccountSecretIfNeeded(currentNamespace, currentService)
+	clientID, err := api.createServiceAccountSecretIfNeeded(currentNamespace)
 	if err != nil {
 		return kubeclient.TranslatedKubernetesErrors(api.CommandEnvironment, err)
+	}
+
+	if clientID != "" {
+		currentService.PrintAccessCommands(clientID)
 	}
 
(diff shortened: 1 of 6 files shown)


wtrocki (approved):

wtrocki:
Still verifying but code changes look good!

pkg/localize/locales/en/cmd/cluster.en.toml
```diff
@@ -301,11 +305,21 @@ Client ID:     {{.ClientID}}
 Make a copy of the client ID to store in a safe place. Credentials won't appear again after closing the terminal.
 
 You will need to assign permissions to service account in order to use it. 
-For example for Kafka service you should execute the following command to grant access to the service account:
+'''
+
+[cluster.kubernetes.printKafkaAccessCommands]
+one = '''
+For Kafka service you should execute the following command to grant access to the service account:
```
wtrocki, line 312: After verifying I found this quite odd message.
I'm not sure if there is benefit of mentioning service name here due

I will change it to something like 

```
You need to grant access to you Kafka separately by issuing following command
```

pkg/localize/locales/en/cmd/cluster.en.toml
```diff
@@ -301,11 +305,21 @@ Client ID:     {{.ClientID}}
...
 
   $ rhoas kafka acl grant-access --producer --consumer --service-account {{.ClientID}} --topic all --group all
 '''
- 
+
+[cluster.kubernetes.printRegistryAccessCommands]
+one = '''
+For service registry you should execute the following command to grant access to the service account:
```
wtrocki, line 319: Same as above. 
For registry we do not grant access but `assign new role`. Registry by default will be read only. We need to mention that so if you only need read only access you do not need to use this command

wtrocki (approved):
Verified. Small improvements needed in messages. Otherwise ok

wtrocki:
> One query, things will get tricky when context mode is not used, while we can modify the command for registry, the issue will persist with Kafka.

Not sure I understand. Is there still some work to be done?

pkg/cluster/connect.go
```diff
@@ -218,6 +218,8 @@ func (c *KubernetesClusterAPIImpl) createServiceAccountSecretIfNeeded(namespace
 		localize.NewEntry("ClientID", serviceAcct.GetClientId()),
 	))
 
+	currentService.PrintAccessCommands(serviceAcct.GetClientId())
```
craicoverflow, line 221: My own feeling is that this does not belong in this package and should be printed from the connect command. This package is a Kubernetes API client and by print stuff to the console it is mixing responsibility layers.

wtrocki: I haven't seen requirement to have this as kube client. I know we have this interface for it but IMHO this is slightly over-designing something that is very tightly related to CLI command. There are many places in services where we do the same, use context etc. 
Sole purpose of this package is to serve as execution layer for commands.

pkg/cluster/services/kafka.go
```diff
@@ -78,3 +79,8 @@ func (s KafkaService) BuildServiceDetails(serviceName string, namespace string,
 
 	return &serviceDetails, nil
 }
+
+func (s KafkaService) PrintAccessCommands(clientID string) {
```
craicoverflow, line 83: All public methods should be documented.

pkg/cluster/connect.go
craicoverflow: The thing that highlighted this for me is that we are passing down `currentService` as a new function argument for the sole purpose of printing the access commands. It does not relate to the core purpose of the function (creating a service when needed) and goes against the single-responsibility principle.

wtrocki: Funny. Rama wanted it this way and I have recomended to do it in function also quoting single responsibility rule (dealing with service accounts)

#opinions :)

wtrocki: @rkpattnaik780 to make final call on this

craicoverflow: Yeah, same result at the end of the day :)

rkpattnaik780: I agree with the method part, have updated the method to return a client ID if service account is created and print from connect command.

pkg/cluster/services/kafka.go
```diff
@@ -80,6 +80,7 @@ func (s KafkaService) BuildServiceDetails(serviceName string, namespace string,
 	return &serviceDetails, nil
 }
 
+// Print command to grant service account acccess to the Kafka instance
 func (s KafkaService) PrintAccessCommands(clientID string) {
```
craicoverflow, line 84: All public methods should start with the function name: `// PrintAccessCommand does x`

pkg/cluster/connect.go
```diff
@@ -89,11 +89,15 @@ func (api *KubernetesClusterAPIImpl) ExecuteConnect(connectOpts *v1alpha.Connect
 		return kubeclient.TranslatedKubernetesErrors(api.CommandEnvironment, err)
 	}
 
-	err = api.createServiceAccountSecretIfNeeded(currentNamespace, currentService)
+	clientID, err := api.createServiceAccountSecretIfNeeded(currentNamespace)
```
craicoverflow, line 92: :)

craicoverflow (approved):

PR #1331, opened 2021-11-23
chore: add a self-maintaining help command to Makefile

craicoverflow:
This trick found [in this article](http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html) introspects the Makefile to dynamically create an output for `make help`. 
I've changed some commands to not include a `/` as it did not work for these.

### Verification Steps

Run `make help`


rkpattnaik780 (changes requested):
Make help works great!
`check-docs` and `generate-modular-docs` fail with `No rule to make target 'docs/generate'`

rkpattnaik780 (approved):

PR #1335, opened 2021-11-24
fix: remove dev preview

wtrocki:
WIP

### DIFF
--- a/docs/commands/rhoas.adoc
+++ b/docs/commands/rhoas.adoc
@@ -82,10 +82,10 @@ ifdef::pantheonenv[]
 endif::[]
 
 ifdef::env-github,env-browser[]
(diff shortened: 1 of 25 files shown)


wtrocki:
@smccarthy-ie will finish that tommorow.

wtrocki:
@smccarthy-ie is that ok to be released today?

smccarthy-ie:
> @smccarthy-ie is that ok to be released today?

@wtrocki Thanks, yes ideal to merge this today so we will have time to check content before publication on Dec 1.

wtrocki:
I meant if we ok to release CLI with that help changes today. I guess that is not a issue right?

smccarthy-ie:
OK, thanks. Don't think this is an issue.

PR #1336, opened 2021-11-26
fix: broken mock make target

wtrocki:


wtrocki:
FYI @craicoverflow . Also updated docs build pipeline.

PR #1155, opened 2021-09-29
chore: added a new go-i18n-linter

alexal:
Closes #1076 

### Verification Steps
1. Run `make lint`
2. Verify the output. As of right now, there should be three strings that doesn't exist in *.toml files

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Documentation added for the feature
- [ ] CI and all relevant tests are passing
- [ ] Code Review completed
- [ ] Verified independently by reviewer

As I already mentioned here https://github.com/redhat-developer/app-services-cli/issues/1076#issuecomment-925750071 the plugin is definitely working once I use it with golangci-lint that I built locally. 

@craicoverflow, @wtrocki I need your help with finding a solution for that.

PR #1338, opened 2021-11-29
chore: add custom i18n linter

wtrocki:


wtrocki:
This downloads and compiles linter as part of the make target. 
Alternative to commiting binary to repo as in https://github.com/redhat-developer/app-services-cli/pull/1155

I think that we can even simplify this by doing `go get or something`

wtrocki:
Note: I changed binary to something simpler.

wtrocki:
I'm getting strange error in cli just error code. @alexal I know it is been a while but do you remember all required arguments to get things running.

Is running binary in root of cli without arguments ok?

Makefile
```diff
@@ -92,7 +92,14 @@ generate-modular-docs: generate-docs ## Generate modular command-line reference
 	SRC_DIR=$$(pwd)/docs/commands DEST_DIR=$$(pwd)/dist go run ./cmd/modular-docs
 .PHONY: generate-modular-docs
 
+lint-lang:
```
craicoverflow, line 95: Needs documentation to appear in the help command output (see others for example).

Makefile
```diff
@@ -92,7 +92,14 @@ generate-modular-docs: generate-docs ## Generate modular command-line reference
...
+	git clone git@github.com:redhat-developer/app-services-go-linter
```
craicoverflow, line 96: `go install github.com/redhat-developer/app-services-go-linter`

Makefile
```diff
@@ -92,7 +92,14 @@ generate-modular-docs: generate-docs ## Generate modular command-line reference
...
+	cd app-services-go-linter && make binary
+	app-services-go-linter/rhoas-lint
+.PHONY: lint-lang
```
craicoverflow, line 99: It seems a little hacky. It would be better to be able to include it as part of the overall lint configuration. Is that achievable or were there issues?

Makefile
wtrocki: I could not make it work:

go install github.com/redhat-developer/app-services-go-linter
no required module provides package github.com/redhat-developer/app-services-go-linter; to add it:
        go get github.com/redhat-developer/app-services-go-linter

craicoverflow: It needs to be to the path to the main package:

```shell
go install github.com/redhat-developer/app-services-go-linter/cmd/app-services-go-linter@latest
```

wtrocki: Nice. Fixing it

Makefile
wtrocki: Yes. We should do that in the long run.

alexal:
> I'm getting strange error in cli just error code. @alexal I know it is been a while but do you remember all required arguments to get things running.
> 
> Is running binary in root of cli without arguments ok?

If I'm not wrong, you have to provide one argument, which is `./...`

wtrocki:
@craicoverflow @rkpattnaik780 

This is ready for review - since we were actually missing keys I would consider this important PR.

As note why this is separate target and not golang cli lint. 

The binary that we build doesn't seem to work well as custom linter - it does not print out proper errors.
I do not think there is benefit for us to actually have that in golangci lint as we still need to push binary to repo or compile it in the fly like we do here.

Makefile
```diff
@@ -92,7 +92,13 @@ generate-modular-docs: generate-docs ## Generate modular command-line reference
 	SRC_DIR=$$(pwd)/docs/commands DEST_DIR=$$(pwd)/dist go run ./cmd/modular-docs
 .PHONY: generate-modular-docs
 
+lint-lang: ## Lint i18n files
```
craicoverflow, line 95:
```suggestion
lint-lang: ## Lint i18n files for missing keys
```

pkg/localize/locales/en/cmd/login.en.toml
```diff
@@ -113,7 +113,7 @@ one = 'Logged in successfully to {{.Host}}'
 [login.error.noRealmInURL]
 one = 'the authentication URL is missing a realm'
 
-[login.log.debug.sshLoginDetected]
+[login.log.info.sshLoginDetected]
```
craicoverflow, line 116: I feel like that should be a debug message as users _may_ experience issues logging in, but they also may not depending on whether they have a UI connected to their SSH session.

rkpattnaik780 (approved):

pkg/localize/locales/en/cmd/login.en.toml
wtrocki: agree!

PR #1340, opened 2021-12-01
feat(kafka update): add reauthentication flag

rkpattnaik780:
Add `--reauthentication` flag to Kafka update

### Verification Steps

1. Set value for "enable_reauthentication" using the `--reauthentication` flag
`rhoas kafka update --reauthentication false`
2. Try running kafka update in interactive mode

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/kafka/update/update.go
+++ b/pkg/cmd/kafka/update/update.go
@@ -216,6 +216,10 @@ func run(opts *options) error {
 		updateSummary,
 	)
 
+	if opts.reauth == "false" {
+		opts.logger.Info(opts.localizer.MustLocalize("kafka.update.reauthentication.disclaimer"))
+	}
+
 	if !opts.skipConfirm {
 		//nolint:govet
 		confirm, err := promptConfirmUpdate(opts)
--- a/pkg/cmd/kafka/update/update.go
+++ b/pkg/cmd/kafka/update/update.go
@@ -40,6 +40,7 @@ type options struct {
 
 	interactive    bool
 	userIsOrgAdmin bool
+	enableReauth   string
 
 	IO         *iostreams.IOStreams
 	Config     config.IConfig
@@ -90,7 +91,7 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 					return flag.RequiredWhenNonInteractiveError(missingFlags...)
 				}
 			}
-			if opts.owner == "" {
+			if opts.owner == "" && opts.enableReauth == "" {
 				opts.interactive = true
 			}
 
@@ -116,6 +117,7 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 
 	flags.StringVar(&opts.id, "id", "", opts.localizer.MustLocalize("kafka.update.flag.id"))
 	flags.StringVar(&opts.owner, "owner", "", opts.localizer.MustLocalize("kafka.update.flag.owner"))
+	flags.StringVar(&opts.enableReauth, "enable-reauth", "", opts.localizer.MustLocalize("kafka.update.flag.enableReauth"))
 	flags.AddYes(&opts.skipConfirm)
 	flags.StringVar(&opts.name, "name", "", opts.localizer.MustLocalize("kafka.update.flag.name"))
 
@@ -125,6 +127,7 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 	return cmd
 }
 
+// nolint:funlen
 func run(opts *options) error {
 	conn, err := opts.Connection(connection.DefaultConfigSkipMasAuth)
 	if err != nil {
@@ -132,6 +135,9 @@ func run(opts *options) error {
 	}
 
 	api := conn.API()
+
+	// track if any values have changed
+	var needsUpdate bool
 
 	var kafkaInstance *kafkamgmtclient.KafkaRequest
 	if opts.name != "" {
@@ -153,15 +159,40 @@ func run(opts *options) error {
 		if err != nil {
 			return err
 		}
-	}
 
-	if opts.owner == kafkaInstance.GetOwner() {
-		opts.logger.Info(opts.localizer.MustLocalize("kafka.update.log.info.sameOwnerNoChanges", localize.NewEntry("Owner", opts.owner), localize.NewEntry("InstanceName", kafkaInstance.GetName())))
-		return nil
+		enableReauthPrompt := &survey.Select{
+			Message: opts.localizer.MustLocalize("kafka.update.input.message.enableReauth"),
+			Options: []string{"true", "false"},
+			Default: strconv.FormatBool(kafkaInstance.GetReauthenticationEnabled()),
+		}
+
+		err = survey.AskOne(enableReauthPrompt, &opts.enableReauth)
+		if err != nil {
+			return err
+		}
 	}
 
 	updateObj := kafkamgmtclient.NewKafkaUpdateRequest()
-	updateObj.SetOwner(opts.owner)
+
+	if opts.owner != "" && opts.owner != kafkaInstance.GetOwner() {
+		updateObj.SetOwner(opts.owner)
+		needsUpdate = true
+	}
+
+	if opts.enableReauth != "" && opts.enableReauth != strconv.FormatBool(kafkaInstance.GetReauthenticationEnabled()) {
+		enableBool, newErr := strconv.ParseBool(opts.enableReauth)
+		if newErr != nil {
+			return newErr
+		}
+
+		updateObj.SetReauthenticationEnabled(enableBool)
+		needsUpdate = true
+	}
+
+	if !needsUpdate {
+		opts.logger.Info(opts.localizer.MustLocalize("kafka.update.log.info.nothingToUpdate"))
+		return nil
+	}
 
 	// create a text block with a summary of what is being updated
 	updateSummary := generateUpdateSummary(reflect.ValueOf(*updateObj), reflect.ValueOf(*kafkaInstance))
--- a/pkg/cmd/kafka/update/update.go
+++ b/pkg/cmd/kafka/update/update.go
@@ -2,7 +2,6 @@ package update
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -221,10 +220,9 @@ func run(opts *options) error {
 	}
 
 	if !opts.skipConfirm {
-		//nolint:govet
-		confirm, err := promptConfirmUpdate(opts)
-		if err != nil {
-			return err
+		confirm, promptErr := promptConfirmUpdate(opts)
+		if promptErr != nil {
+			return promptErr
 		}
 		if !confirm {
 			opts.logger.Debug("User has chosen to not update Kafka instance")
@@ -248,9 +246,8 @@ func run(opts *options) error {
 	s.Stop()
 
 	if err != nil {
-		opts.logger.Info(opts.localizer.MustLocalize("kafka.update.log.info.updateFailed"))
 		if apiError := kas.GetAPIError(err); apiError != nil {
-			return errors.New(apiError.GetReason())
+			return opts.localizer.MustLocalizeError("kafka.update.log.info.updateFailed", localize.NewEntry("Reason", apiError.GetReason()))
 		}
 		return err
 	}
--- a/pkg/cmd/kafka/update/update.go
+++ b/pkg/cmd/kafka/update/update.go
@@ -40,7 +40,7 @@ type options struct {
 
 	interactive    bool
 	userIsOrgAdmin bool
-	enableReauth   string
+	reauth         string
 
 	IO         *iostreams.IOStreams
 	Config     config.IConfig
@@ -49,6 +49,10 @@ type options struct {
 	localizer  localize.Localizer
 	Context    context.Context
 }
+
+var (
+	validAuthOptions = []string{"true", "false"}
+)
 
 func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 	opts := options{
@@ -82,6 +86,10 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 				return nil
 			}
 
+			if opts.reauth != "" && !flagutil.IsValidInput(opts.reauth, validAuthOptions...) {
+				return flag.InvalidValueError("reauthentication", opts.reauth, validAuthOptions...)
+			}
+
 			if !opts.IO.CanPrompt() {
 				var missingFlags []string
 				if !opts.skipConfirm {
@@ -91,7 +99,7 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 					return flag.RequiredWhenNonInteractiveError(missingFlags...)
 				}
 			}
-			if opts.owner == "" && opts.enableReauth == "" {
+			if opts.owner == "" && opts.reauth == "" {
 				opts.interactive = true
 			}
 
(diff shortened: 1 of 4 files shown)


pkg/cmd/kafka/update/update.go
```diff
@@ -125,6 +127,7 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 	return cmd
 }
 
+// nolint:funlen
```
wtrocki, line 130: Ideally this should be removed before we merge.

pkg/cmd/kafka/update/update.go
```diff
@@ -153,15 +159,40 @@ func run(opts *options) error {
...
+		err = survey.AskOne(enableReauthPrompt, &opts.enableReauth)
+		if err != nil {
+			return err
+		}
 	}
 
 	updateObj := kafkamgmtclient.NewKafkaUpdateRequest()
-	updateObj.SetOwner(opts.owner)
+
+	if opts.owner != "" && opts.owner != kafkaInstance.GetOwner() {
```
wtrocki, line 177: This logic missing else and `"kafka.update.log.info.sameOwnerNoChanges` that were there previously. 
When adding flags we can eather:

- Make command execute only single operation per flag (owner, reauth)
- Make command execute multiple actions at the same time - will require priorities and decision how command should fail - Aggregation of errors that we done for acl etc.

pkg/cmd/kafka/update/update.go
```diff
@@ -116,6 +117,7 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 
 	flags.StringVar(&opts.id, "id", "", opts.localizer.MustLocalize("kafka.update.flag.id"))
 	flags.StringVar(&opts.owner, "owner", "", opts.localizer.MustLocalize("kafka.update.flag.owner"))
+	flags.StringVar(&opts.enableReauth, "enable-reauth", "", opts.localizer.MustLocalize("kafka.update.flag.enableReauth"))
```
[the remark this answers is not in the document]
wtrocki: Although vocabulary is not my strong point our flags should probably stick with nouns.
Also auth as shortcut doesn't carry that much info so my take is to drop enable and extend it to the full sentence.

reauthentication=true

pkg/cmd/kafka/update/update.go
rkpattnaik780: > will require priorities and decision how command should fail - Aggregation of errors that we done for acl etc.
I think this is something backend takes care of already, running a `kafka update --owner inexistent_user --reauthentication false` fails with error `user doesn't belong to your organization` and doesn't update the enable_reauthtentication field

wtrocki: Yes, but what is the plan to handle this in the CLI?

Do we want to even send request if we know that it will be wrong?
Do we provide info to user that error in owner also discarded other update (that might be assumed, but better to be explicit)

rkpattnaik780: > Do we want to even send request if we know that it will be wrong?

While we can check by fetching the list of users before sending request for update. I think we should let the backend check it for us.

> Do we provide info to user that error in owner also discarded other update

I would prefer not to be explicit about the discard as update sounds very much atomic. I will let you take the final call on it, maybe we could display a simple message like `update aborted`.

wtrocki: Please use your jugement based on my review. I just point out that relying on server side (especially server doesn't provide reliable error codes for that case that will help us to provide proper message to user) is bad UX. 

See entire terms and conditions check - backend does that but we still check that before :)

rkpattnaik780: current preview:
Screenshot from 2021-12-02 13-20-22

pkg/localize/locales/en/cmd/kafka_update.en.toml
```diff
@@ -36,30 +39,49 @@ one = 'Name of the Kafka instance you want to update'
...
+
 [kafka.update.flag.yes]
 one = 'Forcibly update the Kafka instance without confirmation'
 
 [kafka.update.summaryTitle]
 one = 'Summary of Proposed Changes'
 
+[kafka.update.reauthentication.disclaimer]
+one = '''
+A suspected incident of credential leak for a Kafka instance with disabled
```
wtrocki, line 53: This looks different than what doc had :)
Should I correct this one?

rkpattnaik780: The reviewed para looked too long, tried to combine  best bits from that and the other para without losing critical info.

wtrocki: It lost meaning and form IMHO.
We need to inform user about side effect - 
this one has number of issues etc.

pkg/cmd/kafka/update/update.go
wtrocki: Lets remove that

pkg/cmd/kafka/update/update.go
```diff
@@ -203,6 +248,7 @@ func run(opts *options) error {
 	s.Stop()
 
 	if err != nil {
+		opts.logger.Info(opts.localizer.MustLocalize("kafka.update.log.info.updateFailed"))
```
wtrocki, line 251: Redundant. We usually want to return error with that message.

pkg/cmd/kafka/update/update.go
```diff
@@ -203,6 +248,7 @@ func run(opts *options) error {
...
 		if apiError := kas.GetAPIError(err); apiError != nil {
 			return errors.New(apiError.GetReason())
```
wtrocki, line 253: Use message here ^^

pkg/localize/locales/en/cmd/kafka_update.en.toml
```diff
@@ -20,6 +20,9 @@ $ rhoas kafka update --name=my-kafka --owner=other-user
 # update the owner of the current Kafka instance
 $ rhoas kafka update --owner=other-user
 
+# update the reauthentication configuration of the current Kafka instance
+$ rhoas kafka update --reauthentication true
```
wtrocki, line 24: mixing `=` and ` ` in arguments

pkg/localize/locales/en/cmd/kafka_update.en.toml
```diff
@@ -36,30 +39,49 @@ one = 'Name of the Kafka instance you want to update'
...
 one = 'Summary of Proposed Changes'
 
+[kafka.update.reauthentication.disclaimer]
+one = '''
+If a bad actor obtains credentials or service account to Kafka instance that has disabled reauthentication,
+deactivating the user account or service account will not disconnect already established connections.
+
+Options to disconnect client when reauthentication is disabled:
+- Contact Red Hat support
+- Use ACLs to prevent unauthorized connections from performing any operation
```
wtrocki, line 58:
```suggestion
- Use ACLs to prevent unauthorized connections from performing any operation
- Create support case for RHOSAK product
```

pkg/cmd/kafka/update/update.go
```diff
@@ -153,15 +169,40 @@ func run(opts *options) error {
...
+		err = survey.AskOne(reauthenticationPrompt, &opts.reauth)
+		if err != nil {
+			return err
+		}
 	}
 
 	updateObj := kafkamgmtclient.NewKafkaUpdateRequest()
-	updateObj.SetOwner(opts.owner)
+
+	if opts.owner != "" && opts.owner != kafkaInstance.GetOwner() {
```
wtrocki, line 187:
```suggestion
	if opts.owner != ""  {
	  if opts.owner == kafkaInstance.GetOwner() {
	    // warn user
	  }
```

pkg/cmd/kafka/update/update.go
```diff
@@ -153,15 +169,40 @@ func run(opts *options) error {
...
+		updateObj.SetOwner(opts.owner)
+		needsUpdate = true
+	}
+
+	if opts.reauth != "" && opts.reauth != strconv.FormatBool(kafkaInstance.GetReauthenticationEnabled()) {
```
wtrocki, line 192: Same as above.

pkg/cmd/kafka/update/update.go
```diff
@@ -49,6 +50,10 @@ type options struct {
 	Context    context.Context
 }
 
+var (
+	validAuthOptions = []string{"true", "false"}
```
wtrocki, line 54: Why?

pkg/localize/locales/en/cmd/kafka_update.en.toml
```diff
@@ -36,30 +39,49 @@ one = 'Name of the Kafka instance you want to update'
...
+one = 'Kafka instance could not be updated.'
+
 [kafka.update.log.info.loadingUsers]
 one = 'Loading users...'
 
 [kafka.update.input.message.selectOwner]
 one = "Select new owner:"
 
+[kafka.update.input.message.reauthentication]
+one = "Enable connection reauthentication [optional]:"
```
wtrocki, line 83:
```suggestion
one = "Enable connection reauthentication"
```

wtrocki: As every flag is optional unless validated

pkg/cmd/kafka/update/update.go
```diff
@@ -116,15 +125,19 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 
 	flags.StringVar(&opts.id, "id", "", opts.localizer.MustLocalize("kafka.update.flag.id"))
 	flags.StringVar(&opts.owner, "owner", "", opts.localizer.MustLocalize("kafka.update.flag.owner"))
+	flags.StringVar(&opts.reauth, "reauthentication", "", opts.localizer.MustLocalize("kafka.update.flag.reauthentication"))
```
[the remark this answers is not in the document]
rkpattnaik780: Using boolvar would mean it should be initialised to a default bool value. If the user runs command to change the owner, there would be no way for CLI to know if the `--reauthentication` was intended to change.

pkg/localize/locales/en/cmd/kafka_update.en.toml
rkpattnaik780: added the [optional] tag in line with `kafka topic update` interactive. Basically to imply that it could be left blank

pkg/cmd/kafka/update/update.go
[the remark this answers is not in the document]
wtrocki: Yes. I hit that problem many times. We need tribool in CLI :D

pkg/localize/locales/en/cmd/kafka_update.en.toml
wtrocki: Ah.. this is for interactive mode. Sorry I missed that mode. Please ignore.

pkg/cmd/kafka/update/update.go
rkpattnaik780: We need it as the run method fails with `Function 'run' has too many statements (58 > 50)`

pkg/cmd/kafka/update/update.go
rkpattnaik780: Not very much for of it as the existing update flows don't check for same values.

pkg/cmd/kafka/update/update.go
wtrocki: Can we make function have less statements :)

pkg/cmd/kafka/update/update.go
wtrocki: Not sure what you mean?

wtrocki:
Is that still draft?

pkg/cmd/kafka/update/update.go
```diff
@@ -49,6 +49,10 @@ type options struct {
 	Context    context.Context
 }
 
+var (
+	validAuthOptions = []string{"true", "false"}
```
wtrocki, line 53: I would prefer if we declare tribool type in utils with those values rater than define something like this in the command

pkg/cmd/kafka/update/update.go
wtrocki: Not addressed

rkpattnaik780: The method has 120+ lines of code, same goes for multiple places where we have excluded lints.

wtrocki: Not sure what you mean. Do you mean that we do not want to address that line? or suggestion is to increase limit?

pkg/localize/locales/en/cmd/kafka_update.en.toml
```diff
@@ -36,30 +39,49 @@ one = 'Name of the Kafka instance you want to update'
...
+
 [kafka.update.flag.yes]
 one = 'Forcibly update the Kafka instance without confirmation'
 
 [kafka.update.summaryTitle]
 one = 'Summary of Proposed Changes'
 
+[kafka.update.reauthentication.disclaimer]
+one = '''
+If a bad actor obtains credentials or service account to Kafka instance that has disabled reauthentication,
```
wtrocki, line 53: This change can affect your Kafka security
...If a bad actor

pkg/localize/locales/en/cmd/kafka_update.en.toml
```diff
@@ -36,30 +39,49 @@ one = 'Name of the Kafka instance you want to update'
...
+deactivating the user account or service account will not disconnect already established connections.
+
+Options to disconnect client when reauthentication is disabled:
+- Use ACLs to prevent unauthorized connections from performing any operation
```
wtrocki, line 57: When `-` is used both markdown and ascidocs rendering doesn't provide values for that properly. 
We should use `*` across entire cli for lists. 

Note: let's create seprate issue

wtrocki:
Verified, Reviewed

pkg/cmd/kafka/update/update.go
rkpattnaik780: I think this should be addressed in a separate issue, funlen check is skipped at several places and the the lines in this method is more than twice the limit.

pkg/cmdutil/flagutil/tribool.go
```diff
@@ -0,0 +1,60 @@
...
+	TRIBOOL_DEFAULT,
+}
+
+func isTriboolValid(val string) error {
+	switch Tribool(val) {
+	case TRIBOOL_FALSE, TRIBOOL_TRUE:
+		return nil
+	}
+
+	return errors.New("invalid tribool")
```
wtrocki, line 27: Missing i18n

pkg/cmd/kafka/update/update.go
wtrocki: I fully understyand what you are saying but that is new place introduced in that specific PR. Using that logic every of review comment can be addressed as separate github issue. I'm really giving up on this.

pkg/cmdutil/flagutil/tribool.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+type Tribool string
```
wtrocki, line 30: Missing docs. 

Inspiration: https://github.com/grignaak/tribool/blob/master/tribool.go

pkg/cmdutil/flagutil/tribool.go
```diff
@@ -0,0 +1,60 @@
...
+	TRIBOOL_DEFAULT = ""
+)
+
+var validTribools = []string{
+	TRIBOOL_TRUE,
+	TRIBOOL_FALSE,
+	TRIBOOL_DEFAULT,
+}
+
+func isTriboolValid(val string) error {
```
wtrocki, line 21: I do thing this should be public method

wtrocki (approved):

wtrocki:
The most polished PR I have seen for CLI.
Absolute masterpiece! 

Good to merge. Reviewed and verified!

wtrocki (approved):

pkg/cmd/kafka/update/update.go
```diff
@@ -305,6 +320,55 @@ func generateUpdateSummary(new reflect.Value, current reflect.Value) string {
...
+	} else {
+		kafkaInstance, _, err = kafka.GetKafkaByID(opts.Context, api, opts.id)
+		if err != nil {
+			return nil, err
+		}
+		opts.name = kafkaInstance.GetName()
+	}
+
+	return kafkaInstance, nil
+}
```
craicoverflow, line 340: This would be a great helper as there a few commands which repeat that logic.

PR #1344, opened 2021-12-08
feat(service-registry): add description flag for create command

rkpattnaik780:
Add `--description` flag to `service-registry create` command.

### Verification Steps
1. Try creating a service registry instance with description flag.
```
rhoas service-registry create --name test-instance --description "random description"
```
2. Create a registry instance through interactive mode:
```
rhoas service-registry create
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/registry/create/create.go
+++ b/pkg/cmd/registry/create/create.go
@@ -114,8 +114,11 @@ func runCreate(opts *options) error {
 		}
 	} else {
 		payload = &srsmgmtv1.RegistryCreate{
-			Name:        &opts.name,
-			Description: &opts.description,
+			Name: &opts.name,
+		}
+
+		if opts.description != "" {
+			payload.SetDescription(opts.description)
 		}
 	}
 
@@ -209,8 +212,11 @@ func promptPayload(opts *options) (payload *srsmgmtv1.RegistryCreate, err error)
 	}
 
 	payload = &srsmgmtv1.RegistryCreate{
-		Name:        &answers.Name,
-		Description: &answers.Description,
+		Name: &answers.Name,
+	}
+
+	if answers.Description != "" {
+		payload.SetDescription(answers.Description)
 	}
 
 	return payload, nil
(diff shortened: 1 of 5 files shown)


pkg/localize/locales/en/cmd/registry_crud.en.toml
```diff
@@ -130,6 +137,12 @@ one = 'Name of Service Registry instance:'
 [registry.cmd.create.input.name.help]
 one = 'Name can be any alphanumeric characters'
 
+[registry.cmd.create.input.description.message]
+one = 'Description of Service Registry instance [optional]:'
```
craicoverflow, line 141:
```suggestion
one = 'Description [optional]:'
```

Simply "Description" should be enough, like we do with Kafka. It should be clear from the context already.

pkg/cmd/registry/create/create.go
```diff
@@ -194,9 +201,23 @@ func promptPayload(opts *options) (payload *srsmgmtv1.RegistryCreate, err error)
 		return nil, err
 	}
 
+	promptDescription := &survey.Input{
+		Message: opts.localizer.MustLocalize("registry.cmd.create.input.description.message"),
+		Help:    opts.localizer.MustLocalize("registry.cmd.create.input.description.help"),
+	}
+
+	err = survey.AskOne(promptDescription, &answers.Description)
```
craicoverflow, line 209: It is likely that a description is long, possibly multiline. We need to ensure that the user can create a multi-line description.

craicoverflow (changes requested):
A question about validation left inline. One thing I think that we _need_ is multiline descriptions. Verified and works great otherwise.

wtrocki:
Also when doing edit we should always have description available as empty string:

rhoas service-registry artifact metadata-set --artifact-id=test

Does

```
{
 "labels": [],
 "properties": {}
}
```

It should

```
{
 "description": "",
 "labels": [],
 "properties": {}
}
```

wtrocki (approved):

craicoverflow:
I see this was approved, so I guess it should be merged?

craicoverflow (approved):

wtrocki:
I tend to leave PRs that were previously rejected by one team member for them to take look if they are happy with resolution. I'm certainly are so merging :D

PR #1345, opened 2021-12-08
feat(service-registry artifact): allow loading artifact from remote URL

rkpattnaik780:
It should be possible to create/update artifacts using remote URLs.

### Verification Steps
1. Pass url as value for `--file` while creating an artifact.
```
rhoas service-registry artifact create --type OPENAPI --file "https://raw.githubusercontent.com/bolcom/avro-schema-viewer/master/docs/assets/avsc/1.0/schema.avsc" --artifact-id foo_id
```
2. Pass url as value for `--file` to update an artifact.
```
rhoas service-registry artifact update --artifact-id foo_id --group DEFAULT --file "https://raw.githubusercontent.com/bolcom/avro-schema-viewer/master/docs/assets/avsc/1.0/schema.avsc"
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/registry/artifact/util/files.go
+++ b/pkg/cmd/registry/artifact/util/files.go
@@ -1,9 +1,13 @@
 package util
 
 import (
+	"context"
+	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
+	"strings"
 )
 
 func CreateFileFromStdin() (*os.File, error) {
@@ -25,4 +29,52 @@ func CreateFileFromStdin() (*os.File, error) {
 		return nil, err
 	}
 	return specifiedFile, nil
+}
+
+// IsURL accepts a string and determines if it is a URL
+func IsURL(s string) bool {
+	return strings.HasPrefix(s, "http:/") || strings.HasPrefix(s, "https:/")
+}
+
+// GetContentFromFileURL loads file content from the provided URL
+func GetContentFromFileURL(ctx context.Context, url string) (*os.File, error) {
+
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	client := http.DefaultClient
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("error loading file: %s", http.StatusText(resp.StatusCode))
+	}
+
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	tmpfile, err := ioutil.TempFile("", "rhoas-std-input")
+	if err != nil {
+		return nil, fmt.Errorf("error initializing temporary file: %w", err)
+	}
+
+	_, err = (*tmpfile).Write(data)
+	if err != nil {
+		return nil, err
+	}
+	_, err = (*tmpfile).Seek(0, io.SeekStart)
+	if err != nil {
+		return nil, err
+	}
+
+	return tmpfile, nil
 }
--- a/pkg/cmdutil/cmdutil.go
+++ b/pkg/cmdutil/cmdutil.go
@@ -1,14 +1,8 @@
 package cmdutil
 
 import (
-	"context"
 	"fmt"
-	"io"
-	"io/ioutil"
-	"net/http"
-	"os"
 	"strconv"
-	"strings"
 )
 
 func ConvertPageValueToInt32(s string) int32 {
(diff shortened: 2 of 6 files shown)


craicoverflow:
I don't see any reviewer for this or #1344 s not sure whether you would like them to be reviewed or not?

craicoverflow (commented):
```shell
❯ ./rhoas service-registry artifact update --artifact-id foo_id --group DEFAULT --file "https://raw.githubusercontent.com/bolcom/avro-schema-viewer/master/docs/assets/avsc/1.0/schema.avsc"
Loading file from url: https://raw.githubusercontent.com/bolcom/avro-schema-viewer/master/docs/assets/avsc/1.0/schema.avsc
❌ 404 Not Found. Run the command in verbose mode using the -v flag to see more information
```

This makes it seem like the file is not found, but in actual fact it is because the artifact does not exist. The user will not know that.

Overall this works really well 👍🏻  Not approving as I think there improvements that can be made but not outright rejecting it either. If you feel you want things kept as s that is fine, just re-request my review and let me know :)

pkg/cmdutil/cmdutil.go
```diff
@@ -38,3 +44,54 @@ func StringSliceToListStringWithQuotes(validOptions []string) string {
...
 	return listF
 }
+
+// IsURL accepts a string and determines if it is a URL
+func IsURL(s string) bool {
+	return strings.HasPrefix(s, "http:/") || strings.HasPrefix(s, "https:/")
+}
+
+// GetContentFromFileURL loads file content from the provided URL
+func GetContentFromFileURL(url string, ctx context.Context) (*os.File, error) {
```
wtrocki, line 54: FYI We already have some helper that has similar content: CreateFileFromStdin.

pkg/cmdutil/cmdutil.go
```diff
@@ -38,3 +44,54 @@ func StringSliceToListStringWithQuotes(validOptions []string) string {
...
+	defer resp.Body.Close()
+
+	tmpfile, err := ioutil.TempFile("", "rhoas_file-*")
+	if err != nil {
+		return nil, fmt.Errorf("error initializing temporary file: %w", err)
+	}
+
+	defer func() {
+		_ = tmpfile.Close()
+		_ = os.Remove(tmpfile.Name())
```
wtrocki, line 83: Not sure if we should be actually delete that file

pkg/cmdutil/cmdutil.go
wtrocki: https://github.com/redhat-developer/app-services-cli/blob/analitics/pkg/cmd/registry/artifact/util/files.go#L12:L13

pkg/cmdutil/cmdutil.go
```diff
@@ -38,3 +44,54 @@ func StringSliceToListStringWithQuotes(validOptions []string) string {
...
+	}()
+
+	_, err = io.Copy(tmpfile, respbody)
+	if err != nil {
+		return nil, err
+	}
+
+	specifiedFile, err := os.Open(tmpfile.Name())
```
wtrocki, line 91: Logic here works but it is opening the same file twice.
It is better to 1) write content into that file. Set buffer to beginning 
```
 _, err = (*specifiedFile).Write(data)
	if err != nil {
		return nil, err
	}
	_, err = (*specifiedFile).Seek(0, io.SeekStart)
	if err != nil {
		return nil, err
	}
```

wtrocki:
It works

pkg/cmdutil/cmdutil.go
```diff
@@ -38,3 +44,54 @@ func StringSliceToListStringWithQuotes(validOptions []string) string {
...
+		return nil, fmt.Errorf("error loading file: %s", http.StatusText(resp.StatusCode))
+	}
+
+	respbody := resp.Body
+
+	defer resp.Body.Close()
+
+	tmpfile, err := ioutil.TempFile("", "rhoas_file-*")
+	if err != nil {
+		return nil, fmt.Errorf("error initializing temporary file: %w", err)
```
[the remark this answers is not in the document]
rkpattnaik780: Should we use a struct for all the utility methods to pass common things like localizer, context, it seems bit odd to have an additional argument for everything?

wtrocki (approved):

craicoverflow:
This should be merged right?

rkpattnaik780:
> This should be merged right?

Will merge once approved.

craicoverflow:
> > This should be merged right?
> 
> Will merge once approved.

I see that @wtrocki has approved. Do PRs need 2 approvals?

pkg/cmd/registry/artifact/util/files.go
```diff
@@ -1,9 +1,13 @@
 package util
```
craicoverflow, line 1: I know it pre-exists this PR, but going forward we should not call packages "util", and we should update it when we see them.

pkg/cmd/registry/artifact/util/files.go
```diff
@@ -26,3 +30,51 @@ func CreateFileFromStdin() (*os.File, error) {
...
+
+	client := http.DefaultClient
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("error loading file: %s", http.StatusText(resp.StatusCode))
```
craicoverflow, line 55: i18n?

pkg/cmd/registry/artifact/util/files.go
```diff
@@ -26,3 +30,51 @@ func CreateFileFromStdin() (*os.File, error) {
...
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	tmpfile, err := ioutil.TempFile("", "rhoas-std-input")
+	if err != nil {
+		return nil, fmt.Errorf("error initializing temporary file: %w", err)
```
craicoverflow, line 67: i18n?

craicoverflow (approved):

pkg/cmdutil/cmdutil.go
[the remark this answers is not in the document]
craicoverflow: I agree, it also means we must continuously "break" APIs when new requirements are added such as i18n, which may not exist in a util but then if we need to add it we automatically break it..

A better thing to do might be to return a special error type like `InitializeTemporaryFileErr` and in the usage, interpret that and use i18n to create the message outside of the util..

Neither option seems ideal to me. Maybe merge it as it as its such a minor error, there's no point sacrificing usability.

PR #1350, opened 2021-12-15
refactor: refresh oauth tokens only when needed

famartinrh:
### Verification Steps

1. `rhoas login`
2. then run whatever other command several times with `-v` flag, you will see SSO token is only refreshed when it expires. MAS SSO token seems to be refreshed very frequently, I guess that's because it expires very quickly.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/connection/keycloak_connection.go
+++ b/pkg/connection/keycloak_connection.go
@@ -71,8 +71,6 @@ type KeycloakConnection struct {
 // RefreshTokens will fetch a refreshed copy of the access token and refresh token from the authentication server
 // The new tokens will have an increased expiry time and are persisted in the config and connection
 func (c *KeycloakConnection) RefreshTokens(ctx context.Context) (err error) {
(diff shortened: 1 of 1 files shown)


wtrocki:
Thank you for pointing that out. This change might improve snappyness of the CLI in long run.

> MAS SSO token seems to be refreshed very frequently, I guess that's because it expires very quickly.

This token would be removed in Jan/Feb so we do not care about efficiency here.

craicoverflow (approved):

craicoverflow:
Thanks for the PR.

> MAS SSO token seems to be refreshed very frequently, I guess that's because it expires very quickly.

Yeah it's weird why it is still being refreshed frequently, it looks as though `NeedsRefresh()` always returns true for that token only. It is possibly because as you said, the lifespan is shorter than the current acceptable limit (5 minutes) before we should refresh.

PR #1368, opened 2022-01-06
chore: fix CODEOWNERS not working

craicoverflow:
The CODEOWNERS file needs to pair a glob with an owner. It was missing the glob.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
-@redhat-developer/app-services-cli-maintainers
+# All files are maintained by @redhat-developer/app-services-cli-maintainers
+*	@redhat-developer/app-services-cli-maintainers


.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
-@redhat-developer/app-services-cli-maintainers
\ No newline at end of file
+# All files are maintained by @redhat-developer/app-services-cli-maintainers
+*	@redhat-developer/app-services-cli-maintainers
```
craicoverflow, line 2: @wtrocki I noticed that the **app-services-cli-maintainers** team is gone. It was probably removed during the clean up of teams. I think it makes sense if we have it back.

wtrocki: Ah.. so that was the purpose of that team. I did ask. We have new devexp team but will not be suitable here. 

I'm not sure who will be on maintainers list right now

craicoverflow: > Ah.. so that was the purpose of that team. I did ask.

Yeah I remember that you did 😄 But I forgot what its purpose was tbh.

> I'm not sure who will be on maintainers list right now

- @rkpattnaik780 
- @wtrocki 
- @craicoverflow

wtrocki: I do not see much value in being added as reviewer automatically. I do monitor every event in the repository and assign myself for reviews manually etc.

rkpattnaik780 (approved):

PR #1371, opened 2022-01-06
WIP: refactor: place utility methods in pkg/cmd

rkpattnaik780:
Moved utility methods to respective commands.

The sub-directories of `kafkautil` could be moved to the respective command specific directory.
Certain methods of `kafkautil`, `serviceregistryutil` and `serviceaccountutil` had few methods used by core, specific methods have been moved to respective directory.

### Directory Structure

```
|─ pkg
|── cmd
|   ├── kafka
|           |─── create
|           |─── update
|           |─── topic
|                   |─── create
|                   |─── update
|                   |─── ...
|                   |─── sdk
|           |─── ......
|           |─── sdk
|   ├── cluster
|           |─── sdk
|           |─── ...
|   ├── sdk
|──  kafkautil
|── serviceregistryutil
|── serviceaccountutil

```

### Type of change

- [ ] Documentation change
- [X] Refactor

### Checklist

- [ ] Code Review completed


pkg/cmd/kafka/acl/admin/admin.go
```diff
@@ -83,7 +83,7 @@ func NewAdminACLCommand(f *factory.Factory) *cobra.Command {
 			}
 
 			// user and service account should not allow wildcard
-			if userID == aclutil.Wildcard || serviceAccount == aclutil.Wildcard || userID == aclutil.AllAlias || serviceAccount == aclutil.AllAlias {
+			if userID == sdk.Wildcard || serviceAccount == sdk.Wildcard || userID == sdk.AllAlias || serviceAccount == sdk.AllAlias {
```
craicoverflow, line 86: If you look at this line, it is very difficult to understand what `sdk` relates to. `aclutil` is much more self descriptive.

`<thing>util` is a common pattern in Go: https://pkg.go.dev/io/ioutil

craicoverflow (changes requested):
Calling all of these packages a generic `sdk` makes things less readable.

pkg/cmd/cluster/bind/bind.go
```diff
@@ -3,10 +3,10 @@ package bind
...
 	"context"
 
-	"github.com/redhat-developer/app-services-cli/pkg/cluster"
-	"github.com/redhat-developer/app-services-cli/pkg/cluster/kubeclient"
-	"github.com/redhat-developer/app-services-cli/pkg/cluster/services/resources"
-	"github.com/redhat-developer/app-services-cli/pkg/cluster/v1alpha"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/cluster/sdk"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/cluster/sdk/kubeclient"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/cluster/sdk/services/resources"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/cluster/sdk/v1alpha"
```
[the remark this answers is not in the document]
wtrocki: https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/#:~:text=Why%20utils%20is%20a%20terrible,as%20well%20contain%20almost%20anything. 

:)

wtrocki:
Reviewed it later than Enda. Fully agree with the naming scheme that Enda proposed. 
I'm not 100% sure what to do with cluster logic right now until we have `shared service package` where we put code that needs to be aware of the all services etc. That is why I think cluster might stay where it is for now? WDYT?

pkg/cmd/cluster/bind/bind.go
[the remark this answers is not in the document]
craicoverflow: Yeah I agree. sdk and util is the same thing to me, different names.. `cluster` should be the package name here.

rkpattnaik780:
> That is why I think cluster might stay where it is for now? WDYT?

I will agree, reverting the commit.

pkg/cmd/kafka/acl/admin/admin.go
rkpattnaik780: I have followed `<thing>util` for utility methods at top level, i.e `pkg/kafkautil`, `pkg/serviceregistryutil`. I was hoping to follow a different naming convention for utils specific to commands, maybe `aclcmdutil`, `topiccmdutil` could be better alternative? Wdyt?

wtrocki: I think it is all about import path.

if util/sdk/common has subpackages then we have nice pattern - single name used across all cli packages with specific sub packages

wtrocki: Quick look at other projects confirms what enda said. Havents seen packages with common subpackages being used in go

wtrocki: Although our case is different as people might want import sdk etc. so it is self documenting change. etc

craicoverflow: Yeah, the import path does not matter a bit, what is important is the name of the package which will be used in the code.

craicoverflow: > I have followed `<thing>util` for utility methods at top level, i.e `pkg/kafkautil`, `pkg/serviceregistryutil`. I was hoping to follow a different naming convention for utils specific to commands, maybe `aclcmdutil`, `topiccmdutil` could be better alternative? Wdyt?

Yes I like that!.

wtrocki:
Amazing work! 
I think we could start documenting our approach in the docs explaining approaches etc.

pkg/cmd/kafka/acl/delete/delete.go
```diff
@@ -162,7 +162,7 @@ func runDelete(instanceID string, opts *aclutil.CrudOptions) error {
 		defer httpRes.Body.Close()
 	}
 
-	err = aclutil.ValidateAPIError(httpRes, opts.Localizer, err, "delete", kafkaInstance.GetName())
+	err = aclcmdutil.ValidateAPIError(httpRes, opts.Localizer, err, "delete", kafkaInstance.GetName())
```
craicoverflow, line 165: I think that `ValidateAPIError` belongs more in an SDK instead of a cmd util, but we can revisit that when the error handling refactor happens.

pkg/cmd/kafka/consumergroup/describe/describe.go
```diff
@@ -189,9 +189,9 @@ func printConsumerGroupDetails(w io.Writer, consumerGroupData kafkainstanceclien
 	fmt.Fprintln(w, "")
 	consumers := consumerGroupData.GetConsumers()
 
-	activeMembersCount := consumergrouputil.GetActiveConsumersCount(consumers)
-	partitionsWithLagCount := consumergrouputil.GetPartitionsWithLag(consumers)
-	unassignedPartitions := consumergrouputil.GetUnassignedPartitions(consumers)
+	activeMembersCount := groupcmdutil.GetActiveConsumersCount(consumers)
```
craicoverflow, line 192: nice and concise now :)

pkg/cmd/kafka/topic/create/create.go
```diff
@@ -203,7 +202,7 @@ func runCmd(opts *options) error {
 }
 
 func runInteractivePrompt(opts *options) (err error) {
-	validator := topicutil.Validator{
+	validator := topiccmdutil.Validator{
```
craicoverflow, line 205: I'd say eventually, validators wont be in the util package as they are not util methods, again that can happen during the SDK creation.

pkg/cmd/kafka/topic/topiccmdutil/validators.go
```diff
@@ -1,4 +1,4 @@
-package topicutil
+package topiccmdutil
```
craicoverflow, line 1: We should probably start to document all pakcage names and their meaning too, this should happen at the top of the `package topiccmdutil` declaration, but to prevent duplication can be added in a doc.go file: https://go.dev/blog/godoc

craicoverflow (approved):
LGTM!

craicoverflow:
> Amazing work! I think we could start documenting our approach in the docs explaining approaches etc.
> 
> We can do that after PR is merged

https://go.dev/blog/godoc

We could document packages using the available Go tooling. You can define it all in a `doc.go` file: https://github.com/googleapis/google-cloud-go/blob/main/datastore/doc.go

wtrocki:
@craicoverflow @rkpattnaik780 Since we talking - what is your preference:

A) Package based docs like: https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cluster/cluster.go or 
 https://github.com/googleapis/google-cloud-go/blob/main/datastore/doc.go
 
 B) Markdown format:
 https://github.com/kubernetes/client-go/blob/master/INSTALL.md

wtrocki:
Actually realized that syntax for the ones we have is wrong: https://pkg.go.dev/github.com/redhat-developer/app-services-cli@v0.37.0/pkg/cluster and as enda mentioned we should probably use docs.go instead.

PR #1380, opened 2022-01-10
Remove AMS from codebase

wtrocki:


wtrocki:
@craicoverflow @rkpattnaik780 Side effect from this change is that we no longer generate any openapi in CLI.

I feel like api/rbac/rbacutil might be actually embedded into specific set of commands as implementation detail. If we feel that this RBAC API will be used anywhere else we might move that somewhere else

WDYT?

craicoverflow:
I will review this tomorrow. FYI you do not need to duplicate assigning reviewers for people in @redhat-developer/app-services-cli-maintainers

wtrocki:
I would like to reserve right to not use app-services-cli-maintainers and will ping exact person to review PRs.

craicoverflow:
Okay. I was just saying in case you did not know.

wtrocki:
I follow every interaction in repository and see every comment etc, so generally found that redundant - also it is pinging everyone which is not clear who you would like to look.

wtrocki:
Actually there is not much to review. All we need is verification which I have done by creating Kafka.

rkpattnaik780 (approved):

craicoverflow (approved):

PR #1382, opened 2022-01-11
fix: remove logging from segment

wtrocki:
Remove any logging from analytics so it is not issuing/obfuscating user response. 

Telemetry delivery can fail with bad network etc. so we should not print that. It is not critical for user to see it as well we cannot do much with error anyway.

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -116,7 +116,7 @@ func (t *Telemetry) Finish(event string, cmdError error) {
 	}
 	defer telemetryClient.Close()
 
-	telemetryClient.Upload(t.telemetryData)
+	_ = telemetryClient.Upload(t.telemetryData)
 	telemetryClient.Close()
 
 }
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -1,11 +1,12 @@
 package telemetry
 
 import (
-	"github.com/redhat-developer/app-services-cli/pkg/core/cmdutil/factory"
 	"os"
 	"runtime"
 	"strconv"
 	"time"
+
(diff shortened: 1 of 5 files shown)


internal/telemetry/telemetry.go
```diff
@@ -115,11 +116,7 @@ func (t *Telemetry) Finish(event string, cmdError error) {
 	}
 	defer telemetryClient.Close()
 
-	err = telemetryClient.Upload(t.telemetryData)
-	if err != nil {
-		t.factory.Logger.Info("Cannot send data to telemetry: %q", err)
-	}
-
+	_ = telemetryClient.Upload(t.telemetryData)
```
craicoverflow, line 119: Should we still log the error in verbose mode?

pkg/core/ioutil/editor/linux.go
```diff
@@ -1,4 +1,5 @@
 // nolint
+//go:build !windows
```
craicoverflow, line 2: Does that replace `// +build !windows`?

craicoverflow (approved):

pkg/core/ioutil/editor/linux.go
wtrocki: That was added by golang. Maybe I have different version. Should I resolve that?

internal/telemetry/telemetry.go
wtrocki: Good question. I will add that to debug

pkg/core/ioutil/editor/linux.go
craicoverflow: I have no idea, it could be how to do it with new versions or something.

PR #1383, opened 2022-01-12
fix: add cli prefix to adoc files

wtrocki:


wtrocki:
CC @bhardesty I have resolved conflicts downstream caused by sync job. Wanted to apply latest changes but also noticed that we renamed files. I have restored original file names.

PR #1384, opened 2022-01-13
chore(website): add UX guidelines base structure

craicoverflow:
- chore: use root 'docs' folder as website source
- chore: add UX guidelines base structure
- chore: use custom website sidebar

## Verification

From the `website` folder run `yarn start` to run the website locally.


docs/commands/_category_.json
```diff
@@ -1,4 +0,0 @@
-{
```
wtrocki, line 1: [Question] Why - better to autogenerate that?

wtrocki (approved):

wtrocki:
Amazing work! Let's not publish that yet as we do not have content

docs/commands/_category_.json
craicoverflow: Sorry I don't fully understand your question.

wtrocki: Why file is removed vs we have that info in website. More of question so I will understand it.

craicoverflow: https://github.com/craicoverflow/app-services-cli/blob/ux-guidelines-base-structure/website/sidebars.js#L24-L30

I have added this to the sidebars file instead which does the same thing. I wanted to have the contributing section open by default and this one collapsed, and wanted more control over ordering.

wtrocki: Really cool! Thank you for explaining

PR #1388, opened 2022-01-19
fix: parse yq version for different outputs

craicoverflow:
YAML output was not being output in color due to a change in the output text of `yq --version`, from which we detect the YQ version.

### Verification

**Prerequisites**

- yq 4.x

Run `rhoas kafka describe -o yaml` and verify that it has been colored. 

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/core/ioutil/dump/dump.go
+++ b/pkg/core/ioutil/dump/dump.go
@@ -5,11 +5,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


wtrocki:
This is good change. In long run it will be good to minimize dependency on yq or any external tool.

craicoverflow:
> This is good change. In long run it will be good to minimize dependency on yq or any external tool.

Agreed 100%, the new guidelines will be adding that bit and so we can change it then when implementing the guidelines.

PR #1389, opened 2022-01-19
docs(ux-guidelines): add Output chapter with rules

craicoverflow:
This PR adds the Output section guidelines to for the UX guidelines website.

Some things to note:

* For simplicity I am using PNG images which is not optimal, we can use something better like SVG when we know we have settled on the correct ones.
* I opted to use the [RFC 2119 Convention](https://www.ietf.org/rfc/rfc2119.txt) for writing guidelines against something more informal like seen on [CLIG](https://clig.dev/#output) as our rules are more concrete and specific.


docs/contributing/ux-guidelines.md
```diff
@@ -3,8 +3,119 @@ id: ux-guidelines
...
+
+User's may need or want color output disabled. When the `NO_COLOR` enviroment variable is set then the additional of ANSI color to the output should be prevented.
+
+### **SHOULD NOT** display any animations in `stdout`
+
+Animations should only be displayed in `stderr`, as `stdout` needs to be kept for machine-readable output.
+
+### **MAY** use symbols and emoji where it makes things clearer
+
+Symbols and emojis can sometimes tell more than words can and they can also be useful to draw someone's attention to something.
```
wtrocki, line 72: [Question] Style of this guide would be more informal ?

```suggestion
Symbols and emojis can sometimes tell more than words can and they can also be useful to draw someone's attention to something.
```

If we want to be more formal then we can say it shorter: `Developers are allowed to use emojis`

docs/contributing/ux-guidelines.md
```diff
@@ -3,8 +3,119 @@ id: ux-guidelines
...
+
+Be careful and don't overdo it - too many symbols can dilute their significance and can make the application look like a toy.
+
+### **MUST** use color in all JSON and YAML output
+
+All JSON and YAML output **MUST** be syntax highlighted by default.
+
+**JSON:**
+
+![JSON Output example](../../website/static/img/ux-assets/json-output.png)
```
wtrocki, line 85: Image references are not canonical. In docusaurus they say to use basepath.

so urls will be `/assets/img/ux-assets/json-output.png`

wtrocki (approved):
Amazing work! No comment about actual content. Let's standarize image references before merge: https://docusaurus.io/docs/next/static-assets

docs/contributing/ux-guidelines.md
craicoverflow: Informal is okay. The formal bit is the way everything is split into RFC 2119 headings instead of just flowing into each other naturally.

docs/contributing/ux-guidelines.md
craicoverflow: As the docs are in the root of the repo outside of the website directory, doing so would break the raw files, if someone would like to view it that way. For example if you open the file directly: https://github.com/craicoverflow/app-services-cli/blob/ux-guidelines/docs/contributing/ux-guidelines.md you can still see the images rendered. It would also be the case when viewing the files while checked out locally in your IDE.

wtrocki: Good point. if that works in production then we good

PR #1390, opened 2022-01-20
docs: api standards file

wtrocki:


wtrocki:
https://github.com/redhat-developer/app-services-api-guidelines/pull/34 created duplicate in standard repository so we can discuss options on follow up call

PR #1391, opened 2022-01-20
feat: make the spinner cyan color

craicoverflow:
Changed the color of the spinner to cyan to make it stand out more.

Screenshot 2022-01-20 at 17 35 17

I actually wanted to do it in red to align with the product color scheme but it could be mistaken to mean an error.

### DIFF
--- a/pkg/core/ioutil/spinner/spinner.go
+++ b/pkg/core/ioutil/spinner/spinner.go
@@ -2,9 +2,10 @@ package spinner
 
 import (
 	"fmt"
-	"github.com/redhat-developer/app-services-cli/pkg/core/localize"
 	"io"
 	"time"
+
+	"github.com/redhat-developer/app-services-cli/pkg/core/localize"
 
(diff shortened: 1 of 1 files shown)


wtrocki:
Generally enforcing colour of the text printed in the terminal should be discouraged due to people having different collour profiles in terminals (with even photo backgrounds)

This case is for progress bar so not that important and colour will work with two main backgrounds so I guess we are ok.

wtrocki (approved):

craicoverflow:
> Generally enforcing colour of the text printed in the terminal should be discouraged due to people having different colour profiles in terminals (with even photo backgrounds)

This uses ansi colors (notice it is "cyan" not a specific color code) so the colour will adapt to the profile in use.

Screenshot 2022-01-21 at 08 44 40
Screenshot 2022-01-21 at 08 49 30
Screenshot 2022-01-21 at 08 50 46

PR #1401, opened 2022-01-27
SDK errors

wtrocki:
## Verification

```
rhoas service-registry create # reach limits 
rhoas service registry describe --name=whatever
rhoas service registry describe --id=whatever

```

## Notes (outdated)
Current problems I see:

1) Lack way to handle errors per SDK (API exposing hasError etc.) so we need to define errors in multiple places/defefine them or have helper methods for creating proper message per type.

2) Errors type should be probably moved to the core and not defined for kafka specifically

3) SDK on its own defines all errors on root package. Maybe we should have `kafkaerrors.ERROR_40` instead?
Screenshot 2022-01-27 at 12 57 00


pkg/kafkautil/error.go
```diff
@@ -1,83 +1,37 @@
...
-)
-
-type ServiceErrorCode string
-
+// TODO - this methods should not be deifned here - or even defined this way.
 type Error struct {
 	Err error
 }
 
+// TODO - this methods should not be deifned here - or even defined this way.
```
rkpattnaik780, line 12: They seem unused, we can remove them I guess

pkg/cmd/kafka/create/create.go
```diff
@@ -211,11 +212,11 @@ func runCreate(opts *options) error {
...
 	}
 
-	if apiErr := pkgKafka.GetAPIError(err); apiErr != nil {
+	if apiErr := kafkamgmt.GetAPIError(err); apiErr != nil {
 		switch apiErr.GetCode() {
-		case pkgKafka.ErrorCode24:
+		case string(kafkamgmt.ERROR_24):
 			return opts.localizer.MustLocalizeError("kafka.create.error.oneinstance")
-		case pkgKafka.ErrorCode36:
+		case string(kafkamgmt.ERROR_36):
```
wtrocki, line 219: What is the namespace we can use. Is `errors` ok?

rkpattnaik780: Sounds good :+1:

craicoverflow:
> Current problems I see:
> SDK on its own defines all errors on root package. Maybe we should have kafkaerrors.ERROR_40 instead?

I just have a question - can you explain further what the problem you see is?

wtrocki:
There are 80 fields in the root of the SDK (from which 95% are errors constants)
image

craicoverflow:
Thanks that is clear!

pkg/shared/serviceregistryutil/error.go
```diff
@@ -0,0 +1,15 @@
+package serviceregistryutil
+
+import (
+	"fmt"
+)
+
+func NotFoundByIDError(id string) error {
+	NotFoundByIDErr := fmt.Errorf(`Registry instance with ID "%v" not found`, id)
```
wtrocki, line 8: It wasn't possible to add localization to those (same as kafka) due to them being used in different context that do not have localiser injected. I think it is good to have those 2 localized and will add that after this PR (it will require a looot of changes that will obfuscate this PR)

pkg/shared/kafkautil/error.go
```diff
@@ -1,83 +1,21 @@
...
-	if mappedErr == nil {
-		return false
-	}
-
-	return mappedErr.GetCode() == string(code)
-}
-
 func NotFoundByIDError(id string) error {
-	NotFoundByIDErr = fmt.Errorf(`Kafka instance with ID "%v" not found`, id)
+	NotFoundByIDErr := fmt.Errorf(`Kafka instance with ID "%v" not found`, id)
```
craicoverflow, line 8: You could return the error inline?

pkg/shared/serviceregistryutil/error.go
craicoverflow: You could create a ServiceRegistryErrors struct to wrap these errors?

craicoverflow (approved):
LGTM

rkpattnaik780 (approved):

pkg/shared/kafkautil/error.go
wtrocki: Oh.. yes. That was copy paste from kafka errors.

wtrocki:
It feels like this was overkill to handle only 3 errors from srs (as all kafka errors were handled already). 
We should try to refactor error handling in the CLI but I see this as step process where we improve new and existing commands rather than doing this as whole.

PR #1404, opened 2022-01-31
refactor: split `pkg` to shared, cmd and core

rkpattnaik780:
`pkg` has been split into:
* `core` - elements independent of service
* `shared` - common code assuming existence of all services.
* `cmd` - Commands along with their helpers

### Current directory structure
|─ pkg
|── cmd
|   ├── kafka
|   ├── registry
|   ├── serviceaccount
|   ├── cluster
|   ├── .....
|── core
|   ├── cmdutil
|   ├── config
|   ├── errors
|   ├── logging
|   ├── localize
|   ├── ...
|── shared
|   ├── accountmgmtutil
|   ├── cluster
|   ├── connection
|   ├── factory
|   ├── kafkautil
|   ├── remote
|   ├── serviceregistryutil
|   ├── servicespec
|   ├── svcstatus
|   ├── ...

### Type of change

- [ ] Documentation change
- [X] Refactor

### Checklist

- [ ] Code Review completed


craicoverflow:
> shared - common code assuming existence of all services.

Would we not just assume that everything NOT in core will be shared?

rkpattnaik780:
> > shared - common code assuming existence of all services.
> 
> Would we not just assume that everything NOT in core will be shared?

That will be a better explanation, will change it.

rkpattnaik780:
@craicoverflow , completions will be better suited for `shared` imho, wdyt?

wtrocki:
Reviewed and verified locally! Great job!

wtrocki:
Actually. I still see some circular dependencies:

- pkg/core/config - has kafka and registry - I think we can ignore that for moment
- completions.go references shared folder kafka utils

wtrocki:
https://github.com/redhat-developer/app-services-cli/blob/refactor_2/pkg/core/cmdutil/flagutil/completions.go#L90:L103 

those two can be eather:
- moved to shared
- Move to internal in the cmd.

craicoverflow:
> @craicoverflow , flag completions will be better suited for `shared` imho, wdyt?

Which package is "flag completions"?

craicoverflow:
> Move to internal in the cmd (my prefference)

`flagutil` is actually a base package - flagutil packages for subcommands actually embed this a lot.

wtrocki:
I meant only this two util methods.

craicoverflow:
> I meant only this two util methods.

* `RegisterTopicCompletionFunc` - related to topics - should be in the flagutil for topics as it could be used in multiple commands (delete and update)
* `RegisterGroupCompletionFunc` - relates to consumer groups - should be in consumergroup flag util package.

rkpattnaik780:
> * `RegisterTopicCompletionFunc` - related to topics - should be in the flagutil for topics as it could be used in multiple commands (delete and update)
> * `RegisterGroupCompletionFunc` - relates to consumer groups - should be in consumergroup flag util package.

Since they are used by acls, I think we should keep them at higher level or aclcmdutil.

craicoverflow:
> Since they are used by acls, I think we should keep them at higher level.

If that is the case - agreed. though it should be at the higher level then like `kafkadataflagutil` or something.

> or aclcmdutil

Probably not, because we would want them in the topic and consumer-group commands?

rkpattnaik780:
> Probably not, because we would want them in the topic and consumer-group commands?

Since they are specific to `--topic` and `--group` flag, we can keep it in `aclcmdutil`

wtrocki:
LGTM

PR #1406, opened 2022-01-31
fix: change partition limit for new topic creation

wtrocki:
## Verification

```
#fail 
rhoas kafka topic create --partitions=3000
# pass
rhoas kafka topic create --partitions=900
```


rkpattnaik780:
Looks good. These [tests](https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/kafka/topic/topiccmdutil/validators_test.go#L247-L258) need to be updated.

wtrocki:
Done

rkpattnaik780 (approved):

PR #1409, opened 2022-02-01
fix: commit linting github workflow only on main

wtrocki:
Running job on PRs when we always squashing commits makes less sense and it changes status of the build (where sometimes we want to get notified about build failing). To minimize spam I think we should change job to run on main only

### DIFF
--- a/.github/workflows/commit_lint.yml
+++ b/.github/workflows/commit_lint.yml
@@ -1,6 +1,9 @@
 name: Validate
-on: [pull_request]
-
+on: 
+  push:
+    branches:
+      - main
+ 
 jobs:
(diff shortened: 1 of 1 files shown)


craicoverflow:
But then you merge a pull request with a commit message that does not conform to conventional commits, but it is too late at that stage since it was merged..

wtrocki:
> But then you merge a pull request with a commit message that does not conform to conventional commits, but it is too late at that stage since it was merged..

yes but that happens when squashing in github that is not covered by this job. so unless we:

- Change workflow to squash locally and do keep message as it is (too much hassle)
- Have some hook that reject commits etc.

PR #1413, opened 2022-02-01
fix!: align roles for service registry

wtrocki:
## Motivation

UI using some different values so we see this as confusing/mixed UX that should be aligned. 
New roles are introduced in breaking fashion as we can easily provide feedback to users about the change

## Verification

```
rhoas service-registry role add --role Viewer --username wtrocki_kafka_devexp
rhoas service-registry role list
```

### DIFF
--- a/docs/commands/rhoas_cluster_connect.md
+++ b/docs/commands/rhoas_cluster_connect.md
@@ -20,7 +20,7 @@ After running this command, you need to grant access for the service account tha
 
 For the Service Registry application service, enter this command:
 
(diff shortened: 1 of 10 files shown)


wtrocki:
@bredamc Link to content team review: 

https://github.com/redhat-developer/app-services-cli/pull/1413/files?file-filters%5B%5D=.toml&show-viewed-files=true

rkpattnaik780 (approved):

wtrocki:
Fun fact: More than 70% of the code in this PR were entirely generated by AI bot:
Screenshot 2022-02-01 at 23 14 26

PR #1414, opened 2022-02-02
fix!: use lowercase values for registry roles

craicoverflow:
We want the CLI to be a relatively quick way to execute operations against the service. By forcing users to toggle the case of a value, we are actually slowing them down.
As a rule of thumb we should use lowercase in almost all cases except for exceptional circumstances.

### DIFF
--- a/docs/commands/rhoas_cluster_connect.md
+++ b/docs/commands/rhoas_cluster_connect.md
@@ -20,7 +20,7 @@ After running this command, you need to grant access for the service account tha
 
 For the Service Registry application service, enter this command:
 
-  $ rhoas service-registry role add --role=Manager --service-account your-sa
+  $ rhoas service-registry role add --role=manager --service-account your-sa
 
 
 ```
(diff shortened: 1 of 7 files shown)


wtrocki:
Hi

Missunderstanding. Feedback we got  was to align those values with the UI and sadly UI already uses capital values :/

craicoverflow:
> Hi
> 
> Missunderstanding. Feedback we got was to align those values with the UI and sadly UI already uses capital values :/ I can merge that quickly but I will need to check with PM if that is something they ok with

I don't think it needs to be aligned down to this level - it is still the same value just in lowercase.

wtrocki (approved):

wtrocki:
I still see `Admin` as capital othewise it is ok

wtrocki:
https://github.com/redhat-developer/app-services-cli/blob/main/pkg/core/localize/locales/en/cmd/role.toml#L21:L21

wtrocki (approved):
Great work! Let's do path release after merge

PR #1432, opened 2022-02-14
chore: ignore generated markdown in pull requests

craicoverflow:


wtrocki (approved):

wtrocki:
It is hart to test it now without content so we need to rebase our feature branch

PR #1438, opened 2022-02-14
Interactive create

wtrocki:
## Reason

When creating kafka should:

- not validate interactive mode - it would be valid by default 
- list only supported types for creation

## Verification

```
wtrocki@graphapi app-services-cli (interactive-create)]$ rhoas kafka create            
? Name: test
? Cloud Provider: aws
? Cloud Region:  [Use arrows to move, type to filter, ? for more help]
> us-east-1
```


rkpattnaik780 (approved):
LGTM!

PR #1439, opened 2022-02-15
feature - service-registry rule

wtrocki:
Creating for global view on all changes for registry rule commands. 
THIS BRANCH SHOULD NOT BE SQUASHED. Original commits should be kept as it is . 

To fix issues please create PR against this branch


wtrocki:
FYI @bredamc . Looking for review

bredamc:
Apologies @wtrocki and @rkpattnaik780, I didn't get to review your PRs yet but will do so next week. I have other rules-related doc updates to make, so if you prefer, I can review your changes while making my updates in a separate doc PR?

Issue #1430, opened 2022-02-13
Unable to delete ACLs for existing service account


ppatierno:
I have a service account together with some ACLs related to it.

```shell
rhoas kafka acl list --service-account srvc-acct-adg23480-dsdf-244a-gt65-d4vd65784dsf

PRINCIPAL (9)                                    PERMISSION   OPERATION          DESCRIPTION              
------------------------------------------------ ------------ ------------------ ------------------------- 
All Accounts                                     allow        describe           cluster is "*"           
All Accounts                                     allow        describe           group is "*"             
All Accounts                                     allow        describe           topic is "*"             
All Accounts                                     allow        describe-configs   topic is "*"             
srvc-acct-adg23480-dsdf-244a-gt65-d4vd65784dsf   allow        create             topic starts with "f1-"  
srvc-acct-adg23480-dsdf-244a-gt65-d4vd65784dsf   allow        describe           topic starts with "f1-"  
srvc-acct-adg23480-dsdf-244a-gt65-d4vd65784dsf   allow        write              topic starts with "f1-"  
srvc-acct-adg23480-dsdf-244a-gt65-d4vd65784dsf   allow        describe           transactional-id is "*"  
srvc-acct-adg23480-dsdf-244a-gt65-d4vd65784dsf   allow        write              transactional-id is "*"
```

I would like to delete all the ACLs related to that specific service account, so I tried the following:

```shell
rhoas kafka acl delete --operation all --permission any --topic f1- --prefix --service-account srvc-acct-adg23480-dsdf-244a-gt65-d4vd65784dsf
```

But the result is the following:

```shell
? All ACLs matching the criteria provided will be deleted from the Kafka instance "formula1-kafka". Are you sure you want to proceed? Yes


ℹ  No ACLs match the specified criteria
```

ppatierno:
So it seems that the problem is with `--operation all` because if I run the same command by specifying the specific operation (i.e. `create`), the corresponding ACL is deleted.
I can understand that the command is trying to match exactly the ACL so it should have `all` as operation and in this case there is no match. I was just wondering if it could be understandable by the user.

wtrocki:
Will discuss that with @mikeedgar

craicoverflow:
`--operation all` does not mean _all_ operations - "all" is a specific operation type. You cannot currently delete all ACLs tied to an account.

wtrocki:
[...]

PR #1441, opened 2022-02-15
fix: allow to delete all ACLs for principal

wtrocki:
Creating for early feedback for https://github.com/redhat-developer/app-services-cli/issues/1430

Validation for delete command were redundant preventing users from deleting multiple acls. 
For example it would not be possible to remove all ACLs for Service account.

After this change it is possible:

```
✔️  Deleted 5 ACLs from Kafka instance "wtrocki"
The following ACL rules were deleted:

  PRINCIPAL (5)    PERMISSION   OPERATION   DESCRIPTION               
 ---------------- ------------ ----------- -------------------------- 
  test             allow        write       transactional-id is "*"   
  test             allow        describe    topic starts with "test"  
  test             allow        write       topic starts with "test"  
  test             allow        create      topic starts with "test"  
  test             allow        describe    transactional-id is "*"   
  ```

However we need to still spend some time cleaning up messages/validation and arguments to API.
Posted for early feedback

## Verification
```
[wtrocki@graphapi app-services-cli (delete-acl)]$ rhoas kafka acl delete --pattern-type=any --service-account=test
? All ACLs matching the criteria provided will be deleted from the Kafka instance "wtrocki". Are you sure you want to proceed? Yes


✔️  Deleted 5 ACLs from Kafka instance "wtrocki"
The following ACL rules were deleted:

  PRINCIPAL (5)    PERMISSION   OPERATION   DESCRIPTION               
 ---------------- ------------ ----------- -------------------------- 
  test             allow        describe    topic starts with "test"  
  test             allow        write       transactional-id is "*"   
  test             allow        describe    transactional-id is "*"   
  test             allow        write       topic starts with "test"  
  test             allow        create      topic starts with "test" 
  ```

Delete all write operations for sa
```
rhoas kafka acl delete --service-account=test1 --operation=write -v --pattern-type=any
```

other variations

### DIFF
--- a/pkg/cmd/kafka/acl/delete/delete.go
+++ b/pkg/cmd/kafka/acl/delete/delete.go
@@ -55,13 +55,11 @@ func NewDeleteCommand(f *factory.Factory) *cobra.Command {
 
 			var errorCollection []error
 
-			// if opts.Operation == "" {
-			// 	errorCollection = append(errorCollection, opts.Localizer.MustLocalizeError("kafka.acl.common.flag.operation.required"))
-			// }
-
-			// if resourceErrors := aclcmdutil.ValidateAndSetResources(opts, aclFlagUtil.ResourceTypeFlagEntries); resourceErrors != nil {
-			// 	errorCollection = append(errorCollection, resourceErrors)
-			// }
+			selectedResourceTypeCount := aclcmdutil.SetACLResources(opts)
+			if selectedResourceTypeCount > 1 {
+				errorCollection = append(errorCollection,
+					opts.Localizer.MustLocalizeError("kafka.acl.common.error.oneResourceTypeAllowed", aclFlagUtil.ResourceTypeFlagEntries...))
+			}
 
 			if principalErrors := validateAndSetOpts(opts); principalErrors != nil {
 				errorCollection = append(errorCollection, principalErrors)
@@ -148,15 +146,29 @@ func runDelete(instanceID string, opts *aclcmdutil.CrudOptions) error {
 
 	requestParams := getRequestParams(opts)
 
-	deletedACLs, httpRes, err := adminAPI.AclsApi.DeleteAcls(ctx).
-		//ResourceType(requestParams.resourceType).
-		Principal(requestParams.principal).
+	requestDeleteAcls := adminAPI.AclsApi.DeleteAcls(ctx)
+	if requestParams.resourceType != "" {
+		requestDeleteAcls = requestDeleteAcls.ResourceType(requestParams.resourceType)
+	}
 
-		// PatternType(requestParams.patternType).
-		// ResourceName(requestParams.resourceName).
-		// Operation(requestParams.operation).
-		// Permission(requestParams.permission).
-		Execute()
+	if requestParams.principal != "" {
+		requestDeleteAcls = requestDeleteAcls.Principal(requestParams.principal)
+	}
+
+	if requestParams.resourceName != "" {
+		requestDeleteAcls = requestDeleteAcls.ResourceName(requestParams.resourceName)
+	}
+	if requestParams.patternType != "" {
+		requestDeleteAcls = requestDeleteAcls.PatternType(requestParams.patternType)
+	}
+	if requestParams.operation != "" {
+		requestDeleteAcls = requestDeleteAcls.Operation(requestParams.operation)
+	}
+	if requestParams.permission != "" {
+		requestDeleteAcls = requestDeleteAcls.Permission(requestParams.permission)
+	}
+
+	deletedACLs, httpRes, err := requestDeleteAcls.Execute()
 
 	if httpRes != nil {
 		defer httpRes.Body.Close()
--- a/pkg/cmd/kafka/acl/delete/delete.go
+++ b/pkg/cmd/kafka/acl/delete/delete.go
@@ -91,13 +91,7 @@ func NewDeleteCommand(f *factory.Factory) *cobra.Command {
 	flags.AddServiceAccount(&serviceAccount)
 	flags.AddAllAccounts(&allAccounts)
 	flags.AddYes(&opts.SkipConfirm)
-
-	cmd.Flags().BoolVar(
-		&prefix,
-		"prefix",
-		false,
-		flagutil.DeprecateFlag(opts.Localizer.MustLocalize("kafka.acl.common.flag.delete.prefix.description")),
-	)
+	flags.AddPrefix(&prefix)
 
 	cmd.Flags().StringVar(
 		&patternTypeFlag,
--- a/pkg/cmd/kafka/acl/delete/delete.go
+++ b/pkg/cmd/kafka/acl/delete/delete.go
@@ -4,7 +4,6 @@ import (
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/kafka/acl/aclcmdutil"
 	aclFlagUtil "github.com/redhat-developer/app-services-cli/pkg/cmd/kafka/acl/flagutil"
-	"github.com/redhat-developer/app-services-cli/pkg/core/cmdutil"
 	"github.com/redhat-developer/app-services-cli/pkg/core/cmdutil/flagutil"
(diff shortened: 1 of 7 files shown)


rkpattnaik780:
Deleting all acls for a principal.
Passing other arguments needs to be handled for deleting specific ACL. Ignore if it is on todo.

wtrocki:
Not sure if we should allow all accounts as well?

```

[wtrocki@graphapi app-services-cli (delete-acl ✗)]$ rhoas kafka acl delete --all-accounts         
? All ACLs matching the criteria provided will be deleted from the Kafka instance "wtrocki". Are you sure you want to proceed? Yes


✔️  Deleted 4 ACLs from Kafka instance "wtrocki"
The following ACL rules were deleted:

  PRINCIPAL (4)    PERMISSION   OPERATION          DESCRIPTION     
 ---------------- ------------ ------------------ ---------------- 
  All Accounts     allow        describe-configs   topic is "*"    
  All Accounts     allow        describe           group is "*"    
  All Accounts     allow        describe           cluster is "*"  
  All Accounts     allow        describe           topic is "*"  
  
  ```

**Comment by wtrocki** [comment]
Seeing some strange issue/inconsitencies:

```
[wtrocki@graphapi app-services-cli (delete-acl)]$ rhoas kafka acl list                    

  PRINCIPAL (5)    PERMISSION   OPERATION   DESCRIPTION               
 ---------------- ------------ ----------- -------------------------- 
  wtrocki          allow        create      topic starts with "test"  
  wtrocki          allow        describe    topic starts with "test"  
  wtrocki          allow        write       topic starts with "test"  
  wtrocki          allow        describe    transactional-id is "*"   
  wtrocki          allow        write       transactional-id is "*"   
[wtrocki@graphapi app-services-cli (delete-acl)]$ rhoas kafka acl delete --service-account=wtrocki
? All ACLs matching the criteria provided will be deleted from the Kafka instance "wtrocki". Are you sure you want to proceed? Yes


✔️  Deleted 2 ACLs from Kafka instance "wtrocki"
The following ACL rules were deleted:

  PRINCIPAL   PERMISSION   OPERATION   DESCRIPTION              
 ----------- ------------ ----------- ------------------------- 
  wtrocki     allow        describe    transactional-id is "*"  
  wtrocki     allow        write       transactional-id is "*"  
  
```

**Comment by wtrocki** [comment]
Ok. So problem is with Pattern Type. because we assume Literal pattern type is provided by default there is no easy way for us to fix that without breaking CLI. I will need to add explicit pattern-type=none and deprecated --prefix to get around this

**Comment by wtrocki** [comment]
@craicoverflow @rkpattnaik780 I have managed to workaround problem by introducing new flag called pattern type and deprecated prefix. 

Not sure if we should be deprecating prefix. I'm testing all gazillion test use cases that this change introduced

**Comment by craicoverflow** [comment]
Pattern type is fine to use. I used "prefix" because this is what confluent CLI does.

**Comment by rkpattnaik780** [comment]
> Not sure if we should allow all accounts as well?

UI allows deleting permissions for all accounts. We should keep it too.

File: pkg/cmd/kafka/acl/delete/delete.go [hunk 1]
```diff
@@ -226,11 +256,20 @@ func validateAndSetOpts(opts *aclcmdutil.CrudOptions) error {
 		return opts.Localizer.MustLocalizeError("kafka.acl.common.error.noPrincipalsSelected")
 	}
 
-	opts.PatternType = aclcmdutil.PatternTypeLITERAL
+	// Backwards compatibility:
```
rkpattnaik780, line 259: This should be moved after the switch in 264.
Current behavior keeps pattern type as literal if we pass `--prefix` flag.

```
./rhoas kafka acl delete --user enda --topic "rhoas-" --prefix -v
```
```
/rest/acls?patternType=LITERAL&permission=ANY&principal=User%3Aenda&resourceName=rhoas-&resourceType=TOPIC
```

pkg/core/localize/locales/en/cmd/acl.en.toml
```diff
@@ -85,6 +85,9 @@ one = 'Set the resource type to cluster'
 [kafka.acl.common.flag.prefix.description]
 one = 'Determine if the resource should be exact match or prefix'
 
+[kafka.acl.common.flag.patterntypes.description]
+one = 'Determine if the resource should be exact match, prefix or any {{.Types}}'
```
rkpattnaik780, line 89: The flag description looks odd

```
Determine if the resource should be exact match, prefix or any [any literal prefix] (default "literal")
```

Should we keep options in description only?

rkpattnaik780 (changes requested):

pkg/cmd/kafka/acl/delete/delete.go
wtrocki: Oh.. I did refactor if into switch and introduced that bug. Really hate that golanglint forces us to use switch statements even when they do not make sense

pkg/core/localize/locales/en/cmd/acl.en.toml
wtrocki: It is good to give list of values in canonical form. Will remove them from description

craicoverflow:
Screenshot 2022-02-17 at 09 12 40

What is this submodule for?

wtrocki:
> What is this submodule for?

That was different POC. Removing it

craicoverflow (approved):
Verified. I left one suggestion regarding printing deprecation message from within the command output aswell as in the description.

wtrocki:
> You could always look to remove that rule from the lint config file? It is not a rule coming from the Go stdlib or anything.

Rule itself is good, but it does have lots of false positives - where doing switch is not possible. 
I will need to start disabling linting for this instead.

pkg/core/localize/locales/en/cmd/acl.en.toml
```diff
@@ -291,6 +294,9 @@ $ rhoas kafka acl delete --operation write --permission allow --topic all --user
 # Delete an ACL for a service account
 $ rhoas kafka acl delete --operation all --permission any --topic "rhoas" --prefix --service-account "srvc-acct-11924479-43fe-42b4-9676-cf0c9aca81"
 
+# Delete all ACLs for a service account
+$ rhoas kafka acl delete --service-account "srvc-acct-11924479-43fe-42b4-9676-cf0c9aca81 --pattern-type=all"
```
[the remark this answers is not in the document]
wtrocki: Yes. While it is counter intuitive we need backwards compatibility.

So:

- nothing provided - exact match
- all - take all
- prefix (use prefix)

etc.

rkpattnaik780 (approved):

pkg/cmd/kafka/acl/delete/delete.go
```diff
@@ -92,6 +91,19 @@ func NewDeleteCommand(f *factory.Factory) *cobra.Command {
...
+		&patternTypeFlag,
+		"pattern-type",
+		aclcmdutil.PatternTypeLITERAL,
+		opts.Localizer.MustLocalize("kafka.acl.common.flag.patterntypes.description",
+			localize.NewEntry("Types", aclcmdutil.PatternTypes)),
+	)
+
+	_ = cmd.RegisterFlagCompletionFunc("pattern-type", func(cmd *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return aclcmdutil.PatternTypes, cobra.ShellCompDirectiveNoSpace
+	})
```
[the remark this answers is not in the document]
craicoverflow: The package already exists though, so don't see how (there is no extra code involved).

pkg/core/cmdutil/flagutil/deprecation.go
```diff
@@ -0,0 +1,6 @@
+package flagutil
+
+// DeprecateFlag provides a way to deprecate a flag by appending standard prefixes to the flag description.
+func DeprecateFlag(flagDescription string) string {
+	return "DEPRECATED: " + flagDescription
+}
```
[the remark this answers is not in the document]
wtrocki: Fully agree.  I will remove that method as it is not used right now and it is not properly implemented.

PR #1444, opened 2022-02-16
feat(service-registry rule): add disable and and update command

rkpattnaik780:
Disable and update command for service-registry rules.

### Verification Steps

1. Run the following command to update config value of an enabled rule:
```
rhoas service-registry rule update --rule-type compatibility --config forward 
```
2. Run the following command to disable all global rules.
```
rhoas service-registry rule disable 
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/docs/commands/rhoas_service-registry_rule.md
+++ b/docs/commands/rhoas_service-registry_rule.md
@@ -35,6 +35,8 @@ $ rhoas service-registry rule describe --rule-type=validity
 
 * [rhoas service-registry](rhoas_service-registry.md)	 - Service Registry commands
 * [rhoas service-registry rule describe](rhoas_service-registry_rule_describe.md)	 - Display configuration of a rule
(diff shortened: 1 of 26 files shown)


rkpattnaik780:
> Me thinking: can disable be just flag as part of the update? Does this make your life easier? If not ignore

I will prefer having a separate disable command.

rkpattnaik780:
@wtrocki I wanted update to have interactive mode too, for suggesting valid config values. The problem is that unlike `enable`, update needs to check if a rule exists (is already enabled) , for which we will need values of `registry-instance ID`, `group` and `artifact-id` beforehand.

wtrocki:
verifying now

wtrocki:
@rkpattnaik780 I see some strange behaviour where text with progress disappears 

> Disabling rule.. 

and then it is gone in second

wtrocki (approved):
Verified *all* commands:

- Global rules
- rules for artifact. 

Made incompatible changes with rule applied and artifact update was rejected.

PR #1447, opened 2022-02-17
fix: clarify errors for kafka create

wtrocki:
Switching to support all error types that we can get back from the API


wtrocki:
FYI @machi1990

craicoverflow (approved):

PR #1463, opened 2022-03-02
ci: notify tools repo when there is a new release

craicoverflow:
Sends a workflow dispatch event to https://github.com/redhat-developer/app-services-tools to initiate a new image release to Quay.

There is no way to verify this until our next release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-name: goreleaser
+name: Release
 
 on:
   push:
@@ -6,7 +6,7 @@ on:
       - 'v*'
 
 jobs:
-  goreleaser:
+  release:
     runs-on: ubuntu-latest
     steps:
       -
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -72,4 +72,21 @@ jobs:
...
+    needs: release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Notify tools repo
+        uses: actions/github-script@v6
+        with: 
+          script: |
+            github.request('POST /repos/redhat-developer/app-services-tools/actions/workflows/20880234/dispatches', {
+              workflow_id: 20880234,
```
craicoverflow, line 87: Workflow ID from here: https://api.github.com/repos/redhat-developer/app-services-tools/actions/workflows/20880234

wtrocki (approved):

.github/workflows/release.yml
wtrocki: Better way of triggering jobs?

craicoverflow: it is different to repository dispatch, not sure if it is objectively better or not. It is the reason we can trigger manually with a form and also can be triggered on non-main branches so it is a bit more flexible.

wtrocki: Reason I'm asking is that I would need to setup another repo for that today and was wondering which job to use. 
Will try this now. Thank you

PR #1470, opened 2022-03-11
fix: changes required for Kafka SDK update

wtrocki:
I have verified all commands manually:

- Topics
- Consumer-Groups
- ACLs

Just minor changes were required.

### DIFF
--- a/pkg/cmd/kafka/consumergroup/resetoffset/reset_offset.go
+++ b/pkg/cmd/kafka/consumergroup/resetoffset/reset_offset.go
@@ -133,8 +133,11 @@ func runCmd(opts *options) error {
 		return err
 	}
 
+	offsetEnum, _ := kafkainstanceclient.NewOffsetTypeFromValue(opts.offset)
 	offsetResetParams := kafkainstanceclient.ConsumerGroupResetOffsetParameters{
-		Offset: opts.offset,
+		Offset: *offsetEnum,
+		Value:  new(string),
+		Topics: &[]kafkainstanceclient.TopicsToResetOffset{},
 	}
 
 	if opts.value != "" {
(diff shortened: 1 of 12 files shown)


pkg/cmd/kafka/consumergroup/resetoffset/reset_offset.go
```diff
@@ -133,8 +133,11 @@ func runCmd(opts *options) error {
 		return err
 	}
 
+	offsetEnum, _ := kafkainstanceclient.NewOffsetTypeFromValue(opts.offset)
```
wtrocki, line 136: Not sure if we should handle error as we are doing validation before so we are guaranteed to have valid offset.

pkg/shared/connection/api/defaultapi/default_client.go
```diff
@@ -154,7 +154,7 @@ func (a *defaultAPI) KafkaAdmin(instanceID string) (*kafkainstanceclient.APIClie
 		apiHost := fmt.Sprintf("admin-server-%v", host)
 		apiURL, _ = url.Parse(apiHost)
 		apiURL.Scheme = "https"
-		apiURL.Path = "/rest"
+		apiURL.Path = "/"
```
wtrocki, line 157: This should have zero impact when we move to /api/v1

pkg/cmd/kafka/consumergroup/resetoffset/reset_offset.go
rkpattnaik780: We can skip handling here.

rkpattnaik780 (approved):

wtrocki:
@rkpattnaik780 I will merge so we can do automatic tests and rebase other PRs

PR #1471, opened 2022-03-16
Kafka Plan/Size implementation for RHOAS CLI based on mock

wtrocki:
This PR uses alpha version of the go sdk that has new changes in cloud providers and new instance types endpoint. 
PR will be used as test bed for the new API

## Progress

- [x] Preset valid kafka sizes to user when suggesting changes
- [x] Validate kafka sizes when provided from flag
- [x] Present list of the sizes when in interactive mode
- [ ] Include cloud provider size when validating selected sizes


pkg/shared/accountmgmtutil/ams.go
```diff
@@ -72,3 +73,25 @@ func GetOrganizationID(ctx context.Context, conn connection.Connection) (account
 
 	return account.Organization.GetId(), nil
 }
+
+// PickInstanceType - Standard instance always wins!!!
+// This function should not exist but it does represents some requirement
+// from business to only pick one instance type when two are presented.
+// When standard instance type is present in user instances it should always take precedence
+func PickInstanceType(amsType *[]string) (string, error) {
```
wtrocki, line 81: @pb82 Because new API returns both standard and developer types 
client logic is really complex and cannot be made generic.

We will need to hardcode this logic and possibly break cli
later phase.

Do you think it is possible to check user ams quota types on backend and return only
instance sizes for the quota that user have access to

wtrocki: Lets chat about it next week

pkg/cmd/kafka/create/create.go
```diff
@@ -297,6 +317,22 @@ func runCreate(opts *options) error {
...
+	}
+	sizes, err := kafkautil.GetValidSizes(conn, opts.Context, opts.provider, opts.region, &amsType)
+	if err != nil {
+		return err
+	}
+	if !slices.Contains(sizes, opts.size) {
+		// TODO error message
+		return errors.New("Whatever") //opts.localizer.MustLocalizeError("")
+	}
+	return nil
```
wtrocki, line 333: We also need cloud region value here to be able to match if size is allowed

wtrocki: Thinking about it - would it be easier to return from API only sizes that will fit to that region?

pkg/shared/kafkautil/util.go
```diff
@@ -90,6 +92,48 @@ func GetCloudProviderRegionCompletionValues(f *factory.Factory, providerID strin
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, instanceType := range instanceTypes.GetItems() {
+		if amsType != nil && amsType != instanceType.Id {
+			continue
+		}
+		for _, instanceSize := range instanceType.GetSizes() {
+			validSizes = append(validSizes, instanceSize.GetId())
```
wtrocki, line 131: So method is getting instance types with their plans.
Should we name types/fields as plan so there is less confusion

wtrocki:
Currently every client needs 3 different API to create kafka + some business logic that will likely to be breaking our clients in future.

## Current workflow 

1. First we do call AMS to get specific quotas. 
Those quotas can change and break clients. 

2. Then we call cloud regions endpoint to get provider and region.
3. We use that data to fetch kafka plans endpoint. 
4. Once that happens we need to look for items with quota we have.
5.  Pick one that should be preferred (standard over trial/developer) and 
6. then make call to the cloud providers again - this time to check and validate size (endpoint returns max)

##  How API could work 

1. Get cloud providers and regions 
2. Fetch kafka plans and api will return only the ones that user have quota and also the sizes that are allowed in the specified region

I'm not sure if it is even feasible to consider changing API at this point but there is significant number of complexity 
in the way that api operates now that prevents us to implement it in client in non breaking way.

craicoverflow (commented):
Scanned the code and it looks fine, could not run it though.

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/redhat-developer/app-services-cli/pkg/shared/connection"
+	"k8s.io/utils/strings/slices"
```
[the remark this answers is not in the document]
wtrocki: Good idea. Will create util method and replace it everywhere

pkg/cmd/kafka/create/create.go
```diff
@@ -119,6 +123,7 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 	flags.StringVar(&opts.name, "name", "", opts.localizer.MustLocalize("kafka.create.flag.name.description"))
 	flags.StringVar(&opts.provider, kafkaFlagutil.FlagProvider, "", opts.localizer.MustLocalize("kafka.create.flag.cloudProvider.description"))
 	flags.StringVar(&opts.region, kafkaFlagutil.FlagRegion, "", opts.localizer.MustLocalize("kafka.create.flag.cloudRegion.description"))
+	flags.StringVar(&opts.size, "size", "", opts.localizer.MustLocalize("kafka.create.flag.size.description"))
```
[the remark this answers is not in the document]
wtrocki: I have written method that suggest values from API and validate them as well. 
Logic for obtaining those values is quite complex right now so I'm working on proposal to change API slightly to make it possible to only provide values that are valid.

wtrocki:
Jesus - so many conflicts :D

PR #1484, opened 2022-03-23
refactor(service-registry): get current instance from context file

rkpattnaik780:
Service registry commands now fetch currently selected instance from service-context rather than service config.

### Verification Steps

Following commands have been modified and hence need to be verified:

Control plane commands - create delete, describe, use
Data plane commands - artifact, rule, role

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


rkpattnaik780:
Noticed the cli makes same request twice when not passing the instance ID. Will get that fixed in subsequent PRs.

PR #1498, opened 2022-03-30
fix: refactor status command

wtrocki:
## Motivation

I really do think that doing some form of factory object mapping to custom opts structure is total waste of time and doesn't fit much of purpose. Factory is good abstraction on it's own. Removed it from this command specifically.

Added support for name and location

Added common handling for services missing so we can now reuse it.

### DIFF
--- a/pkg/shared/contextutil/util.go
+++ b/pkg/shared/contextutil/util.go
@@ -58,6 +58,7 @@ func GetCurrentKafkaInstance(f *factory.Factory) (*kafkamgmtclient.KafkaRequest,
 	return GetKafkaForServiceConfig(currCtx, f)
 }
 
+// GetKafkaForServiceConfig fetching kafka service data using ServiceConfig
 func GetKafkaForServiceConfig(currCtx *servicecontext.ServiceConfig, f *factory.Factory) (*kafkamgmtclient.KafkaRequest, error) {
 	conn, err := f.Connection(connection.DefaultConfigRequireMasAuth)
 	if err != nil {
@@ -92,6 +93,7 @@ func GetCurrentRegistryInstance(f *factory.Factory) (*registrymgmtclient.Registr
 
 }
 
+// GetRegistryForServiceConfig fetching registry data for provided service config
 func GetRegistryForServiceConfig(currCtx *servicecontext.ServiceConfig, f *factory.Factory) (*registrymgmtclient.Registry, error) {
 	conn, err := f.Connection(connection.DefaultConfigRequireMasAuth)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/shared/contextutil/util.go
```diff
@@ -81,12 +83,17 @@ func GetCurrentRegistryInstance(f *factory.Factory) (*registrymgmtclient.Registr
...
 	if err != nil {
 		return nil, err
 	}
 
-	currCtx, err := GetCurrentContext(svcContext, f.Localizer)
+	return GetRegistryForServiceConfig(currCtx, f)
+
+}
+
+func GetRegistryForServiceConfig(currCtx *servicecontext.ServiceConfig, f *factory.Factory) (*registrymgmtclient.Registry, error) {
```
rkpattnaik780, line 95: Add comment.

pkg/shared/contextutil/util.go
```diff
@@ -50,16 +50,19 @@ func GetCurrentKafkaInstance(f *factory.Factory) (*kafkamgmtclient.KafkaRequest,
...
+	currCtx, err := GetCurrentContext(svcContext, f.Localizer)
 	if err != nil {
 		return nil, err
 	}
 
-	currCtx, err := GetCurrentContext(svcContext, f.Localizer)
+	return GetKafkaForServiceConfig(currCtx, f)
+}
+
+func GetKafkaForServiceConfig(currCtx *servicecontext.ServiceConfig, f *factory.Factory) (*kafkamgmtclient.KafkaRequest, error) {
```
rkpattnaik780, line 61: Add comment.

rkpattnaik780 (commented):
Works great! I think we should still show status if a service is missing.

pkg/shared/contextutil/util.go
wtrocki:
```suggestion
// GetRegistryForServiceConfig fetching registry data for provided service config
func GetRegistryForServiceConfig(currCtx *servicecontext.ServiceConfig, f *factory.Factory) (*registrymgmtclient.Registry, error) {
```

pkg/shared/contextutil/util.go
wtrocki:
```suggestion
func GetKafkaForServiceConfig(currCtx *servicecontext.ServiceConfig, f *factory.Factory) (*kafkamgmtclient.KafkaRequest, error) {
```
```suggestion
// GetKafkaForServiceConfig fetching kafka service data using ServiceConfig
func GetKafkaForServiceConfig(currCtx *servicecontext.ServiceConfig, f *factory.Factory) (*kafkamgmtclient.KafkaRequest, error) {
```

PR #1499, opened 2022-03-30
feat(context): add generate-config command

rkpattnaik780:
Added `generate-config` command for contexts.

Closes # 

### Verification Steps

1. Run the following command to generate configuration file for your context.
```
./rhoas context generate-config --type env
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


pkg/cmd/context/generate/build-configs.go
```diff
@@ -0,0 +1,101 @@
...
+		Localizer:      opts.localizer,
+		Connection:     opts.Connection,
+		ServiceContext: opts.ServiceContext,
+	}
+
+	configurations := &configValues{}
+
+	var serviceAvailable bool
+
+	kafkaInstance, _ := contextutil.GetCurrentKafkaInstance(factory)
```
wtrocki, line 63: not handled error

pkg/cmd/context/generate/build-configs.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if kafkaInstance != nil {
+		serviceAvailable = true
+		configurations.KafkaURL = kafkaInstance.GetBootstrapServerHost()
+	}
+
+	registryInstance, _ := contextutil.GetCurrentRegistryInstance(factory)
```
wtrocki, line 70: Same profile might be broken

pkg/cmd/context/generate/build-configs.go
rkpattnaik780: What if use wants configuration to be generated with only one service?

pkg/cmd/context/generate/configurations.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"bytes"
+	"html/template"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+var (
```
wtrocki, line 11: For readability I would keep templates in separate file

pkg/cmd/context/generate/build-configs.go
wtrocki: Generating config on invalid context should not be allowed. Can we unset service from context?

wtrocki: Like this would 100% be logged as bug. I have my service in context but not getting url in config. See my handling for status that does similar thing. One service breaking context returns error for entire process no matter what

wtrocki: #1498

wtrocki (approved):
Works amazing. We need to add property support

pkg/cmd/context/generate/build-configs.go
rkpattnaik780: Unsetting a service is not there yet. A freshly created context can have an unset service.

wtrocki: Actually I see problem now. When registry is not set it will return error
but for this case this is ok.

wtrocki: I would say that usage of those helper methods doesn't fit to this use case. We actually need to duplicate logic for those for purpose of generation.

wtrocki (approved):

PR #1501, opened 2022-03-31
fix(context generate-config): remove hard codings

rkpattnaik780:
Add short description for created service account.
Removed hard coded strings, added flag validation and completion for `--type`.

### Verification Steps

1. The following command should create a service account with short description with prefix "rhoascli-"
`./rhoas context generate-config --type properties`
2. The following command should throw an error indicating invalid flag value.
`./rhoas context generate-config --type propertiees`

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/context/generate/build-configs.go
+++ b/pkg/cmd/context/generate/build-configs.go
@@ -87,7 +87,7 @@ func BuildConfiguration(svcConfig *servicecontext.ServiceConfig, opts *options) 
 		return opts.localizer.MustLocalizeError("context.generate.log.info.noSevices")
 	}
 
-	serviceAccount, err := createServiceAccount(opts, fmt.Sprintf("rhoascli-%v", time.Now().Unix()))
+	serviceAccount, err := createServiceAccount(opts, fmt.Sprintf("%s-%v", opts.name, time.Now().Unix()))
 	if err != nil {
 		return err
 	}
--- a/pkg/cmd/context/generate/build-configs.go
+++ b/pkg/cmd/context/generate/build-configs.go
@@ -1,6 +1,9 @@
 package generate
 
 import (
+	"fmt"
+	"time"
+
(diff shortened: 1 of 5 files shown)


pkg/cmd/context/generate/build-configs.go
```diff
@@ -84,7 +87,7 @@ func BuildConfiguration(svcConfig *servicecontext.ServiceConfig, opts *options)
 		return opts.localizer.MustLocalizeError("context.generate.log.info.noSevices")
 	}
 
-	serviceAccount, err := createServiceAccount(opts, "to-do-generate-description")
+	serviceAccount, err := createServiceAccount(opts, fmt.Sprintf("rhoascli-%v", time.Now().Unix()))
```
wtrocki, line 90: Wondering - should that be profile name instead? How we would be able to find that SA if not by it's name.

wtrocki: It is important to be able to revoke it once it was exposed etc.

rkpattnaik780: That makes sense. Have updated the PR.

pkg/cmd/context/generate/build-configs.go
```diff
@@ -84,7 +87,7 @@ func BuildConfiguration(svcConfig *servicecontext.ServiceConfig, opts *options)
 		return opts.localizer.MustLocalizeError("context.generate.log.info.noSevices")
 	}
 
-	serviceAccount, err := createServiceAccount(opts, "to-do-generate-description")
+	serviceAccount, err := createServiceAccount(opts, fmt.Sprintf("%s-%v", opts.name, time.Now().Unix()))
```
wtrocki, line 90: I done some testing I think that might require us to validate name to not have spaces

rkpattnaik780: The regexp for context name has been matched with that of service account short description.

PR #1505, opened 2022-04-01
Plugin architecture

wtrocki:
## What
We need to enable rhoas cli to support plugins.

Plugins should be either:

- Embedded into rhoas CLI at runtime
- Can be released and updated separately (different repositories)

RHOAS CLI should support ecosystem to build production ready command line tools using the Cobra Framework. We should have minimal SDK which allows for the creation of CLIs that all adhere to a similarly structured architecture. It provides a level of elasticity to add and/ or remove a plugin CLI or subset of commands from a plugin CLI to a main 'host' CLI.

## Architecture

Feature should provide a way to build a CLI ecosystem where developers can build multiple fragmented CLI's in various repositories and embed them later in the single host CLI. Plugin engine should provide core base functionality that helps to abstract various elements of the CLI ecosystem:

- Logging
- Authentication
- Configuration
- Commands and Flags standards.

SDK will typically be introduced into two different CLIs:

RHOAS Host CLI - CLI that is being used by end-user that can embed one or more plugin CLIs
Plugin CLIs - Separate CLIs that can be embedded into the host CLI at runtime or build process.

## Open questions

Should RHOAS CLI contain some plugins at build time/runtime and dynamically switch to the external ones when update appears
Where to host plugin registry (Is github pages sufficient)


wtrocki:
This is early version of plugins architecture. I see numerous challenges related to the plugins - marked as TODO.
In follow up work I will be testing and creating SDK inside core CLI to make features work as standalone plugins

wtrocki:
Done some investigation and I think that we need to make 2 main decisions to move it further:

1) Are we allowing development of separate CLIs outside this repo
2) Do we actually need plugins CLI executable on it's own or we can have them as Plugins/Go modules (so files)
3) Do we need plugins to update separately or can we bundle all plugins into single SO/github release file?

Answering those questions will help us to move this POC further

plugins/request/.goreleaser.yml
```diff
@@ -0,0 +1,66 @@
...
+      name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
+    wrap_in_directory: "true"
+    replacements:
+      darwin: macOS
+    format: tar.gz
+
+  - id: windows
+    builds: [windows]
+    <<: *archive_defaults
+    wrap_in_directory: false
```
rkpattnaik780, line 52: Double quotes missing

rkpattnaik780:
> Done some investigation and I think that we need to make 2 main decisions to move it further:
> 
> 1. Are we allowing development of separate CLIs outside this repo (that will have significant drawbacks)

I dont think separate repo for plugins is required as of now. Just curious what the drawbacks will be.

> 2. Do we actually need plugins CLI executable on it's own or we can have them as Plugins/Go modules (so files)

Is it absolutely necessary to run them as standalone CLIs?

> 3. Do we need plugins to update separately or can we bundle all plugins into single SO/github release file?

Separation sounds good.

wtrocki:
Closing - This have been proven to be very complex architecture. We can always review this later.

PR #1514, opened 2022-04-08
fix: update go

wtrocki:


rkpattnaik780:
I was expecting the go version in mod file to be updated as well.

PR #1517, opened 2022-04-09
feat: add token command

ajaypratap003:
Closes # https://issues.redhat.com/browse/MGDSTRM-8103

### Verification Steps
Run the following command
```
##Generate current user auth token
rhoas auth-token
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### Checklist

- [ ] Code Review completed

### DIFF
--- a/pkg/cmd/token/token.go
+++ b/pkg/cmd/token/token.go
@@ -1,0 +1,64 @@
+package token
+
+import (
+	"fmt"
+
+	"github.com/redhat-developer/app-services-cli/pkg/core/config"
+	"github.com/redhat-developer/app-services-cli/pkg/core/ioutil/iostreams"
+	"github.com/redhat-developer/app-services-cli/pkg/core/localize"
+	"github.com/redhat-developer/app-services-cli/pkg/core/logging"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/connection"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/factory"
+	"github.com/spf13/cobra"
+)
+
+type options struct {
+	Config     config.IConfig
+	Connection factory.ConnectionFunc
+	IO         *iostreams.IOStreams
+	Logger     logging.Logger
+	localizer  localize.Localizer
+}
+
+func NewTokenCmd(f *factory.Factory) *cobra.Command {
+	opts := &options{
+		Config:     f.Config,
+		Connection: f.Connection,
+		IO:         f.IOStreams,
+		Logger:     f.Logger,
+		localizer:  f.Localizer,
+	}
+
+	cmd := &cobra.Command{
+		Use:     "token",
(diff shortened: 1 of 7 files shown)


pkg/cmd/token/token.go
```diff
@@ -0,0 +1,64 @@
...
+	opts := &options{
+		Config:     f.Config,
+		Connection: f.Connection,
+		IO:         f.IOStreams,
+		Logger:     f.Logger,
+		localizer:  f.Localizer,
+	}
+
+	cmd := &cobra.Command{
+		Use:     "token",
```
wtrocki, line 33: @rkpattnaik780 we have whoami and token now. 
Maybe it is worth to introduce auth and keep two commands there?

Otherwise we can do

```suggestion
		Use:     "auth-token",
```

wtrocki (approved):
World class work! Thank you so much.

wtrocki:
FYI token we return is control plane token. 
We could support dataplane token as well but I think it doesnt make sense now since we migrating out of that sso instance

pkg/cmd/token/token.go
wtrocki: Alternatively we can put token as hidden command

ajaypratap003: @wtrocki  I updated cmd name as auth-token but build is failing due to test failing, seems `validator_test` is failing due to  Punctuation rule.

wtrocki: That was actually bad advice on my side. I will wait for @rkpattnaik780 opinion about the command name and we can merge it.

rkpattnaik780: `whoami` command looks good where it is right now, I dont feel good about introducing "auth".

rkpattnaik780: > @wtrocki I updated cmd name as auth-token but build is failing due to test failing, seems `validator_test` is failing. If I'm not wrong validator_test.go rules doesn't allow to use `- dash` in cmd name.


[Examples](https://github.com/redhat-developer/app-services-cli/blob/c8aff19a3673d961ba409277d4487a7bef29a1f1/pkg/core/localize/locales/en/cmd/token.en.toml#L13) need to be updated.

wtrocki: Command name is not going to pass validation. @rkpattnaik780 do you mind adjusting name to authtoken or maybe we can have separate group for whoami and token?

rkpattnaik780 (approved):
Looks great! :D

pkg/cmd/token/token.go
rkpattnaik780: > do you mind adjusting name to authtoken or maybe we can have separate group for whoami and token?

authtoken seems a better name. I strongly feel whoami should be at root.

wtrocki:
Changing name in separate PR. All changes will be merged

PR #1521, opened 2022-04-11
fix: suspend updating modular docs for prereleases

wtrocki:


wtrocki:
CC @bhardesty we going to get sync PR that should be rejected as it would contain changes that haven't been really released. We are planning release tommorow

PR #1526, opened 2022-04-19
Changes for Kafka Creation business logic

wtrocki:
Add ability to create kafka with the plan specified by users.
In this PR we implement logic that  does contact AMS to pick instance types. Then get extra cloud providers and regions.


wtrocki:
Rebased version of [#create](https://github.com/redhat-developer/app-services-cli/pull/1471)

pkg/cmd/kafka/create/data.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Matches QuotaStandardType = standard
+	StandardType CloudProviderId = "standard"
+
+	// Deprecated by DeveloperType
+	TrialType CloudProviderId = "eval"
+)
+
+// mapAmsTypeToBackendType - Cloud providers API is not using AMS types but some other values (CloudProviderValues)
+func mapAmsTypeToBackendType(amsType *accountmgmtutil.QuotaSpec) CloudProviderId {
```
wtrocki, line 27: Area of interest 1

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -48,19 +65,64 @@ func GetUserSupportedInstanceTypes(ctx context.Context, spec remote.AmsConfig, c
...
+
+	return BattleOfInstanceBillingModels(quotas), err
+}
 
-		if quotaId == spec.TrialQuotaID {
-			quotas = append(quotas, QuotaTrialType)
+// This function selects the billing model that should be used
+// It represents some requirement to always use the same standard billing models
+// This function should not exist but it does represents some requirement that we cannot do on backend
+func BattleOfInstanceBillingModels(quotas []QuotaSpec) []QuotaSpec {
```
wtrocki, line 88: Area of interest 2

pkg/cmd/kafka/create/data.go
pmuir: Do we plan to update Account Management Service types to match the kas-fleet-manager types so we can avoid this mapping on the client side? If not, we probably need to add the Account Management Service type to the kas-fleet-manager type. Then the client would need to create a map still, but wouldn't have to hardcode all the mapped values.

pkg/cmd/kafka/create/data.go
```diff
@@ -0,0 +1,160 @@
...
+		Execute()
+	if err != nil {
+		return nil, err
+	}
+
+	var desiredInstanceType string
+	if amsType != nil {
+		desiredInstanceType = mapAmsTypeToBackendType(amsType)
+	} else {
+		// If we do not know instance type from AMS we should assume standard instance type
```
pmuir, line 61: This logic seems dangerous to me. The kas-fleet-manager api should somehow provide the default value.

pkg/cmd/kafka/create/data.go
```diff
@@ -0,0 +1,160 @@
+package create
```
pmuir, line 1: @wtrocki what do you think about extracting some of this logic to the SDKs when it's finalised so that it is reusable across clients?

pkg/cmd/kafka/create/data.go
pmuir: In other words, I think the else here should be an error condition.

pkg/cmd/kafka/create/data.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	cloudProviderResponse, _, err := conn.API().KafkaMgmt().GetCloudProviders(f.Context).Execute()
+	if err != nil {
+		return validProviders, err
+	}
+
+	cloudProviders := cloudProviderResponse.GetItems()
+	cloudProviderNames := []string{}
+	for _, provider := range cloudProviders {
```
pmuir, line 95: Could we avoid this if we added a filter to `GetCloudProviders`?

pkg/cmd/kafka/create/data.go
wtrocki: Agree. That will reduce business logic on the client by half. 

Additionally AMS should probably not return us instances that we cannot create (like Trial instance when standard is there), however I'm not sure if that would be possible to implement on AMS side

pkg/cmd/kafka/create/data.go
```diff
@@ -0,0 +1,160 @@
...
+	cloudProviderResponse, _, err := conn.API().
+		KafkaMgmt().
+		GetCloudProviderRegions(f.Context, providerID).
+		Execute()
+	if err != nil {
+		return validRegions, err
+	}
+
+	var serverInstanceType string
+	if userAllowedAMSInstanceType != nil {
```
pmuir, line 130: Would the backend implementing a filter by Account Management Service instance type help here?

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -316,11 +323,14 @@ one = 'name is required. Run "rhoas kafka create --name my-kafka"'
 [kafka.create.error.conflictError]
 one = 'Kafka instance "{{.Name}}" already exists'
 
-[kafka.create.error.oneinstance]
-one = 'maximum number of allowed instances reached. Please review all instances that your user has access to and delete one or more instances before creating a new one.'
+[kafka.create.error.userInstanceType.notFound]
+one = 'Cannot fetch user allowed instance type"'
```
pmuir, line 327: I'm not sure this error message is correct. What is a "user allowed instance type"? I think this should be more like "Cannot fetch entitlements"

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -48,19 +65,64 @@ func GetUserSupportedInstanceTypes(ctx context.Context, spec remote.AmsConfig, c
 		return nil, err
 	}
 
-	var quotas []string
+	var quotas []QuotaSpec
 	for _, quota := range quotaCostGet.GetItems() {
-		quotaId := strings.TrimSpace(quota.GetQuotaId())
+		for _, quotaResource := range quota.GetRelatedResources() {
+			if quotaResource.GetResourceName() == spec.ResourceName {
+				if quotaResource.GetProduct() == spec.TrialProductQuotaID {
```
pmuir, line 72: This is because we don't properly support quota for trials on Account Management Service right?

pkg/shared/accountmgmtutil/ams.go
pmuir: Billing models shouldn't be exposed to clients at all.

pmuir: We need to somehow abstract this properly on the backend.

pkg/cmd/kafka/create/data.go
wtrocki: Yes. I'm still missing here logic that filters out sizes outside availabeQuota. When backend filtering is enabled all we need is just api call

pkg/core/localize/locales/en/cmd/kafka.en.toml
wtrocki: Yes. I have struggled with finding good world/user facing message. Will change that

pkg/shared/accountmgmtutil/ams.go
wtrocki: Yes

pkg/cmd/kafka/create/data.go
wtrocki: I have removed need to have else :)

rkpattnaik780:
Not able to generate the binary, failing with the following error:
```
github.com/redhat-developer/app-services-sdk-go/kafkamgmt/apiv1/client
../../../../go/pkg/mod/github.com/redhat-developer/app-services-sdk-go/kafkamgmt@v0.11.1/apiv1/client/model_cloud_region.go:44:6: this.SupportedInstanceTypes undefined (type CloudRegion has no field or method SupportedInstanceTypes)
```

wtrocki:
@rkpattnaik780 Should be fixed now

pkg/cmd/kafka/create/completions.go
```diff
@@ -0,0 +1,55 @@
...
+
+	conn, err := f.Connection(connection.DefaultConfigSkipMasAuth)
+	if err != nil {
+		return nil, directive
+	}
+
+	userInstanceType, err := accountmgmtutil.GetUserSupportedInstanceType(f.Context, constants.Kafka.Ams, conn)
+
+	// Not including quota in this request as it takes very long time to list quota for all regions in suggestion mode
+	validRegions, _ = GetValidKafkaSizes(f, providerID, regionId, *userInstanceType)
```
rkpattnaik780, line 52:
```suggestion
	validSizes, _ = GetValidKafkaSizes(f, providerID, regionId, *userInstanceType)
```

pkg/cmd/kafka/create/completions.go
```diff
@@ -0,0 +1,55 @@
...
+
+// GetCloudProviderCompletionValues returns the list of supported cloud providers for creating a Kafka instance
+// This is used in the cmd.RegisterFlagCompletionFunc for dynamic completion of --provider
+func GetCloudProviderCompletionValues(f *factory.Factory) (validProviders []string, directive cobra.ShellCompDirective) {
+	validProviders, _ = GetEnabledCloudProviderNames(f)
+
+	return validProviders, cobra.ShellCompDirectiveNoSpace
+}
+
+// GetCloudProviderCompletionValues returns the list of region IDs for a particular cloud provider
```
rkpattnaik780, line 19:
```suggestion
// GetCloudProviderRegionCompletionValues returns the list of region IDs for a particular cloud provider
```

pkg/cmd/kafka/create/completions.go
```diff
@@ -0,0 +1,55 @@
...
+	if providerID == "" {
+		return
+	}
+
+	validRegions, _ = GetEnabledCloudRegionIDs(f, providerID, nil)
+
+	return validRegions, cobra.ShellCompDirectiveNoSpace
+}
+
+// GetEnabledCloudProviderNames returns a list of valid kafka sizes for the specifed region and ams instance types
```
rkpattnaik780, line 30:
```suggestion
// GetKafkaSizeCompletionValues returns a list of valid kafka sizes for the specifed region and ams instance types
```

pkg/cmd/kafka/create/completions.go
```diff
@@ -0,0 +1,55 @@
...
+func GetKafkaSizeCompletionValues(f *factory.Factory, providerID string, regionId string) (validRegions []string, directive cobra.ShellCompDirective) {
```
rkpattnaik780, line 31: Doesnt seem to be returning values.

rkpattnaik780 (changes requested):
Interactive mode seems to be broken.

rkpattnaik780:
> Interactive mode seems to be broken.

```
./rhoas kafka create                                                                               
X Sorry, your reply was invalid: Kafka instance "test-instance" already exists
? Name: test-instance-2
? Cloud Provider: aws
❌ Please provide options to select from. Run the command in verbose mode using the -v flag to see more information
```

wtrocki:
@rkpattnaik780 All fixed. Issue was that quota cost between AMS and kas-fleetshard do not match it. We have added temporary workaround for now.

pkg/cmd/kafka/create/data.go
wtrocki: +1 although I would like to do it once we get some more confidence that this logic is future proof.

pkg/cmd/kafka/create/api_validators.go
```diff
@@ -0,0 +1,134 @@
...
+		if !selectedRegion.Enabled {
+			regionEntry := localize.NewEntry("Region", input.region)
+			validRegionsEntry := localize.NewEntry("Regions", regionsString)
+			providerEntry := localize.NewEntry("Provider", input.provider)
+			return f.Localizer.MustLocalizeError("kafka.create.region.error.invalidRegion", regionEntry, providerEntry, validRegionsEntry)
+		}
+
+		return nil
+
+		// Current API doesn't allow us to validate regions properly.
```
wtrocki, line 98: Not sure if we should handle region out of the capacity problem or leave it for the API result. 

It is possible to implement this but logic will be quite tricky and subject to breaking changes

pkg/cmd/kafka/create/data.go
```diff
@@ -0,0 +1,149 @@
...
+				for _, instanceSize := range instanceType.GetSizes() {
+					validSizes = append(validSizes, instanceSize.GetId())
+				}
+			}
+		}
+	} else {
+		for _, instanceType := range instanceTypes.GetInstanceTypes() {
+			if desiredInstanceType == instanceType.GetId() {
+				for _, instanceSize := range instanceType.GetSizes() {
+					if instanceSize.GetQuotaConsumed() <= int32(amsType.Quota) {
```
pb82, line 72: Ok, I understand now why you're relying on `quota_consumed` being `0` for developer instances.

wtrocki: Ah.. I forgot to changed it. In chat you have suggesed using different value. Does it matter?

pkg/shared/accountmgmtutil/ams.go
pb82: How is this used? Is it to determine what to send in the `plan` field when creating a new instance? The backend can already decide between `standard` and `eval` based on Quota. So technically it's only the size that we need from the UI. Unless of course we add other plans in the future that the user can choose from.

pkg/cmd/kafka/create/data.go
wtrocki: I meant should be using CapacityConsumed or Quota consumed to filter this?

wtrocki: Overal I do not think we should have that much logic around instance types in the client. Limitation right now is that AMS and Backend diverge a lot so we would need to have those values in generic form to build generic client. 

I would even say that if we cannot reduce business logic we can pull out some validation logic you have on backend and reuse it as source of truth for SDK.

pkg/shared/remote/service-constants.json
```diff
@@ -1,19 +1,25 @@
...
-  }
\ No newline at end of file
+  },
+  "serviceRegistry": {
+    "ams": {
+      "termsAndConditionsEventCode": "register",
+      "termsAndConditionsSiteCode": "ocm",
+      "instanceQuotaId": "cluster|rhinfra|rhosr|any",
+      "trialQuotaId": "cluster|rhinfra|rhosrtrial|any",
+      "quotaProductId": "RHOSR",
```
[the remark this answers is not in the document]
wtrocki: This object is for service registry, although this reminded me that we need to tests service registry creation as well

pkg/cmd/kafka/create/api_validators.go
pb82: Backend will also check if the instance type is supported in the region and if there is capacity.

pb82 (approved):
Looks good as far as I can tell. Some minor comments.

pkg/cmd/kafka/create/api_validators.go
wtrocki: Yes. I think while UI for this is not great. I would ignore that use case for both clients and 100% rely on the backend error code. That will save us some problems with breaking this model once we get more instance types/sizes etc.

wtrocki:
Thank you so much @pb82

wtrocki:
Some very small changes in messages @bhardesty 
This PR targets summit timeframe (Kafka sizes)

PR #1533, opened 2022-04-21
fix: remove multiaz flags

wtrocki:
Can be used for testing backend not validating that values anymore:

```
{"id":"35","kind":"Error","href":"/api/kafkas_mgmt/v1/errors/35","code":"KAFKAS-MGMT-35","reason":"Only multiAZ Kafkas are supported, use multi_az=true","operation_id":"c9ghqeq65lndnf0m55h0"}
```


wtrocki:
Cherry-picked to other PR

PR #1542, opened 2022-04-27
refactor(context): rename use-service to set-service

rkpattnaik780:
BREAKING CHANGE: rename rhoas context use-<service> to rhoas context set-<service>

Command has been renamed after getting feedbacks.

### Verification Steps

1. Run the following command to set Kafka instance for the current context
```
rhoas context set-kafka
```

### Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation change
- [ ] Other (please specify)


wtrocki (approved):
I think if we make this change quite fast we can introduce breaking change without issues.
The only thing would be to make sure that downstream docs remove those commands

rkpattnaik780:
> I think if we make this change quite fast we can introduce breaking change without issues. The only thing would be to make sure that downstream docs remove those commands

Yes this requires an update in docs.

PR #1544, opened 2022-04-28
feat(context): add unset command

rkpattnaik780:
`rhoas context unset` command can be used to unset services in the contexts.

### Verification Steps
1. Set a Kafka instance in the context using:
```
rhoas context set-kafka
``` 
2. Check the status of the kafka instance. It should print the kafka details
```
rhoas context status kafka
```
3. Unset Kafka instance for the context
```
rhoas context unset --services kafka
```
4. Check status of Kafka instance again, it should throw error.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/core/localize/locales/en/cmd/context.en.toml
+++ b/pkg/core/localize/locales/en/cmd/context.en.toml
@@ -55,6 +55,36 @@ $ rhoas context use --name dev
 [context.use.successMessage]
 one='Current context set to "{{.Name}}"'
 
+[context.unset.cmd]
+
+[context.unset.cmd.shortDescription]
+one='Unset services in context'
+
+[context.unset.cmd.longDescription]
+one='''
+Unset services in context
+
+When you unset a service in context, it will no longer point to any instance of that service.
+'''
+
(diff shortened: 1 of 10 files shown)


pkg/core/localize/locales/en/cmd/context.en.toml
```diff
@@ -55,6 +55,36 @@ $ rhoas context use --name dev
...
+one='''
+Unset services in context
+
+When you unset a service in context, it will no longer point to any instance of that service.
+'''
+
+[context.unset.cmd.example]
+one='''
+# Unset services for current context
+$ rhoas context unset --services kafka,service-registry
```
rkpattnaik780, line 73: Need feedback on the intuivity of the implementation.

wtrocki (approved):

pkg/core/localize/locales/en/cmd/context.en.toml
wtrocki: It is the same as status command so it is good. Sometimes we might use array string in cobra etc.

PR #1540, opened 2022-04-25
Refactor output format framework to support tables and empty format 

gowriswarupk:
# Motiviation

Addressing issue #1429 

### Changes made:
---
- fix:Kafka topic creation command "mute" successful. [1]

### Verification Steps:
---
- for fix [1] :
`./rhoas kafka topic create --name=kannan --output=none`

Issue #1429, opened 2022-02-12
Allow to "mute" the output of the Kafka topic creation command


ppatierno:
The `rhoas kafka topic create` allows to specify the output of the command as YAML and JSON.
I was thinking that sometimes for automated tasks and avoid to pollute the console could be useful to "mute" the output, exiting ok if the topic was creating but without printing any representation of the topic info. Maybe an `--output none` could be enough.

wtrocki:
I think we might only print one sentence for topic by default. Or use some output as suggested 

As workarounds for now we offer users to use std and error stream pipes.
```
rhoas kafka topic create --name=test -y 2> /dev/null
```

I think we should also document this as it is not obvious pattern we use

PR #1548, opened 2022-04-29
Refactor output format framework to support tables and empty format #1540

gowriswarupk:
# Motiviation

Addressing issue #1429 

### Changes made:
---
- feat Output Format supports 'table' and 'none' output types. [1]

### Verification Steps:
---
-`./rhoas kafka topic create --name test --output none`
-`./rhoas kafka topic list --output table`


Makefile
```diff
@@ -86,12 +86,8 @@ generate-downstream-docs: ## Generate command-line reference documentation in ad
 
 I18N_LINTER_DEF := $(shell command -v app-services-go-linter 2> /dev/null)
 
-lint-lang: ## Lint i18n files
-ifndef I18N_LINTER_DEF # check if the linter is installed, install it if not
-	go install github.com/redhat-developer/app-services-go-linter/cmd/app-services-go-linter@latest
-endif
-	app-services-go-linter -path ./pkg/core/localize/locales ./...
-.PHONY: lint-lang
```
wtrocki, line 94: This should not be there./

go.mod
```diff
@@ -4,21 +4,26 @@ go 1.16
 
 require (
 	github.com/AlecAivazis/survey/v2 v2.3.4
-	github.com/BurntSushi/toml v1.0.0
+	github.com/BurntSushi/toml v1.1.0
```
wtrocki, line 7: WE should not update those

wtrocki:
Needs addressing comments from here: 
https://github.com/redhat-developer/app-services-cli/pull/1540

wtrocki:
@rkpattnaik780 WDYT about this PR. I think we could drive that to main branch with some minimal work.

rkpattnaik780:
> @rkpattnaik780 WDYT about this PR. I think we could drive that to main branch with some minimal work.

That will be great. Few changes need to be made and should be properly tested before merging.

PR #1552, opened 2022-05-04
Addressing feedback for instance sizes

wtrocki:
Refactor sizes to introduce better handling:

- Dot not sent plan for trial kafka. There is no concept of plan for those 
- Fix bypas_ams_check flag breaking creation
- Rename flag to better reflect what it does.
-

### DIFF
--- a/pkg/cmd/kafka/create/api_validators.go
+++ b/pkg/cmd/kafka/create/api_validators.go
@@ -102,18 +102,19 @@ func validateProviderRegion(input *ValidatorInput, selectedProvider kafkamgmtcli
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


wtrocki:
We are still blocked to remove x1,x2 values due to lack of backend support.

wtrocki:
Because of compatibility layer we can avoid specifying size for standard instance. @pb82 do you plan to remove that capability at some point. It is important for us to not break existing clients that create kafka with just specifying name.

pb82:
@wtrocki there are no plans to remove the ability to skip the `plan` field at that time.

PR #1570, opened 2022-05-17
Revert "fix: update kafka mgmt sdk for the latest version"

wtrocki:
Reverts redhat-developer/app-services-cli#1569 that breaks CLI


wtrocki:
@rkpattnaik780 Sadly. Problem was on the backend API generation. We need to revert SDK

PR #1586, opened 2022-06-03
fix: marketplace for friday testing

wtrocki:
Initial version of marketplace we have been using for todays testing.


pkg/cmd/kafka/create/create.go
```diff
@@ -219,6 +226,13 @@ func runCreate(opts *options) error {
 			CloudProvider: &opts.provider,
 		}
 
+		if opts.marketplaceId != "" && opts.marketplaceType != "" {
```
wtrocki, line 229: It looks like we always need to require and send pair of marketplace type and marketplace id

wtrocki:
@rkpattnaik780 key idea here was to be able to test a the cases to see if everything works as expected. You would need to be added to @pb82 organization to perform tests

rkpattnaik780:
Which organization?

wtrocki:
@pb82 to be sure: Are clients always expected to pass cloudAccountId to backend or this is just temporary?

Thinking about case where only one accountz exist

pkg/cmd/kafka/create/create.go
```diff
@@ -363,12 +361,12 @@ func runCreate(opts *options) error {
...
-	Size            string
-	Region          string
-	CloudProvider   string
-	AccountID       string
-	CloudProviderID string
+	Name          string
+	Size          string
+	Region        string
+	CloudProvider string
+	MarketplaceID string
```
wtrocki, line 368: marketplaceID would be `aws` `rhm` etc.  Value that we passing represent "Cloud Provider Account ID" that have been used to purchase kafka.

rkpattnaik780: Referring to the KFM OpenAPI. It seems that would be [marketplace](https://github.com/redhat-developer/app-services-sdk-go/blob/7fc1e6307584087b8fab7706f1947550e1ef71a0/.openapi/kas-fleet-manager.yaml#L1095).
Would like the name of this flag to convey it is an `accountID`, while at the same time it should have something to indicate it is related to `marketplace`. Renaming it to `MarketplaceAccountID`

wtrocki: FYI @pb82

wtrocki:
Should we select AWS first. Cloud Account ID will be per marketplace provider.
```
[wtrocki@graphapi app-services-cli (marketplace)]$ rhoas kafka create --dry-run
? Name: test
? Cloud Provider: aws
? Cloud Region: eu-west-1
? Instance type: x2
? Select a cloud account ID for the marketplace 568056954830
? Select a marketplace: aws
Kafka instance with size 2 is being created. (Size id: x2)
Selected Kafka instance size is a Technology Preview feature. Do not use it for production workloads. For more information, see https://access.redhat.com/support/offerings/techpreview
Dry run successful, arguments are valid

```

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -322,6 +328,12 @@ one = "Cloud Region:"
 [kafka.create.input.plan.message]
 one = 'Instance type:'
 
+[kafka.create.input.marketPlace.message]
+one = 'Select a marketplace:'
+
+[kafka.create.input.accountID.message]
+one = 'Select a cloud account ID for the marketplace'
```
wtrocki, line 335:
```suggestion
one = 'Select a your cloud provider ID associated with {{.Marketplace}} marketplace'
```

wtrocki:
When selection is only one item I think we can skip that option completely.

rkpattnaik780:
> When selection is only one item I think we can skip that option completely.

It occurred to me. Dropped the idea as I felt this step should be more explicit. Won't it make things bit tough if the marketplace options are selected in background and user doesn't get to know what they are?

PR #1588, opened 2022-06-05
WIP: feat(generate-config): create config with existing service account

rkpattnaik780:
`rhoas generate-config` should not create service-account.

### Verification Steps

1. Generate configuration by specifying `--type` to be `configmap`
```
rhoas generate-config --type=configmap
```
2. It should create a configmap object.
3. Unlike previous version, `rhoas generate-config` shouldn't create a service account.

### ToDo:

- Add validation for flag combinations
- Add validation to check if the service account credentials provided are valid
- Add a message for the breaking change

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


wtrocki:
Service accounts (generally any credentials cannot be passed thru flags due to bash history. We had initial design for that case in docs. Idea was to use json file provided from service account command

rkpattnaik780:
> We had initial design for that case in docs. Idea was to use json file provided from service account command. WDYT?

I was planning to go with it in the beginning, but it occurred to me that credntial can be in some other format too, for which user wll have to reconstruct it, that could be bit time taking, hence dropped the idea to have it as main method to get the credentials.

> Service accounts (generally any credentials) cannot be passed thru flags due to them being saved in plain form to bash history. 

I will agree, hence there are twoflags - `--client-secret-stdin` that will take up the secret from standard input and `--client-secret` to pass as argument for CI/CD environment. WDYT?

wtrocki:
We can use json file as input and env variables presence(user running .env file) Do we need more options?

wtrocki:
> some other format too

I would not be worried about it because:

- Json is default 
- User will generate service account specifically for this purpose
- We will add validation and description
- guide/docs will reflect it

wtrocki:
Since we designed that command many things changed:

- we now have secret in service accounts as output - enabling openshift integration there


I'm wondering if we actually approached problem right:

Do we really want to supply users service account?
Remember that we would have that already in user desired format. Maybe we just want to stop creating service account as part of generate command?

That will have numerous benefits:
- Generate can be called many times without side effects and paired with service account file
- We can better reflect openshift story:
Credentials would be in secret but urls can be configmap

wtrocki:
@rkpattnaik780 we need to put some research on this as there are different opinions how we should do it.

Mainly there is big trend in current users to disable sa creation in context generate command.
Another option is to provide flag for that use case etc. We might need to drive some proper pros and cons and see how this affects our future roadmap (helm etc.)

wtrocki:
FYI (low priority) @jackdelahunt @dimakis

pkg/cmd/generate/templates.go
```diff
@@ -15,57 +15,39 @@ var (
...
-		{{end}}
-		## Authentication Configuration
-		rhoasClientID={{.ClientID}}
-		rhoasClientSecret={{.ClientSecret}}
-		rhoasOauthTokenUrl={{.TokenURL}}
+		serviceRegistryCompatPath=` + registrycmdutil.REGISTRY_COMPAT_PATH + `{{end}}
 	`)
 
-	templateSecret = heredoc.Doc(`
+	templateConfigMap = heredoc.Doc(`
```
rkpattnaik780, line 39: @wtrocki 
This PR changes `generate-config` command to skip creating a service account. And replaces secret with configmap.

wtrocki:
Is that good to be verified?

rkpattnaik780:
> Is that good to be verified?

Yes, I will be updating help texts and examples in a while.

pkg/cmd/generate/configurations.go
```diff
@@ -15,16 +15,16 @@ const (
...
 
 var (
-	envConfig            = template.Must(template.New(envFormat).Parse(templateEnv))
-	jsonConfig           = template.Must(template.New(jsonFormat).Parse(templateJSON))
-	propertiesConfig     = template.Must(template.New(propertiesFormat).Parse(templateProperties))
-	secretTemplateConfig = template.Must(template.New(secretFormat).Parse(templateSecret))
+	envConfig             = template.Must(template.New(envFormat).Parse(templateEnv))
+	jsonConfig            = template.Must(template.New(jsonFormat).Parse(templateJSON))
+	propertiesConfig      = template.Must(template.New(propertiesFormat).Parse(templateProperties))
+	congMapTemplateConfig = template.Must(template.New(configmapFormat).Parse(templateConfigMap))
```
wtrocki, line 27: image 
;)
```suggestion
	kongMapTemplateConfig = template.Must(template.New(configmapFormat).Parse(templateConfigMap))
```

wtrocki (approved):

wtrocki:
I think we might need to add some documentation/msg. Something like:

`Generate config finished successfully. You can now create new service accounts in order to connect to the service`

pkg/core/localize/locales/en/cmd/generate_config.en.toml
```diff
@@ -43,4 +43,8 @@ one = 'file {{.FilePath}} already exists. Use --overwrite to overwrite the file,
 one='No services available to generate configurations'
 
 [generate.log.info.credentialsSaved]
-one='Configurations successfully saved to "{{.FilePath}}"'
\ No newline at end of file
+one='''
+Configurations successfully saved to "{{.FilePath}}"
+
+You can now create new service accounts or use existing ones to connect to the service(s)
```
wtrocki, line 49: As user I'm wondering how.
Can we be more specific.

Run rhoas service-account create etc.

wtrocki (approved):

jackdelahunt:
Verification steps work 👍

PR #1592, opened 2022-06-08
fix: add bypass for ams for kas-installer

wtrocki:
## Motivation

Do not call AMS when using upstream version of the platform/Self installed.

## Verification

Non interactive:

rhoas kafka create --dry-run --bypass-checks --name=test -v
Dry run successful, arguments are valid 

Interactive mode:

[wtrocki@graphapi app-services-cli (ski-ams-check ✗)]$ rhoas kafka create --dry-run --bypass-checks
❌ Interactive mode is not supported when using --skip-checks flag. Run the command in verbose mode using the -v flag to see more information

### DIFF
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -153,6 +153,7 @@ func runCreate(opts *options) error {
 		return err
 	}
 
+	var userInstanceType *accountmgmtutil.QuotaSpec
 	if !opts.bypassChecks {
 		f.Logger.Debug("Checking if terms and conditions have been accepted")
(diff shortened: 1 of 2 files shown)


rkpattnaik780:
Verification steps work fine with prod. Can I get the steps to test it against the kas-installer?

rkpattnaik780 (approved):

wtrocki:
I have verified it yesterday but this instance is now down/unavailable.

k-wall:
thank you @wtrocki - that works, but whilst I can create size x1 successfully, if I chose size `x2`, the system creates an `x1`.  I'll raise a separate issue.

PR #1593, opened 2022-06-09
fix(kafka create): completions should consider provider

rkpattnaik780:
Completion and validation of cloud account IDs should take care of the cloud_provider wherever possible.

### Verification step

1. Run `rhoas kafka create` in interactive mode.
2. Selecting marketplace should show Account IDs for that marketplace only.

### Type of change

- [ ] Documentation change
- [X] Enhancement

### DIFF
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -404,7 +404,7 @@ func promptKafkaPayload(opts *options, userQuotaType accountmgmtutil.QuotaSpec) 
 	f := opts.f
 
 	accountIDNullable := kafkamgmtclient.NullableString{}
-	cloudProviderIDNullable := kafkamgmtclient.NullableString{}
+	marketplaceProviderNullable := kafkamgmtclient.NullableString{}
 
 	validator := &kafkacmdutil.Validator{
 		Localizer:  f.Localizer,
@@ -496,7 +496,7 @@ func promptKafkaPayload(opts *options, userQuotaType accountmgmtutil.QuotaSpec) 
 		}
 
 		accountIDNullable.Set(&answers.MarketplaceAcctID)
-		cloudProviderIDNullable.Set(&answers.Marketplace)
+		marketplaceProviderNullable.Set(&answers.Marketplace)
 	}
 
 	payload := &kafkamgmtclient.KafkaRequestPayload{
@@ -504,7 +504,7 @@ func promptKafkaPayload(opts *options, userQuotaType accountmgmtutil.QuotaSpec) 
 		Region:                &answers.Region,
 		CloudProvider:         &answers.CloudProvider,
 		BillingCloudAccountId: accountIDNullable,
-		Marketplace:           cloudProviderIDNullable,
+		Marketplace:           marketplaceProviderNullable,
 	}
 	printSizeWarningIfNeeded(opts.f, answers.Size, sizes)
 	payload.SetPlan(mapAmsTypeToBackendType(&userQuotaType) + "." + answers.Size)
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -93,6 +93,10 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


rkpattnaik780:
Hi @wtrocki . Had to push some commits to #1586 . Pushing them as follow up here, also addresses some of the review comments.

pkg/cmd/kafka/create/create.go
```diff
@@ -107,22 +107,22 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
...
 			}
 
-			if opts.marketplaceAcctId != "" && !flagutil.IsValidInput(opts.marketplaceAcctId, validMarketplaceAcctIDs...) {
-				return flagutil.InvalidValueError(FlagMarketPlaceAcctID, opts.marketplaceAcctId, validMarketplaceAcctIDs...)
+			if opts.marketplace != "" && !flagutil.IsValidInput(opts.marketplace, validMarketplaces...) {
+				return flagutil.InvalidValueError(FlagMarketPlace, opts.marketplace, validMarketplaces...)
 			}
 
-			validMarketplaces, err := accountmgmtutil.GetValidMarketplaces(f.Context, f.Connection)
+			validMarketplaceAcctIDs, err := accountmgmtutil.GetValidMarketplaceAcctIDs(f.Context, f.Connection, opts.marketplace)
```
wtrocki, line 119: I just realized. Why we do separate call to AMS ( very expensive)
This will break kas-installer that doesn't have ams also it is not performant.
It will be better to extend exisiting quota object we have

WDYT?

Basically if we do --skip-checks we should not make any call to ams

wtrocki:
No worries I think we can revert this merge and bring all change again or merge this fix (but after ensuring that we do not call ams for kas-installer

wtrocki:
basically skip checks should turn off all billing extensions

pkg/cmd/kafka/create/create.go
rkpattnaik780: > It will be better to extend exisiting quota object we have

That will be better, will do it

pkg/cmd/kafka/create/create.go
```diff
@@ -93,6 +93,10 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 				}
 			}
 
+			if opts.bypassChecks && (opts.marketplace != "" || opts.marketplaceAcctId != "") {
```
wtrocki, line 96: FYI Bypass checks is basically an hack for kas-installer. We do not need to support flags like this. They can be ignored. Any billing info is not applied there

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -347,6 +347,9 @@ one = "Geographical region where the Kafka instance will be deployed"
 [kafka.create.argument.name.error.requiredWhenNonInteractive]
 one = 'name is required. Run "rhoas kafka create --name my-kafka"'
 
+[kafka.create.error.bypassChecks.marketplace]
+one = 'marketplace is not supported with "--bypass-checks" flag'
```
wtrocki, line 351: This message seems confusing. What marketplace etc.
I would recomend to not print that info

```suggestion
one = '--marketplace flag is not supported with "--bypass-checks" flag'
```

pkg/cmd/kafka/create/create.go
```diff
@@ -397,6 +403,9 @@ type promptAnswers struct {
 func promptKafkaPayload(opts *options, userQuotaType accountmgmtutil.QuotaSpec) (*kafkamgmtclient.KafkaRequestPayload, error) {
 	f := opts.f
 
+	accountIDNullable := kafkamgmtclient.NullableString{}
+	cloudProviderIDNullable := kafkamgmtclient.NullableString{}
```
wtrocki, line 407: I think this is very confusing. Cloud Provider should not be used as name as we have cloud provider as deployment target. Marketplace Provider could be better name. 

Marketplace feature is used for billing. Cloud Provider is deployment target. 
We can use AWS marketplace to pay for deployments for azure cloud provider and some region

rkpattnaik780: +1

wtrocki (approved):

PR #1596, opened 2022-06-13
fix(kafka create): reduce number of calls made to AMS API

rkpattnaik780:
Re-use `QuotaCostList` wherever possible.

Follow up for #1593 

### Type of change

- [ ] Documentation change
- [X] Enhancement


jackdelahunt:
@rkpattnaik780 I was requested to review this but are you still making changes that Wojciech suggested?

wtrocki:
Let's ignore reviewing this for now

pkg/cmd/kafka/create/create.go
```diff
@@ -512,6 +521,21 @@ func promptKafkaPayload(opts *options, userQuotaType accountmgmtutil.QuotaSpec)
 	return payload, nil
 }
 
+func promptBillingModelSelect(localizer localize.Localizer, answers *promptAnswers) error {
+
+	billingModelPrompt := &survey.Select{
+		Message: "Billing model:",
```
jackdelahunt, line 527: Should this not be a localized string

pkg/cmd/kafka/create/create.go
```diff
@@ -225,7 +234,7 @@ func runCreate(opts *options) error {
 			return f.Localizer.MustLocalizeError("kafka.create.error.noInteractiveMode")
 		}
 
-		payload, err = promptKafkaPayload(opts, *userInstanceType)
+		payload, err = promptKafkaPayload(opts, userInstanceType)
```
rkpattnaik780, line 237: The interactive mode  needs more changes, currently the user will need to specify `BillingModel` as flag before starting the interactive mode.

pkg/cmd/kafka/create/api_validators.go
```diff
@@ -119,3 +121,19 @@ func (input *ValidatorInput) ValidateSize() error {
 
 	return nil
 }
+
+// ValidateBillingModel validates if user provided a correct billing model
+func ValidateBillingModel(billingModel string) error {
```
wtrocki, line 126: Oh.. wow. I would not go that deep to validate those. If you do not specify valid billing model you would not get quota. It is great UX so thank you for this

pkg/cmd/kafka/create/create.go
```diff
@@ -212,10 +194,37 @@ func runCreate(opts *options) error {
...
+		if err != nil {
+			return err
+		}
+
+		userInstanceType, err = accountmgmtutil.GetUserSupportedInstanceType(f.Context, &constants.Kafka.Ams, conn, opts.billingModel)
 		if err != nil || userInstanceType == nil {
 			return f.Localizer.MustLocalizeError("kafka.create.error.userInstanceType.notFound")
 		}
+
+		if opts.billingModel == "marketplace" {
```
wtrocki, line 207: I would see this logic happening outside create command

pkg/cmd/kafka/create/create.go
```diff
@@ -212,10 +194,37 @@ func runCreate(opts *options) error {
...
+
+			validMarketplaces, newErr := accountmgmtutil.GetValidMarketplaces(userInstanceType)
```
wtrocki, line 209: Does this make call to AMS? I think we need to deal with all of marketplace stuff in account management package. 
Command itself should not have any logic for it - especially when we would need the same for future services as well. All we want to get is one singular quota object from accountmanagement package and our input should be billing type, marketplaceType, marketplaceAccountId

pkg/cmd/kafka/create/create.go
```diff
@@ -507,11 +527,26 @@ func promptKafkaPayload(opts *options, userQuotaType accountmgmtutil.QuotaSpec)
...
+	payload.SetPlan(mapAmsTypeToBackendType(userQuotaType) + "." + answers.Size)
 
 	return payload, nil
 }
 
+func promptBillingModelSelect(localizer localize.Localizer, answers *promptAnswers) error {
+
+	billingModelPrompt := &survey.Select{
+		Message: "Billing model:",
+		Options: kafkacmdutil.ValidBillingModels,
```
wtrocki, line 539: That will outdate fast as new billing models should be introduced. 
I think the best tradeoffs would be to return error to user from ams package containing billing models he can create when invalid is provided or billing model is missing.

pkg/cmd/kafka/kafkacmdutil/kafka_util.go
```diff
@@ -15,6 +15,13 @@ import (
 	kafkamgmtclient "github.com/redhat-developer/app-services-sdk-go/kafkamgmt/apiv1/client"
 )
 
+var (
```
wtrocki, line 18: DRY?

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -81,37 +82,38 @@ func GetUserSupportedInstanceTypes(ctx context.Context, spec *remote.AmsConfig,
 		for i := range quotaResources {
 			quotaResource := quotaResources[i]
 			if quotaResource.GetResourceName() == spec.ResourceName {
+				// nolint
 				if quotaResource.GetProduct() == spec.TrialProductQuotaID {
-					quotas = append(quotas, QuotaSpec{QuotaTrialType, 0, quotaResource.BillingModel})
-				} else if quotaResource.GetProduct() == spec.InstanceQuotaID {
+					quotas = append(quotas, QuotaSpec{QuotaTrialType, 0, quotaResource.BillingModel, nil})
+				} else if quotaResource.GetProduct() == spec.InstanceQuotaID && quotaResource.GetBillingModel() == "standard" {
```
wtrocki, line 88: Hardcoded value

pkg/shared/accountmgmtutil/api.go
```diff
@@ -4,6 +4,7 @@ package accountmgmtutil
 type QuotaType = string
 
 const (
-	QuotaTrialType    QuotaType = "trial"
-	QuotaStandardType QuotaType = "standard"
+	QuotaTrialType       QuotaType = "trial"
+	QuotaStandardType    QuotaType = "standard"
+	QuotaMarketplaceTYpe QuotaType = "marketplace"
```
wtrocki, line 9: TYpo
```suggestion
	QuotaMarketplaceType QuotaType = "marketplace"
```

wtrocki:
I think PR trying to do too much. I would suggest to:

- Not change too much code ouside accountmgmt package.
- Remove logic for picking instance types etc.
- Have new method for determining quota type:

```go
func getMeQuotaFor(billingType *string, 
       marketplaceProvider *string, 
      cloudAccountId *string) (*Quota, err) {
     // match all arguments to the 
     // quota objects and return single one
     // return error if: cannot find quota object 
     // and size of quota object is larger than 1.
}
```

- create.go changes should only contain adding 3 new flags and pass them to AMS. None of the existing logic should change.
- Additionally create.go should also pass all those values to backend. 
- I think we are fine (for now) to keep billing model outside interactive mode/do not support bash suggestions for 3 of those flags ( I see you commented them)

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -68,72 +58,57 @@ func fetchOrgQuotaCost(ctx context.Context, conn connection.Connection) (*amscli
 
 }
 
-func GetUserSupportedInstanceTypes(ctx context.Context, spec *remote.AmsConfig, conn connection.Connection) (quota []QuotaSpec, err error) {
+func FetchQuotaCost(f *factory.Factory, billingModel string, cloudAccountID string, marketplace string, spec *remote.AmsConfig) (userQuotaSpec *QuotaSpec, err error) {
```
rkpattnaik780, line 61: @wtrocki . Just wanted to get this logic verified.
Also, what should be the case if user specifying `--billing-model` as marketplace should make cloudaccountID and marketplace compulsory?

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -65,9 +67,18 @@ func FetchQuotaCost(f *factory.Factory, billingModel string, cloudAccountID stri
 		return nil, err
 	}
 
+	if billingModel == QuotaStandardType && (cloudAccountID != "" || marketplace != "") {
+		return nil, errors.New("accountID cant be provided with standard billing model")
+	}
+
+	if (cloudAccountID != "") != (marketplace != "") {
```
wtrocki, line 74: Very clever pattern!

wtrocki (approved):
I love how simple this is.
How it handles all business cases - Now and Future!
How clean and future proof this code looks like
Reusability of this code across services

Reviewed!
Verified!
Amazed!


image

wtrocki:
I have noticed that this does not work with Trial instances. I can fix that on friday without worries

PR #1598, opened 2022-06-14
Produce and consume records

jackdelahunt:
This PR adds a produce and cosume command to a kafka topic. Flags like the partition, key, limit, and timestamp are optional. This is useful for debugging purposes and currently you need another thrid-party kafka cli for this feature.

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


jackdelahunt:
Just to comment on the format of the timestamps...

| Format | Value | Worked? |
| ------------- | ------------- | ------------- |
| UTC  | 06/14/2022 @ 2:13pm  | No |
| ISO 8601  | 2022-06-14T14:13:31+00:00  | Yes |
| RFC 822  | Tue, 14 Jun 2022 14:13:31 +0000  | No |
| RFC 2822  | Tuesday, 14-Jun-22 14:13:31 UTC  | No |
| Unix timestamp  | 1655216639  | No | 


If I had to guess ISO is the default and unix could work if I was able to set the timestamp type(currently cannot do this through go-sdk)

wtrocki:
I guess we can send this as email for feedback to the admin backend so we can add better sdk docs for it

jackdelahunt:
@wtrocki basic functionality is in just for consuming and producing, tried to stick to design guidelines but I may have missed something

wtrocki:
`make lint` and `make format` required

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,141 @@
...
+
+	api, _, err := conn.API().KafkaAdmin(opts.kafkaID)
+	if err != nil {
+		return err
+	}
+
+	request := api.RecordsApi.ConsumeRecords(opts.f.Context, opts.topicName).Limit(opts.limit).Partition(opts.partition)
+	if opts.timestamp != "" {
+		// setting timestamp as "" (not set by user) is not valid
+		// not setting timestamp is handled by the request
```
wtrocki, line 96: This comment describes SDK behaviour
```suggestion
```

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,141 @@
...
+		request = request.Timestamp(opts.timestamp)
```
wtrocki, line 97: I would typically validate that timestamp and document values that are possible

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	list, _, err := request.Execute()
+	if err != nil {
+		return err
+	}
+
+	if len(list.Items) == 0 {
+		opts.f.Logger.Info(opts.f.Localizer.MustLocalize("kafka.common.log.info.noRecords"))
```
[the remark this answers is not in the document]
wtrocki: Not saying we should block request. Maybe this is good default behaviour and then we can have --wait option like in kafka?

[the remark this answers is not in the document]
jackdelahunt: Yes currently there is no way to keep consuming if there are no messages. I was thinking some `--watch` flag or something.

pkg/cmd/kafka/topic/consume/consume.go
jackdelahunt: So instead of letting the api throw errors if the format is correct parse it ourselves and then return our custom error. I guess we also add possible format options in the long description of the command.

pkg/cmd/kafka/topic/consume/consume.go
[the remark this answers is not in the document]
wtrocki: rhoas kafka create is your friend.

wtrocki:
> @wtrocki thoughts?

That is why we have offset argument set in the consume api. We can control offset :)

jackdelahunt:
Oh okay that makes sense I didn't think of this...

wtrocki:
> make lint and make format required

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,182 @@
...
+			return runCmd(opts)
+		},
+	}
+
+	flags := kafkaflagutil.NewFlagSet(cmd, f.Localizer)
+
+	flags.AddOutput(&opts.outputFormat)
+	flags.StringVar(&opts.topicName, "name", "", f.Localizer.MustLocalize("kafka.topic.common.flag.name.description"))
+	flags.Int32Var(&opts.partition, "partition", 0, f.Localizer.MustLocalize("kafka.topic.consume.flag.partition.description"))
+	flags.StringVar(&opts.timestamp, "timestamp", "", f.Localizer.MustLocalize("kafka.topic.consume.flag.timestamp.description"))
```
wtrocki, line 71: Do we need date flag as well?

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,182 @@
...
+func mapRecordsToRows(topic string, records *[]kafkainstanceclient.Record) []kafkaRow {
+
+	rows := make([]kafkaRow, len(*records))
+
+	for i := 0; i < len(*records); i++ {
+		record := &(*records)[i]
+		row := kafkaRow{
+			Topic:     topic,
+			Key:       *record.Key,
+			Value:     strings.TrimSuffix(record.Value, "\n"), // trailing new line gives weird printing of table
```
wtrocki, line 173: Value in table format would be very impractical.
Have tried it with 2 examples from here:

https://support.oneskyapp.com/hc/en-us/articles/208047697-JSON-sample-files

Even small multiline json breaks table right now

wtrocki: UI is using table because they have capabilities to view more detail on click

wtrocki: Btw. I still think table is good but not sure what to do with value and key properties that are almost always long

pkg/cmd/kafka/topic/produce/produce.go
```diff
@@ -0,0 +1,150 @@
...
+
+	record := kafkainstanceclient.Record{
+		Key:       &opts.key,
+		Value:     value,
+		Partition: &opts.partition,
+	}
+
+	_, _, err = api.RecordsApi.ProduceRecord(opts.f.Context, opts.topicName).Record(record).Execute()
+	if err != nil {
+		return err
```
wtrocki, line 132: The key is to handle some error codes from kafka admin api here to provide better info or even wrap that error into message like: failed to publish message: %#v

pkg/cmd/kafka/topic/produce/produce.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	dump.Table(opts.f.IOStreams.Out, mapRecordToRow(opts.topicName, &record))
```
wtrocki, line 135: We could probably show json format instead of table.
Table looks aweful for me with json not to mention binary payloads

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -61,6 +61,10 @@ one = 'Select Kafka instance:'
 description = 'Info message when no Kafka instances were found'
 one = 'No Kafka instances were found.'
 
+[kafka.common.log.info.noRecords]
+description = 'Info message when no records for a topic were found'
+one = 'No records were found.'
```
wtrocki, line 66: No messages available for provided message filter.

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
 [kafka.topic.list.flag.output.description]
 one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "yaml")'
 
+[kafka.topic.produce.flag.key.description]
+one = 'The key associated with the value produced. Empty if not set'
```
wtrocki, line 536:
```suggestion
one = 'The message key. Empty if not set'
```

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+
+[kafka.topic.produce.flag.key.file]
+one = 'Path to file containing value'
```
wtrocki, line 539:
```suggestion
one = 'Path to file containing message content'
```

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+
+[kafka.topic.produce.flag.partition.description]
+one = 'The partition receiving the message'
```
wtrocki, line 542:
```suggestion
one = 'The partition number for the message. Must be positive integer value that represents number of partitions for the specified topic'
```

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+
+[kafka.topic.produce.info.produceSuccess]
+description = 'Produce success message'
+one = 'Record successfully produced to "{{.Topic}}"'
```
wtrocki, line 546: We need to print offset (it is important to print that

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+
+[kafka.topic.consume.flag.partition.description]
+one = 'The partition to consume from'
```
wtrocki, line 549:
```suggestion
one = 'The partition number used for consumer. Positive integer'
```

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+
+[kafka.topic.consume.flag.timestamp.description]
+one = 'Timestamp to start consuming from'
```
wtrocki, line 552: We spend a lot of effort tracking timestamp values that backend support.
We should be as specific as possible what timestamp is

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+
+[kafka.topic.consume.flag.wait.description]
+one = 'Continuously waiting for records to consume'
```
wtrocki, line 555:
```suggestion
one = 'Waiting for records to consume'
```

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+
+[kafka.topic.consume.flag.offset.description]
+one = 'Retrieve messages with an offset equal to or greater than this'
```
wtrocki, line 558:
```suggestion
one = 'Retrieve messages within an offset equal to or greater than this'
```

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -635,6 +667,20 @@ one = '''
 $ rhoas kafka topic create --name topic-1
 '''
 
+[kafka.topic.produce.cmd.shortDescription]
+one = 'Produce to a topic'
```
wtrocki, line 671:
```suggestion
one = 'Produce a new message to a topic'
```

wtrocki:
rhoas kafka topic examples can be added as well to provide comprehensive documentation for this

wtrocki:
Very good job. Not sure about what we can do with table format. I will do more non happy path testing once build will be fixed

jackdelahunt:
Table not working is not that bad anyway, there was some issues with it anyway that I couldn't get around when looking at the library we use so not using make it easier.

pkg/cmd/kafka/topic/consume/consume.go
jackdelahunt: Jut going to set default to json output

pkg/cmd/kafka/topic/consume/consume.go
wtrocki: There are number of date formats. We need to be specific what timestamp is. Databases have sepecific format associated with timestamp etc.

More: https://en.wikipedia.org/wiki/ISO_8601

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,34 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+one = 'Timestamp to start consuming from'
+
+[kafka.topic.consume.flag.wait.description]
+one = 'Continuously waiting for records to consume'
+
+[kafka.topic.consume.flag.offset.description]
+one = 'Retrieve messages with an offset equal to or greater than this'
+
+[kafka.topic.consume.flag.limit.description]
+one = 'Max records to consume from topic'
```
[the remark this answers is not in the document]
wtrocki: limit is applied to single request always. Then with --wait we need to document that server is being pulled with arguments specified by user (including limit)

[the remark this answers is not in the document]
jackdelahunt: Even if we were to clearly document this, it still seems kinda weird no? Limit to me always implies only ever give me back this amount of messages no matter what! As a user, I don't really think about individual requests.

pkg/cmd/kafka/topic/consume/consume.go
jackdelahunt: Oh okay I understand, just something to specify the format used

pkg/core/localize/locales/en/cmd/kafka.en.toml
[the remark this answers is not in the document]
wtrocki: What is the suggestion here? Ignore limit when --wait is supplied?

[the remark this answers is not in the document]
jackdelahunt: No don't ignore it all together just accumulate all messages received up until that limit without any consideration of each request.

For example, I have 5 messages on my topic but I set my limit to 6. I should consume all 5 on this topic and then when another message is produced consume that giving a total of 6. Any other message produced is not consumed and consuming is stopped.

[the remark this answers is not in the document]
wtrocki: Would that represent some valid end user use case?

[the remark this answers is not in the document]
jackdelahunt: I think so, don't see a big difference between this and a normal limit flag without waiting. 

Also using a limit for each request as you suggest just ignores the limit, if my limit is 1 but there are more than that on the topic then it will just keep consuming but just as separate requests. 

Limit in this scenario just means keep consuming just slower because we are doing more requests.

wtrocki:
As in the design doc. Let's drop our normal output format and provide custom output formats for that command specicically. 

Many people would probably want to see key value format/json etc.

jackdelahunt:
> custom output formats 

Do you mean outputting the messages in a way that only is done for this command? Not using dump

wtrocki:
> Do you mean outputting the messages in a way that only is done for this command? Not using dump

exactly. It can be some --output-format flag or anything you like to not confuse with general cli -o flag

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/redhat-developer/app-services-cli/pkg/core/localize"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/connection"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/contextutil"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/factory"
+	kafkainstanceclient "github.com/redhat-developer/app-services-sdk-go/kafkainstance/apiv1internal/client"
+	"github.com/spf13/cobra"
+)
+
+const (
+	DefaultOffset    = 0
```
wtrocki, line 23: -1 would be better. 0 is an valid offset number.
If -1 is there we should not pass filter value for offset

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,215 @@
...
+	limit        int32
+	offset       int64
+	wait         bool
+	outputFormat string
+
+	f *factory.Factory
+}
+
+// row is the details of a record produced needed to print to a table
+type kafkaRow struct {
```
wtrocki, line 44: This seems not used?

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		return err
+	}
+
+	api, _, err := conn.API().KafkaAdmin(opts.kafkaID)
+	if err != nil {
+		return err
+	}
+
+	if opts.wait {
```
wtrocki, line 113: I would handle wait in separate method

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,215 @@
...
+			opts.offset = DefaultOffset
+		}
+
+		if opts.from == DefaultTimestamp {
+			// get current time in ISO 8601
+			opts.from = time.Now().Format(time.RFC3339)
+		}
+
+		max_offset := opts.offset
+		for true {
```
wtrocki, line 131: I think we look for equivalent of thread in java or set timeout.
Check go routines. Endless loop does work but we need logic to wait 
to not get rate limited

jackdelahunt: I checked go routines and they seems to not be worth it, it always ended up being that we just wait on the separate thread to end. Adding some kind of sleep that is less than the rate limit is what I'll do.

pkg/cmd/kafka/topic/consume/consume.go
jackdelahunt: Okay, I thought having the -1 would fall into what we said about defaults being specific non-valid values so we know they are not set.

pkg/cmd/kafka/topic/consume/consume.go
jackdelahunt: The kafkaRow is used in the mapRecords to row function. [Here](https://github.com/redhat-developer/app-services-cli/blob/produce-consume/pkg/cmd/kafka/topic/consume/consume.go#L203-L209)

wtrocki: Sorry for not being clear. What is the purpose of that function. Wy we mapping response to another object that looks very similar to api object?

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/spf13/cobra"
+)
+
+const (
+	DefaultOffset    = 0
+	DefaultLimit     = 20
+	DefaultTimestamp = ""
+)
+
+var outputFormatTypes = []string{dump.JSONFormat, dump.YAMLFormat}
```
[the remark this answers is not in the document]
jackdelahunt: Also isn't the key value pair format pretty much just yaml anyway? I am not sure what would make out key value output be different from that. Unless I am thinking of it in a different sense.

pkg/cmd/kafka/topic/consume/consume.go
jackdelahunt: Added sleep in for loop to not hit rate limit

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,65 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+[kafka.topic.produce.info.produceSuccess]
+description = 'Produce success message'
+one = 'Record successfully produced to "{{.Topic}}" with offset {{.Offset}}'
+
+[kafka.topic.consume.cmd.shortDescription]
+one = 'Consume messages from a topic'
+
+[kafka.topic.consume.cmd.longDescription]
+one = '''
+Consume messages from a given topic, by default all messages on the topic will be consumed and printed in the foramat chosen. You can add filters to these message like a starting offset or a time that the messages must of been produced by.
```
wtrocki, line 553: Lets add line breaks

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,65 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+
+# Consume from a topic and wait for messages produced since command was ran
+$ rhoas kafka topic consume --name=topic-1 --wait
+'''
+
+[kafka.topic.consume.flag.partition.description]
+one = 'The partition number used for consumer. Positive integer'
+
+[kafka.topic.consume.flag.from.description]
+one = 'Only messages with a timestamp after this time will be consumed'
```
wtrocki, line 571: Missing info about date format. What is the "timestamp" :D

pkg/core/localize/locales/en/cmd/kafka.en.toml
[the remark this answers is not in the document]
wtrocki: Let's close comments that were addressed.

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -528,6 +532,65 @@ one = 'Format in which to display the Kafka topic (choose from: "json", "yml", "
...
+one = 'Consume messages from a topic'
+
+[kafka.topic.consume.cmd.longDescription]
+one = '''
+Consume messages from a given topic, by default all messages on the topic will be consumed and printed in the foramat chosen. You can add filters to these message like a starting offset or a time that the messages must of been produced by.
+
+Adding the --wait flag will wait for messages to be produced starting from when the command was ran and will ignore any limit or offset given.
+'''
+
+[kafka.topic.consume.cmd.example]
```
wtrocki, line 558: Not complete list of examples IMHO. Those are key for everything

pkg/cmd/kafka/topic/consume/consume.go
wtrocki: Filtering values are problem. Because if they are not specified they need to be not included.
I do not think logic with offset 0 passed all the time will be valid :)

Offset could be "" then we can parse number from that string. WDYT?

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,244 @@
...
+
+	flags := kafkaflagutil.NewFlagSet(cmd, f.Localizer)
+
+	flags.StringVar(&opts.topicName, "name", "", f.Localizer.MustLocalize("kafka.topic.common.flag.name.description"))
+	flags.Int32Var(&opts.partition, "partition", 0, f.Localizer.MustLocalize("kafka.topic.consume.flag.partition.description"))
+	flags.StringVar(&opts.from, "from", DefaultTimestamp, f.Localizer.MustLocalize("kafka.topic.consume.flag.from.description"))
+	flags.BoolVar(&opts.wait, "wait", false, f.Localizer.MustLocalize("kafka.topic.consume.flag.wait.description"))
+	flags.Int64Var(&opts.offset, "offset", DefaultOffset, f.Localizer.MustLocalize("kafka.topic.consume.flag.offset.description"))
+	flags.Int32Var(&opts.limit, "limit", DefaultLimit, f.Localizer.MustLocalize("kafka.topic.consume.flag.limit.description"))
+	flags.StringVar(&opts.outputFormat, "format", "json", f.Localizer.MustLocalize("kafka.topic.consume.flag.format.description"))
```
wtrocki, line 89: We should pass possible values to description of the command

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,244 @@
...
+			} else {
+				opts.f.Logger.Info(fmt.Sprintf("Key: %v\nValue: %v\n", row.Key, row.Value))
+			}
+		} else {
+			_ = dump.Formatted(opts.f.IOStreams.Out, format, *row)
+		}
+	}
+}
+
+func mapRecordsToRows(topic string, records *[]kafkainstanceclient.Record) []kafkaRow {
```
wtrocki, line 226: Question: Why we need this method? Any reason why we do not want to use API structure?
When API structure changes we will need to make manual change here which is not ideal

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,244 @@
...
+	format := opts.outputFormat
+	if format == dump.EmptyFormat {
+		format = dump.JSONFormat
+	}
+
+	for i := 0; i < len(recordsAsRows); i++ {
+		row := &recordsAsRows[i]
+		if format == FormatKeyValue {
+			if row.Key == "" {
+				opts.f.Logger.Info(row.Value)
```
wtrocki, line 216: We will miss message delimiter. Maybe `Value: ..` be better here to see when mesage finished.

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,244 @@
...
+			} else {
+				opts.f.Logger.Info(fmt.Sprintf("Key: %v\nValue: %v\n", row.Key, row.Value))
```
wtrocki, line 218: Replace `Value` (internal field) to `Message`

wtrocki:
Left basic comments. I could not verify commands:

[wtrocki@graphapi app-services-cli (produce-consume)]$ rhoas kafka topic produce --name=missing-topic

Missing info that we are waiting for user input. Missing validation on non existent topic.

[wtrocki@graphapi app-services-cli (produce-consume)]$ rhoas kafka topic produce --name=test --file connector.json 
❌ 404 Not Found. Run the command in verbose mode using the -v flag to see more information

No API error handling present.

pkg/cmd/kafka/topic/consume/consume.go
jackdelahunt: The Record has members we may not want or need like timestamp type. This was the standard way in other parts of cli to just extract the values we want to see. Personally I don't mind either way.

wtrocki:
JSON format is unusable. Value needs to be escaped thus it is hard to read. 
I do not think we can do anything else than using key, value format as default instead. 

[wtrocki@graphapi app-services-cli (produce-consume)]$ rhoas kafka topic consume --name=test                         
{
  "topic": "test",
  "key": "",
  "value": "{\n  \"channel\": \"stable\",\n  \"connector\": {\n    \"kafka_topic\": \"lb-cos\",\n    \"log_show_all\": true\n  },\n  \"connector_type_id\": \"log_sink_0.1\",\n  \"id\": \"c97gajsotd37bufhphu0\",\n  \"kafka\": {\n    \"id\": \"c52ajhf2ek06ri44ife0\",\n    \"url\": \"openbridge-c--ajhf-ek--ri--ifea.bf2.kafka.rhcloud.com:443\"\n  },\n  \"kind\": \"Connector\",\n  \"modified_at\": \"2022-04-07T15:37:51.612514Z\",\n  \"name\": \"cos-wtrocki-logs\",\n  \"namespace_id\": \"c9f8i3njogcbrlqqlt90\",\n  \"owner\": \"wtrocki_kafka_supporting\",\n  \"schema_registry\": {\n    \"id\": \"\",\n    \"url\": \"\"\n  },\n  \"service_account\": {\n    \"client_id\": \"srvc-acct-3de38d77-90ff-4480-9a0f-8a8230ccff23\",\n    \"client_secret\": \"srvc-acct-3de38d77-90ff-4480-9a0f-8a8230ccff23\"\n  }\n}",
  "partition": 0,
  "offset": 0
}

wtrocki:
Key value format seems to be missformatted. Some messages with key will get formatted differently than others:
```
{
  "schema_registry": {
    "id": "",
    "url": ""
  },
  "service_account": {
    "client_id": "srvc-acct-3de38d77-90ff-4480-9a0f-8a8230ccff23",
    "client_secret": "srvc-acct-3de38d77-90ff-4480-9a0f-8a8230ccff23"
  }
}

```
Key: test
Value: {
  "channel": "stable",
  "connector": {
    "kafka_topic": "lb-cos",
    "log_show_all": true
  },
  "connector_type_id": "log_sink_0.1",
  "id": "c97gajsotd37bufhphu0",
  "kafka": {
    "id": "c52ajhf2ek06ri44ife0",
    "url": "openbridge-c--ajhf-ek--ri--ifea.bf2.kafka.rhcloud.com:443"
  },
  "kind": "Connector",
  "modified_at": "2022-04-07T15:37:51.612514Z",
  "name": "cos-wtrocki-logs",
  "namespace_id": "c9f8i3njogcbrlqqlt90",
  "owner": "wtrocki_kafka_supporting",
  "schema_registry": {
    "id": "",

```


I think that we would need always print Key, Message or Message: format to visibly show end of the messages.

**Comment by wtrocki** [comment]
Some valid cases return 400 for me:
```
[wtrocki@graphapi app-services-cli (produce-consume)]$ rhoas kafka topic consume --name=test --partition=1 --offset=1
❌ 400 Bad Request. Run the command in verbose mode using the -v flag to see more information
```

File: pkg/cmd/kafka/topic/consume/consume.go (hunk 28 above)
Comment (wtrocki) [reply pkg/cmd/kafka/topic/consume/consume.go:226]: We have key value output for that (default) and json should be just output of api.

File: pkg/cmd/kafka/topic/consume/consume.go [hunk 31]
```diff
@@ -86,7 +86,7 @@ func NewConsumeTopicCommand(f *factory.Factory) *cobra.Command {
 	flags.Int32Var(&opts.partition, "partition", 0, f.Localizer.MustLocalize("kafka.topic.consume.flag.partition.description"))
 	flags.StringVar(&opts.from, "from", DefaultTimestamp, f.Localizer.MustLocalize("kafka.topic.consume.flag.from.description"))
```
wtrocki, line 87: Wondering if we need 2 flags for Unix Epoch and Actual Date. 
I would prefer 2 flags for simplicity of documentation and validation

jackdelahunt:
> Some valid cases return 400 for me

Was not able to replicate this so will wait for feedback from others to find the issue

pkg/cmd/kafka/topic/consume/consume.go
wtrocki: I think flag modifying another flag can be hard to work on and confusing. 

I would suggest:

--from-time (ISO 8601 date - short format)
--from-timestamp (epoch/unix timestamp)



I would love to save this dicussion somewhere so we can recall this when building CLI guidelines.

wtrocki: Btw. That is my opinion and this is opinionated part so do not feel that you need to do it this way. It is more about my personal preference to keep docs clean.

wtrocki:
> Was not able to replicate this so will wait for feedback from others to find the issue

Will check now

wtrocki:
It works now (throws validation error)

wtrocki:
Used https://awesomeopensource.com/project/jdorfman/awesome-json-datasets 
Tested --wait use case. Tested normal produce consume. This looks very good to be shared. It is just matter of figuring out dates and examples

> Message: Binary or non-UTF-8 encoded data cannot be displayed

Nice!

wtrocki (approved):
We need more examples/docs and we good to merge

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -0,0 +1,297 @@
...
+// NewComsumeTopicCommand creates a new command for producing to a kafka topic.
+func NewConsumeTopicCommand(f *factory.Factory) *cobra.Command {
+	opts := &options{
+		f: f,
+	}
+
+	cmd := &cobra.Command{
+		Use:     "consume",
+		Short:   f.Localizer.MustLocalize("kafka.topic.consume.cmd.shortDescription"),
+		Long:    f.Localizer.MustLocalize("kafka.topic.consume.cmd.longDescription"),
```
wtrocki, line 66: Lets add Hidden: True to hide commands for now

jackdelahunt:
@wtrocki added more examples and made commands hidden, will I squash and merge this now?

wtrocki:
Every new change - new review and verification. doing it now

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -582,6 +582,12 @@ $ rhoas kafka topic consume --name=topic-1
 
 # Consume from a topic and wait for messages produced since command was ran
 $ rhoas kafka topic consume --name=topic-1 --wait
+
+# Consume from a topic starting from a certain time using the default ISO time format
+$ rhoas kafka topic consume --name=topic-1 --wait --from=2022-06-17T07:05:34+00:00Z
+
+# Consume from a topic starting from a certain time using unix time format
+$ rhoas kafka topic consume --name=topic-1 --wait --unix-time --from=812762
```
wtrocki, line 590: missing jq example etc. all examples are for wait - missing non wait examples

wtrocki:
We need more examples for typical use cases:
- getting message at offset
- jq some part of the message stream only
- getting info about current offset
- adding same examples to topic root

PR #1601, opened 2022-06-16
fix: use kafka admin url from control plane

wtrocki:
Hardcoded url that was used were matching only production. For any other env url can have different format. We should always use control plane kafka response in all tooling.


rkpattnaik780 (approved):
Verified

PR #1602, opened 2022-06-16
fix: update error message to reflect proper flag name

wtrocki:


wtrocki:
Minor change. Flag is defined here: https://github.com/redhat-developer/app-services-cli/blob/flag_name/pkg/core/cmdutil/flagutil/flagset.go#L94:L94

PR #1606, opened 2022-06-23
feat: connectors docs and enablement

wtrocki:


wtrocki:
Merged to enable connectors documentation generation

PR #1607, opened 2022-06-27
Better documentation for topic commands

jackdelahunt:
Adds more examples for consume command and adds produce and consume examples to topic root command.

### Type of change
- [x] Documentation change
- [ ] Other (please specify)


wtrocki (approved):
I will try to improve this in separate PR

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -603,6 +603,9 @@ one = '''
 # Consume from a topic
 $ rhoas kafka topic consume --name=topic-1
 
+# Consume from topic and output yaml format
```
wtrocki, line 606: Actually this is problematic right now as command is hidden. I will remove it

PR #1610, opened 2022-06-27
Api change

wtrocki:
Proposal for flags for timestamps being dedicated.

### DIFF
--- a/pkg/core/localize/locales/en/cmd/kafka.en.toml
+++ b/pkg/core/localize/locales/en/cmd/kafka.en.toml
@@ -606,14 +606,14 @@ $ rhoas kafka topic consume --name=topic-1
 # Consume from topic and output yaml format  
 $ rhoas kafka topic consume --name=topic-1 --format=yaml
 
-# Consume from a topic and wait for messages produced since command was ran
+# Consume from a topic continously by pooling for new messages
 $ rhoas kafka topic consume --name=topic-1 --wait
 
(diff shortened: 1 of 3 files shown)


pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -606,14 +606,14 @@ $ rhoas kafka topic consume --name=topic-1
 # Consume from topic and output yaml format  
 $ rhoas kafka topic consume --name=topic-1 --format=yaml
 
-# Consume from a topic and wait for messages produced since command was ran
+# Consume from a topic continously by pooling for new messages
 $ rhoas kafka topic consume --name=topic-1 --wait
 
 # Consume from a topic starting from a certain time using the default ISO time format
-$ rhoas kafka topic consume --name=topic-1 --wait --from=2022-06-17T07:05:34+00:00Z
+$ rhoas kafka topic consume --name=topic-1 --from=2022-06-17T07:05:34+00:00Z
```
wtrocki, line 613: I think we could use user timezone from the operating system instead (we do not need zulu)
also there could be more simpler way to reflect date by skipping timezone element.

```suggestion
$ rhoas kafka topic consume --name=topic-1 --from=2022-06-17T07:05:34
```

wtrocki:
@jackdelahunt this was only example. Closing this as you would probably look to make PR yourself.

PR #1613, opened 2022-06-27
Connector updates

wtrocki:
Minor updates needed after SDK update. 
Removed --eval flag as all namespaces are eval right now


wtrocki:
I'm merging this to fix connectors CLI

PR #1617, opened 2022-06-28
fix: billing model fixes

wtrocki:
Adds billing model support


wtrocki:
Sadly entire logic doesn't seem to make much sense. Both markeplace and standard are broken right now. I think I would revert all those changes now to save our faces and do prolonged release and we can start adding this again next week

PR #1626, opened 2022-07-01
chore: reverting invalid changes breaking kafka creation view

wtrocki:
As mentioned this reverts multiple commits that have been happening for marketplace. 
We have flagged that marketplace work would be delayed but need CLI working with mas-sso migration asap


wtrocki:
Created standard/marketplace/and trial instances successfully. There are still some minor bugs and inconvenience's but basically this would enable use to create all types easily.

PR #1628, opened 2022-07-06
Connector in context

jackdelahunt:
This pr adds a connector to the context used in the cli. With this a new command is added `rhoas connector use` which is also aliased in `rhoas context set-connector`. 

### Verification Steps
Start by deleting the current context and creating a new one, also and create a connector instance in the UI.

1. Run `rhoas context set-connector` to get a list of your connectors and set one
2. Run `rhoas context status` to see the status of that connector.
3. Run `cat /home/$USER/.config/rhoas/contexts.json` to see saved context file

### Type of change


pkg/cmd/connector/connector.go
```diff
@@ -30,6 +31,7 @@ func NewConnectorsCommand(f *factory.Factory) *cobra.Command {
 	cmd.AddCommand(
 		cluster.NewConnectorClusterCommand(f),
 		namespace.NewNameSpaceCommand(f),
+		use.NewUseCommand(f),
```
wtrocki, line 34: Use commands have been planned to be deprecated.
We have the same commands in context set-kafka etc.
It will be good to add set-connector command there.

My take is that we need 2 commands doing the same thing for now

wtrocki:
Very good change/work so far. Thank you for sharing this early

pkg/cmd/connector/connector.go
jackdelahunt: Yep I did see this and it was kinda weird ill add both to make it consistant until the change is made

pkg/shared/connectorutil/error.go
```diff
@@ -0,0 +1,15 @@
+package connectorutil
+
+import (
+	"fmt"
+)
+
+func NotFoundByIDError(id string) error {
+	NotFoundByIDErr := fmt.Errorf(`Coonector instance with ID "%v" not found`, id)
```
wtrocki, line 8: image
```suggestion
	NotFoundByIDErr := fmt.Errorf(`Cooonector instance with ID "%v" not found`, id)
```

No i18n?

pkg/shared/contextutil/util.go
```diff
@@ -109,3 +110,37 @@ func GetRegistryForServiceConfig(currCtx *servicecontext.ServiceConfig, f *facto
...
+	currCtx, err := GetCurrentContext(svcContext, f.Localizer)
+	if err != nil {
+		return nil, err
+	}
+
+	return GetConnectorForServiceConfig(currCtx, f)
+}
+
+func GetConnectorForServiceConfig(currCtx *servicecontext.ServiceConfig, f *factory.Factory) (*connectormgmtclient.Connector, error) {
+	conn, err := f.Connection(connection.DefaultConfigSkipMasAuth)
```
wtrocki, line 131: That would create 2 connections all the time - despite someone not even using connectors. 
I think we might pass connection to this method (entire context object) instead.

pkg/shared/contextutil/util.go
```diff
@@ -109,3 +110,37 @@ func GetRegistryForServiceConfig(currCtx *servicecontext.ServiceConfig, f *facto
...
+	if err != nil {
+		return nil, err
+	}
+
+	if currCtx.ConnectorID == "" {
```
wtrocki, line 136: Alternatively - we can simply move this check above so connection is not created when we do not have id.
(Same for srs and kafka)

wtrocki (approved):
Approved (although some small bits to be fixed)

wtrocki:
```
---------------------------------------------------------------------------------
  ID:                     cb3ech0a86j5tmua8p00
  Name:                   test
  Status:                 ready
  Bootstrap URL:          test-cb-ech-a--j-tmua-p-a.bf2.kafka-stage.rhcloud.com:443

  Connector
  ---------------------------------
  ID:              cb3fhqpl04t2ndvmir7g
  Name:            http
  Status:          ready

  Service Registry
  ------------------------------------------------------------------------------------------------------------
  ID:                    726c2db7-4868-492b-95cc-9ff51e1fafa2
  Name:                  wtrocki
  Status:                ready
  Registry URL:          https://bu98.serviceregistry-stage.rhcloud.com/t/726c2db7-4868-492b-95cc-9ff51e1fafa2
[wtrocki@graphapi app-services-cli (connector-in-context)]$ 
```

I would put connector on the bottom. Add connector_type_id and maybe kafka reference as well :)

pkg/cmd/connector/use/use.go
```diff
@@ -101,6 +89,18 @@ func runUse(opts *options) error {
 		return err
 	}
 
+	if opts.interactive {
+		// run the use command interactively
+		err = runInteractivePrompt(opts, &conn)
+		if err != nil {
+			return err
+		}
+		// no Kafka was connector, exit program
```
wtrocki, line 98: > "no Kafka was connector" 

but Connector, no Kafka can be, Yes..

image

wtrocki (approved):

PR #1629, opened 2022-07-07
chore: skip sso refresh

wtrocki:
We have been doing sso refresh as safety net to have fresh token for users in case we revert back to sso. This have started causing issues as some claims have been migrated. 

This PR stops refreshes


rkpattnaik780 (approved):
Verified the steps with service-registry commands.

PR #1631, opened 2022-07-07
fix: remove mas-sso tokens on login

wtrocki:
## Verification

1. Login to production
2. Try out registry instance artifact list
3. All should work
4. Login to stage
5. Run artifact list
6. All should work


rkpattnaik780 (approved):
Verified the flow in description.

PR #1636, opened 2022-07-12
feat(kafka create): add billing model logic

rkpattnaik780:
This PR adds `--billing-model` flag to `rhoas kafka create` command.
The CLI should be able to create instances with standard, marketplace and trial quotas. 

### Verification Steps

1. For testing against marketplace  (use _kafka_service account).

`rhoas kafka create --name test-instance` should throw error saying multiple quota types are available and user needs to specify billing model.
`rhoas kafka create --name test-instance --billing-model standard` should create a standard instance
`rhoas kafka create --name test-instance --billing-model standard` should create a marketplace instance

`rhoas kafka create` should start the creation in interactive mode and prompt for billing model.

User should be able to pick marketplace provider and market place account if multiple options are available, else the only account gets selected by default.

2. When only trial instances are available (use _kafka_devexp account)

Specifying the billing model should fail with message "only trial instances available".

`rhoas kafka create --name test-instance` should create a trial instance.

Interactive mode should not prompt billing model.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/create/completions.go
+++ b/pkg/cmd/kafka/create/completions.go
@@ -1,10 +1,7 @@
 package create
 
 import (
-	"github.com/redhat-developer/app-services-cli/pkg/shared/accountmgmtutil"
-	"github.com/redhat-developer/app-services-cli/pkg/shared/connection"
 	"github.com/redhat-developer/app-services-cli/pkg/shared/factory"
-	"github.com/redhat-developer/app-services-cli/pkg/shared/remote"
 	"github.com/spf13/cobra"
 )
 
@@ -25,49 +22,4 @@ func GetCloudProviderRegionCompletionValues(f *factory.Factory, providerID strin
 	validRegions, _ = GetEnabledCloudRegionIDs(f, providerID, nil)
 
 	return validRegions, cobra.ShellCompDirectiveNoSpace
-}
-
-// GetKafkaSizeCompletionValues returns a list of valid kafka sizes for the specified region and ams instance types
-func GetKafkaSizeCompletionValues(f *factory.Factory, providerID string, regionId string) (validRegions []string, directive cobra.ShellCompDirective) {
-	directive = cobra.ShellCompDirectiveNoSpace
-
-	// We need both values to provide a valid list of sizes
-	if providerID == "" || regionId == "" {
-		return nil, directive
-	}
-
-	err, constants := remote.GetRemoteServiceConstants(f.Context, f.Logger)
-	if err != nil {
-		return nil, directive
-	}
-
-	conn, err := f.Connection(connection.DefaultConfigSkipMasAuth)
-	if err != nil {
-		return nil, directive
-	}
-
-	userInstanceType, _ := accountmgmtutil.GetUserSupportedInstanceType(f.Context, &constants.Kafka.Ams, conn)
-
-	// Not including quota in this request as it takes very long time to list quota for all regions in suggestion mode
-	validRegions, _ = FetchValidKafkaSizesLabels(f, providerID, regionId, *userInstanceType)
-
-	return validRegions, cobra.ShellCompDirectiveNoSpace
-}
-
-// GetMarketplaceAcctIdCompletionValues returns a list of valid marketplace account IDs for the organization
-func GetMarketplaceAcctIdCompletionValues(f *factory.Factory) (validMarketplaceAcctIDs []string, directive cobra.ShellCompDirective) {
-	directive = cobra.ShellCompDirectiveNoSpace
-
-	validMarketplaceAcctIDs, _ = accountmgmtutil.GetValidMarketplaceAcctIDs(f.Context, f.Connection, "")
-
-	return validMarketplaceAcctIDs, directive
-}
-
-// GetMarketplaceCompletionValues returns a list of valid marketplaces for the organization
-func GetMarketplaceCompletionValues(f *factory.Factory) (validMarketplaces []string, directive cobra.ShellCompDirective) {
-	directive = cobra.ShellCompDirectiveNoSpace
-
-	validMarketplaces, _ = accountmgmtutil.GetValidMarketplaces(f.Context, f.Connection)
-
-	return validMarketplaces, directive
 }
--- a/pkg/cmd/kafka/create/completions.go
+++ b/pkg/cmd/kafka/create/completions.go
@@ -40,7 +40,7 @@ func GetKafkaSizeCompletionValues(f *factory.Factory, providerID string, regionI
 		return nil, directive
 	}
 
-	orgQuota, err := accountmgmtutil.GetUserSupportedInstanceType(f, &constants.Kafka.Ams)
+	orgQuota, err := accountmgmtutil.GetOrgQuotas(f, &constants.Kafka.Ams)
 	if err != nil {
 		return nil, directive
 	}
--- a/pkg/cmd/kafka/create/completions.go
+++ b/pkg/cmd/kafka/create/completions.go
@@ -52,3 +52,45 @@ func GetKafkaSizeCompletionValues(f *factory.Factory, providerID string, regionI
 
 	return validSizes, cobra.ShellCompDirectiveNoSpace
 }
+
+func GetMarketplaceCompletionValues(f *factory.Factory) (validSizes []string, directive cobra.ShellCompDirective) {
+
+	directive = cobra.ShellCompDirectiveNoSpace
+
+	err, constants := remote.GetRemoteServiceConstants(f.Context, f.Logger)
+	if err != nil {
+		return nil, directive
+	}
+
+	orgQuota, err := accountmgmtutil.GetOrgQuotas(f, &constants.Kafka.Ams)
+	if err != nil {
+		return nil, directive
+	}
+
+	validMarketPlaces := FetchValidMarketplaces(orgQuota.MarketplaceQuotas)
+
+	return validMarketPlaces, cobra.ShellCompDirectiveNoSpace
+}
+
+func GetMarketplaceAccountCompletionValues(f *factory.Factory, marketplace string) (validMarketplaceAcctIDs []string, directive cobra.ShellCompDirective) {
+
+	directive = cobra.ShellCompDirectiveNoSpace
+
+	if marketplace == "" {
+		return validMarketplaceAcctIDs, directive
+	}
+
+	err, constants := remote.GetRemoteServiceConstants(f.Context, f.Logger)
+	if err != nil {
+		return nil, directive
+	}
+
+	orgQuota, err := accountmgmtutil.GetOrgQuotas(f, &constants.Kafka.Ams)
+	if err != nil {
+		return nil, directive
+	}
+
+	validMarketplaceAcctIDs = FetchValidMarketplaceAccounts(orgQuota.MarketplaceQuotas, marketplace)
+
+	return validMarketplaceAcctIDs, cobra.ShellCompDirectiveNoSpace
+}
--- a/pkg/cmd/kafka/create/completions.go
+++ b/pkg/cmd/kafka/create/completions.go
@@ -94,3 +94,22 @@ func GetMarketplaceAccountCompletionValues(f *factory.Factory, marketplace strin
 
 	return validMarketplaceAcctIDs, cobra.ShellCompDirectiveNoSpace
 }
+
+func GetBillingModelCompletionValues(f *factory.Factory) (availableBillingModels []string, directive cobra.ShellCompDirective) {
+
+	directive = cobra.ShellCompDirectiveNoSpace
+
+	err, constants := remote.GetRemoteServiceConstants(f.Context, f.Logger)
+	if err != nil {
+		return nil, directive
+	}
+
+	orgQuota, err := accountmgmtutil.GetOrgQuotas(f, &constants.Kafka.Ams)
+	if err != nil {
+		return nil, directive
+	}
+
+	availableBillingModels = FetchSupportedBillingModels(orgQuota)
+
+	return availableBillingModels, directive
+}
--- a/pkg/cmd/kafka/create/completions.go
+++ b/pkg/cmd/kafka/create/completions.go
@@ -1,7 +1,9 @@
 package create
 
 import (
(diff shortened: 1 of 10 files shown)


pkg/cmd/kafka/create/create.go
```diff
@@ -93,10 +95,14 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 				}
 			}
 
-			if opts.bypassChecks && (opts.marketplace != "" || opts.marketplaceAcctId != "") {
+			if opts.bypassChecks && (opts.marketplace != "" || opts.marketplaceAcctId != "" || opts.billingModel != "") {
 				return f.Localizer.MustLocalizeError("kafka.create.error.bypassChecks.marketplace")
```
wtrocki, line 99: Description needs billing model flag

pkg/cmd/kafka/create/create.go
```diff
@@ -93,10 +95,14 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
...
 			}
 
+			if opts.billingModel == "standard" && (opts.marketplaceAcctId != "" || opts.marketplace != "") {
```
wtrocki, line 102: I would suggest to not list any billing model in create.go. Also not hardcoding it's value.

pkg/cmd/kafka/create/create.go
```diff
@@ -157,18 +144,6 @@ func NewCreateCommand(f *factory.Factory) *cobra.Command {
 		return GetCloudProviderRegionCompletionValues(f, opts.provider)
 	})
 
-	_ = cmd.RegisterFlagCompletionFunc(FlagSize, func(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
```
wtrocki, line 160: Why removing those?

pkg/cmd/kafka/create/create.go
```diff
@@ -212,9 +187,15 @@ func runCreate(opts *options) error {
 			return nil
 		}
 
-		userInstanceType, err = accountmgmtutil.GetUserSupportedInstanceType(f.Context, &constants.Kafka.Ams, conn)
+		marketplaceInfo := accountmgmtutil.MarketplaceInfo{
```
wtrocki, line 190: Can be transformed as inline structure passed to method.

pkg/cmd/kafka/create/create.go
```diff
@@ -212,9 +187,15 @@ func runCreate(opts *options) error {
...
+			Provider:       opts.marketplace,
+			BillingModel:   opts.billingModel,
+			CloudAccountID: opts.marketplaceAcctId,
+		}
+
+		userInstanceType, err = accountmgmtutil.GetUserSupportedInstanceType(f, &constants.Kafka.Ams, marketplaceInfo)
 		if err != nil || userInstanceType == nil {
-			return f.Localizer.MustLocalizeError("kafka.create.error.userInstanceType.notFound")
+			return err // || f.Localizer.MustLocalizeError("kafka.create.error.userInstanceType.notFound")
```
wtrocki, line 198: userInstanceType == nil will print `Error: nil` to user

pkg/cmd/kafka/create/create.go
```diff
@@ -282,6 +264,11 @@ func runCreate(opts *options) error {
 
 	}
 
+	if userInstanceType.BillingModel == "standard" || userInstanceType.BillingModel == "marketplace" {
```
wtrocki, line 267: There are no other billing models and if they would be in future we should still pass them. I think this is redundant.

pkg/cmd/kafka/create/create.go
```diff
@@ -474,31 +461,6 @@ func promptKafkaPayload(opts *options, userQuotaType accountmgmtutil.QuotaSpec)
 		}
 	}
 
-	marketplaces, err := accountmgmtutil.GetValidMarketplaces(f.Context, f.Connection)
```
wtrocki, line 477: Why removing this?

wtrocki: If we want to ask user to pick marketplace details this should happen before size and feed back into size (this is current bug in relased version of the CLI). I do not see how interactive mode would work with marketplace right now. Would we fail/still require users to use flags for non interactive selects?

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -36,20 +36,22 @@ func CheckTermsAccepted(ctx context.Context, spec *remote.AmsConfig, conn connec
...
-
-	amsType := PickInstanceType(userInstanceTypes)
+type MarketplaceInfo struct {
+	BillingModel   string
+	Provider       string
+	CloudAccountID string
+}
 
-	return amsType, nil
+type OrgQuotas struct {
```
wtrocki, line 51: While this is good I would myself keep all Quota spec as array and use filter flags (billingModel) to filter them out. 

There is big decision that we can take:

- Have trial as quota object (it will be always there) thus if logic is `len(quotas) == 1 { return quotas[0] } 
- Ignore trial and do not return quota  for it (it will be more accurate from business logic point of view , but it will require rewrite of the create.go logic.

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -68,144 +70,189 @@ func fetchOrgQuotaCost(ctx context.Context, conn connection.Connection) (*amscli
...
-		if quotas[i].BillingModel == alwaysWinsBillingModel {
-			betterQuotasMap[quotas[i].Name] = &quotas[i]
-		} else if betterQuotasMap[quotas[i].Name] == nil {
-			betterQuotasMap[quotas[i].Name] = &quotas[i]
+	return SelectQuotaForUser(f, availableOrgQuotas, marketplace)
+}
 
+func SelectQuotaForUser(f *factory.Factory, orgQuota *OrgQuotas, marketplaceInfo MarketplaceInfo) (*QuotaSpec, error) {
+	if len(orgQuota.StandardQuotas) == 0 && len(orgQuota.MarketplaceQuotas) == 0 {
+		if len(orgQuota.TrialQuotas) == 0 {
```
wtrocki, line 112: Trial would be always present. Also doesn't need to checked in quota selection. 

 Decision we need to take here is: 
 - do we want to fail when user specify rubbish billing model or accoutID that doesn't exist? 
 - Can we build simple state machine for selection filtering that will be future proof?
 I have build diagram to showcase how this would look like:
 
Screenshot 2022-07-12 at 11 31 51

pkg/cmd/kafka/create/completions.go
```diff
@@ -41,33 +40,15 @@ func GetKafkaSizeCompletionValues(f *factory.Factory, providerID string, regionI
...
 
 	// Not including quota in this request as it takes very long time to list quota for all regions in suggestion mode
-	validRegions, _ = FetchValidKafkaSizesLabels(f, providerID, regionId, *userInstanceType)
+	validSizes, _ = FetchValidKafkaSizesLabels(f, providerID, regionId, *userInstanceType)
 
-	return validRegions, cobra.ShellCompDirectiveNoSpace
-}
-
-// GetMarketplaceAcctIdCompletionValues returns a list of valid marketplace account IDs for the organization
-func GetMarketplaceAcctIdCompletionValues(f *factory.Factory) (validMarketplaceAcctIDs []string, directive cobra.ShellCompDirective) {
```
wtrocki, line 58: Reason to remove this?

pkg/cmd/kafka/create/create.go
```diff
@@ -438,7 +443,74 @@ func promptKafkaPayload(opts *options, userQuotaType accountmgmtutil.QuotaSpec)
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if answers.BillingModel == accountmgmtutil.QuotaMarketplaceType {
+		validMarketPlaces := FetchValidMarketplaces(orgQuota.MarketplaceQuotas)
+		if len(validMarketPlaces) == 1 {
+			answers.Marketplace = validMarketPlaces[0]
+		} else {
```
wtrocki, line 470: How this would work with len(validMarketPlaces) == 0?

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -68,144 +70,180 @@ func fetchOrgQuotaCost(ctx context.Context, conn connection.Connection) (*amscli
...
-	return unique(marketplaceAcctIDs), err
+	if len(matchedQuotas) == 0 {
+		return nil, f.Localizer.MustLocalizeError("kafka.create.quota.error.cloudAccountNotFound")
+	}
+
+	return &matchedQuotas[0], nil
 }
 
-func GetValidMarketplaces(ctx context.Context, connectionFunc factory.ConnectionFunc) (marketplaces []string, err error) {
+func pickCloudAccount(f *factory.Factory, cloudAccounts *[]amsclient.CloudAccount, market MarketplaceInfo) (*[]amsclient.CloudAccount, error) {
```
wtrocki, line 218: Should this method return single object instead of array. It does `pick` cloud account right?

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -68,144 +70,180 @@ func fetchOrgQuotaCost(ctx context.Context, conn connection.Connection) (*amscli
...
 	}
 
-	for _, quota := range quotaCostGet.GetItems() {
-		if len(quota.GetCloudAccounts()) > 0 {
-			for _, cloudAccount := range quota.GetCloudAccounts() {
-				marketplaces = append(marketplaces, cloudAccount.GetCloudProviderId())
+	var matchedAccounts []amsclient.CloudAccount
+
+	for _, cloudAccount := range *cloudAccounts {
+		if market.Provider != "" || market.CloudAccountID != "" {
```
wtrocki, line 231: Those variables are set outside for so we might do them once outside loop?

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -68,144 +70,180 @@ func fetchOrgQuotaCost(ctx context.Context, conn connection.Connection) (*amscli
...
+			if *cloudAccount.CloudProviderId == market.Provider || *cloudAccount.CloudAccountId == market.CloudAccountID {
```
wtrocki, line 232: Question I have:
 Do you think we should support people specifying only market.Provider?

Maybe we can just:
```suggestion
			if *cloudAccount.CloudProviderId == market.Provider && *cloudAccount.CloudAccountId == market.CloudAccountID {
```

To keep this convoluted logic little bit simpler?

rkpattnaik780: > Do you think we should support people specifying only market.Provider?

I will agree not allowing marketplace provider alone can make the logic bit simpler. I will be removing it.

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -68,144 +70,180 @@ func fetchOrgQuotaCost(ctx context.Context, conn connection.Connection) (*amscli
...
+	matchedQuotas := []QuotaSpec{}
+
+	for _, quota := range marketplaceQuotas {
+		cloudAccounts := *quota.CloudAccounts
+		for _, cloudAccount := range cloudAccounts {
+			if marketplace.Provider != "" && marketplace.CloudAccountID != "" {
+				if *cloudAccount.CloudProviderId == marketplace.Provider && *cloudAccount.CloudAccountId == marketplace.CloudAccountID {
+					matchedQuotas = append(matchedQuotas, quota)
+				}
+			} else if marketplace.Provider != "" || marketplace.CloudAccountID != "" {
```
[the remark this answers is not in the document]
rkpattnaik780: This was to handle the case of providing only `provider` or `account ID`.

pkg/shared/accountmgmtutil/ams.go
rkpattnaik780: Used array to make it easy to use with quotapsec object.

jackdelahunt:
Ran into an issues.

When only trial instances were available setting the billing model (standard or marketplace) did not throw an error as described in verification steps.

rkpattnaik780:
> Ran into an issue.
> 
> When only trial instances were available setting the billing model (standard or marketplace) did not throw an error as described in the verification steps.

Hi @jackdelahunt 
Not able to reproduce it. Can you paste the command in verbose mode.

jackdelahunt:
Here -> https://gist.github.com/jackdelahunt/81e45bfcbee10b38f9bdaab3e35800ca

rkpattnaik780:
> Here -> https://gist.github.com/jackdelahunt/81e45bfcbee10b38f9bdaab3e35800ca

Thanks for the catch, PR has been updated.

wtrocki:
I have tried to do "Duncan Test" as start: 
```
[wtrocki@graphapi app-services-cli (pr/rkpattnaik780/1636 ✗)]$ rhoas kafka create                   
? Name: fds
? Cloud Provider: aws
? Billing Model: marketplace
? Cloud Region: us-east-1
Kafka instance with size 1 is being created. (Size id: x1)
❌ 400 Bad Request. Run the command in verbose mode using the -v flag to see more information
```

It has:
- Wrong plan selected (developer plan/trial was on where I selected proper billing model (marketplace)
- Sizes provided were wrong only x1 (my plan had support up to x2)

Error detail:
{"reason":"Unable to detect instance type in plan provided: 'developer.x1'","operation_id":"cb82cemo4k31f4mqqo9g","id":"21","kind":"Error","href":"/api/kafkas_mgmt/v1/errors/21","code":"KAFKAS-MGMT-21"}

wtrocki:
@rkpattnaik780 let's do more comprehensive testing with control plane team and mock. Once we have confidence in this we can release.

pkg/cmd/kafka/create/create.go
```diff
@@ -245,7 +245,9 @@ func runCreate(opts *options) error {
 			return err
 		}
 
-		f.Logger.Debug(fmt.Sprintf("Selected quota object: %#v", userQuota))
+		userQuotaJSON, _ := json.MarshalIndent(userQuota, "", "  ")
+		f.Logger.Debug("Selected Quota object:")
+		f.Logger.Debug(string(userQuotaJSON))
```
wtrocki, line 250: Let's create helper for logging that type of stuff. It takes a lot of space.

PR #1638, opened 2022-07-12
chore: feedback produce consume

wtrocki:


jackdelahunt (approved):
:+1:  approved

jackdelahunt:
One comment I have is maybe when you do not set the partition to consume from we show the partition where it is from.

image

pkg/cmd/kafka/topic/consume/consume.go
```diff
@@ -284,6 +286,10 @@ func outputRecords(opts *options, records *kafkainstanceclient.RecordList) {
 			} else {
 				opts.f.Logger.Info(fmt.Sprintf("Key: %v\nMessage: %v", row.Key, row.Value))
 			}
+			opts.f.Logger.Info(fmt.Sprintf("Offset: %v", row.Offset))
```
jackdelahunt, line 289: Double print of offset

jackdelahunt: log message should say partition below

PR #1640, opened 2022-07-13
fix: add env var config

wtrocki:
This fixes problems with offline tokens/refresh tokens when having empty config. I think we should use empty config flag for testing etc. 

## Verification

```
RHOASCONFIG="./config.json" rhoas login

RHOASCONFIG="./config.json" rhoas login --token

RHOASCONFIG="./config.json" rhoas login
```

Notice token login not printing user info but classic login does
Check how config changes when testing


rkpattnaik780 (approved):
LGTM!

PR #1646, opened 2022-07-14
fix: remove redundant refresh token spec

wtrocki:


wtrocki:
This does cleanup changes. Verification and review would be typically checking if commands work :D
image

PR #1651, opened 2022-07-15
fix: print no result when list is nil and using default format

jackdelahunt:
Currently when running `connector list --page=2` when there are no valid connectors for that query and you are using an empty format then no result is returned (expected is to print no result message). 

### Verification Steps

1. Have no connectors runnning
2. On main run `rhoas connector list --page=2` and you get no output
3. On this branch run the same and you get info about empty result

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/connector/connector.go
+++ b/pkg/cmd/connector/connector.go
@@ -38,11 +38,11 @@ func NewConnectorsCommand(f *factory.Factory) *cobra.Command {
 		start.NewStartCommand(f),
 		stop.NewStopCommand(f),
 		connector_type.NewTypeCommand(f),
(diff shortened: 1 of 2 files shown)


jackdelahunt:
@wtrocki 

While this pr does fix the issue I described it is just getting around an issue I found either with SDK or API not sure yet.

When trying to list when page == 0 if there are no connectors then the items are nil but also the size in response is 0. Before we just checked size which I thought would be expected.

But when page == 1 the items in response is still nil but the size is 1 for some reason. 

I also think the items should just be an empty list but that isn't what is causing this bug.

PR #1653, opened 2022-07-16
fix: cache connection to skip refreshes

wtrocki:
Reverting https://github.com/redhat-developer/app-services-cli/pull/1624/files
This change was to ensure that we mas-sso tokens refresh have been added. Now since mas-sso tokens are gone we should cache our regular sso tokens across session.

This is still not perfect but due to very large amount of connections created in various helpers (kafka create creates -7 connections on it's own - that is 7 sso refreshes per command). Some better refactor for connection objects is needed.
For now we can cache that object as there is no chance that tokens will outdate during single command run (5minutes)


wtrocki:
Merging without review as this is reverting previous PR etc.

We can verify this change by running any command - WE should only see single token refresh happening from now (instead of 7 refreshes before)

PR #1654, opened 2022-07-18
fix(kafka create): chose standard if no marketplace info is provided

rkpattnaik780:
Earlier the Kafka create failed with error if multiple quota types were present and billing model was not specified by the user. This PR changes logic to pick `standard` quota if no marketplace detail is provided.

### Verification Steps

1. Run the mock server.
2. Run `rhoas kafka create` command in verbose mode.
```
rhoas kafka create --name test-instance -v --dry-run
```
3. It should log the following as the selected quota:
```
Selected Quota object:
{
  "Name": "standard",
  "Quota": 299,
  "BillingModel": "standard",
  "CloudAccounts": null
}
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/shared/accountmgmtutil/ams.go
+++ b/pkg/shared/accountmgmtutil/ams.go
@@ -146,7 +146,7 @@ func SelectQuotaForUser(f *factory.Factory, orgQuota *OrgQuotas, marketplaceInfo
 
 	if len(orgQuota.StandardQuotas) > 0 && len(orgQuota.MarketplaceQuotas) > 0 {
 
-		if marketplaceInfo.BillingModel == QuotaStandardType {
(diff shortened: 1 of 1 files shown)


rkpattnaik780:
> Do we have that in the Design/Epic brief. I do not think UI does this - it always requires user to pick:
> 
> > CC @jgiardino

Made this change after seeing the demo for "marketplace e2e tests" in sprint review.

rkpattnaik780:
> > Made this change after seeing the demo for "marketplace e2e tests" in sprint review.
> 
> But that was mainly for backend backwards compatibility. Can you summary why we would want that for CLI?

I got the idea that this is an expected flow. You are right, it will become more confusing, closing this PR.

PR #1670, opened 2022-07-21
fix: improved poor search in connector type list

jackdelahunt:
Should now have a better search, still doesn't use ID .

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/connector/connector_type/list/list.go
+++ b/pkg/cmd/connector/connector_type/list/list.go
@@ -82,7 +82,7 @@ func runUpdateCommand(opts *options) error {
 	request = request.Size(strconv.Itoa(opts.limit))
 
 	if opts.search != DefaultSearch {
-		query := fmt.Sprintf("name like %[1]s or label like %[1]s or description like %[1]s", opts.search)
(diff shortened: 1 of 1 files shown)


pkg/cmd/connector/connector_type/list/list.go
```diff
@@ -82,7 +82,7 @@ func runUpdateCommand(opts *options) error {
 	request = request.Size(strconv.Itoa(opts.limit))
 
 	if opts.search != DefaultSearch {
-		query := fmt.Sprintf("name like %[1]s or label like %[1]s or description like %[1]s", opts.search)
+		query := fmt.Sprintf("name like %[2]s%[1]s%[2]s or label like %[2]s%[1]s%[2]s or description like %[2]s%[1]s%[2]s", opts.search, "%")
```
wtrocki, line 85: Label doesn't exist :D 
We are missing id :D

jackdelahunt: Removing label, but id does not work either adding it makes every request not work

PR #1676, opened 2022-07-25
fix(generate-config): update success message and default path for env

rkpattnaik780:
This is a follow up to #1588 .

The PR updates the success message printed after configs are generated.
`rhoas generate-config --type env` should save configurations to `rhoas.env`

### Verification Steps
1. Run the following command:

```
./rhoas generate-config --type env
```

2. The command should save configurations in `rhoas.env`.

```
Configurations successfully saved to "/home/rkpattnaik780/coding/redhat/mk/cli/rhoas.env"

You can now use existing service accounts or create new to connect to the services.
To create new service account, run this command:

  $ rhoas service-account create
```

### Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/generate/configurations.go
+++ b/pkg/cmd/generate/configurations.go
@@ -57,7 +57,7 @@ func WriteConfig(opts *options, config *configValues) (string, error) {
 func getDefaultPath(configType string) (filePath string) {
 	switch configType {
 	case envFormat:
-		filePath = ".env"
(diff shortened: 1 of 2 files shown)


wtrocki:
I would personally go with config.env but rhoas.env is ok.

Issue #1678, opened 2022-07-25
Whoami doesn't return email after login


jackdelahunt:
## Description

- Run `rhoas login`
- Run `rhoas whoami`
   - I get the output "Token has no username" 
   - Cli stil works as normal however
- Run `rhoas whoami` again and I get the correct output
- Logging in and out through cli does not change issue

PR #1679, opened 2022-07-25
fix: refresh token before reading config

wtrocki:
Fixes: https://github.com/redhat-developer/app-services-cli/issues/1678


wtrocki:
I could not replicate your issue but this was 100% the problem. We were reading config before refreshing it.

PR #1693, opened 2022-08-02
fix: minor connectors improvements

wtrocki:
Some minor suggestions from the call we had


jackdelahunt:
Just need to remove some changes for the json library as they are no longer needed

wtrocki:
Outdated

PR #1702, opened 2022-08-03
Connector edit

jackdelahunt:
Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/connector/update/update.go
+++ b/pkg/cmd/connector/update/update.go
@@ -111,6 +111,23 @@ func runUpdate(opts *options) error {
 	}
 
 	if apiErr := connectorerror.GetAPIError(err); apiErr != nil {
+
+		if apiErr.GetCode() == connectorerror.ERROR_11 {
+			return opts.f.Localizer.MustLocalizeError("connector.update.error.authTokenInvalid")
+		}
+
+		if apiErr.GetCode() == connectorerror.ERROR_7 {
+			return opts.f.Localizer.MustLocalizeError("connector.update.error.noMatchingResource")
+		}
+
+		if apiErr.GetCode() == connectorerror.ERROR_25 {
+			return opts.f.Localizer.MustLocalizeError("connector.update.error.doesNotExistAnymore")
+		}
+
+		if apiErr.GetCode() == connectorerror.ERROR_9 {
+			return opts.f.Localizer.MustLocalizeError("connector.update.error.unexpectedError")
+		}
+
 		return opts.f.Localizer.MustLocalizeError("connector.type.update.error.other", localize.NewEntry("Error", apiErr.GetReason()))
 	}
 
--- a/pkg/cmd/connector/update/update.go
+++ b/pkg/cmd/connector/update/update.go
@@ -2,30 +2,36 @@ package update
 
 import (
 	"encoding/json"
-	"io"
+	"fmt"
+	"github.com/jackdelahunt/survey-json-schema/pkg/surveyjson"
+	"github.com/redhat-developer/app-services-cli/pkg/core/cmdutil/flagutil"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/connection"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/connection/api"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/connectorutil"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/contextutil"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/kafkautil"
+	"github.com/spf13/cobra"
 	"os"
 
-	"github.com/redhat-developer/app-services-cli/pkg/cmd/connector/connectorcmdutil"
-	"github.com/redhat-developer/app-services-cli/pkg/cmd/registry/artifact/util"
-	"github.com/redhat-developer/app-services-cli/pkg/core/cmdutil/flagutil"
 	"github.com/redhat-developer/app-services-cli/pkg/core/ioutil/dump"
+	"github.com/redhat-developer/app-services-cli/pkg/core/localize"
+	connectorerror "github.com/redhat-developer/app-services-sdk-go/connectormgmt/apiv1/error"
+	"gopkg.in/AlecAivazis/survey.v1"
 
-	"github.com/redhat-developer/app-services-cli/pkg/shared/connection"
 	"github.com/redhat-developer/app-services-cli/pkg/shared/factory"
-	"github.com/spf13/cobra"
+	connectormgmtclient "github.com/redhat-developer/app-services-sdk-go/connectormgmt/apiv1/client"
 )
 
 type options struct {
-	file string
+	namespaceID string
+	kafkaID     string
+	name        string
 
 	outputFormat string
 	f            *factory.Factory
-	id           string
 }
 
-// NewUpdateCommand creates a new command to update a connector
 func NewUpdateCommand(f *factory.Factory) *cobra.Command {
-
 	opts := &options{
 		f: f,
 	}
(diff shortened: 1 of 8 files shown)


jackdelahunt:
still not done, maybe if you can give it some testing @wtrocki

pkg/cmd/connector/update/update.go
```diff
@@ -2,118 +2,177 @@ package update
...
+	flags.StringVar(&opts.name, "name", "", f.Localizer.MustLocalize("connector.flag.name.description"))
 	flags.AddOutput(&opts.outputFormat)
 
 	return cmd
-
 }
 
-func runUpdateCommand(opts *options) error {
-	f := opts.f
+func runCreate(opts *options) error {
```
wtrocki, line 57:
```suggestion
func runUreate(opts *options) error {
```

pkg/cmd/connector/update/update.go
```diff
@@ -2,118 +2,177 @@ package update
...
 
-	byteValue, err := ioutil.ReadAll(specifiedFile)
-	if err != nil {
-		return err
-	}
-	var connector map[string]interface{}
-	err = json.Unmarshal(byteValue, &connector)
+	var connector *connectormgmtclient.Connector
+
+	connector, err = connectorutil.InteractiveSelect(conn, opts.f)
```
wtrocki, line 68: Interactive mode without option to use flag. 
Interactive mode is not handled by checking if terminal session enables einteractive mode

pkg/cmd/connector/update/update.go
```diff
@@ -2,118 +2,177 @@ package update
...
-			return runUpdateCommand(opts)
+			return runCreate(opts)
 		},
 	}
-
-	flags := connectorcmdutil.NewFlagSet(cmd, f)
-	flags.StringVar(&opts.file, "file", "", f.Localizer.MustLocalize("connector.file.flag.description"))
-	flags.AddConnectorID(&opts.id)
+	flags := flagutil.NewFlagSet(cmd, f.Localizer)
+	flags.StringVar(&opts.name, "name", "", f.Localizer.MustLocalize("connector.flag.name.description"))
```
wtrocki, line 51: Would that flag still make sense?

wtrocki: Btw. Name is not guaranteed to be unique

pkg/cmd/connector/update/update.go
```diff
@@ -2,118 +2,177 @@ package update
...
 		return err
 	}
 
-	f.Logger.Info(f.Localizer.MustLocalize("connector.update.info.success"))
+	opts.f.Logger.Info(opts.f.Localizer.MustLocalize("connector.create.info.success"))
 
 	return nil
 }
+
+func setValuesInInteractiveMode(connector *connectormgmtclient.Connector) (map[string]interface{}, error) {
```
wtrocki, line 102: Curious if we can actually reuse the same codebase across create and update?

pkg/cmd/connector/update/update.go
```diff
@@ -2,118 +2,177 @@ package update
...
 				return flagutil.InvalidValueError("output", opts.outputFormat, validOutputFormats...)
 			}
 
-			return runUpdateCommand(opts)
+			return runCreate(opts)
 		},
 	}
-
-	flags := connectorcmdutil.NewFlagSet(cmd, f)
-	flags.StringVar(&opts.file, "file", "", f.Localizer.MustLocalize("connector.file.flag.description"))
```
wtrocki, line 52: Would we allow users to edit connector spec file and run update on this?

pkg/cmd/connector/update/update.go
```diff
@@ -35,85 +41,178 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 		Short:   f.Localizer.MustLocalize("connector.update.cmd.shortDescription"),
 		Long:    f.Localizer.MustLocalize("connector.update.cmd.longDescription"),
 		Example: f.Localizer.MustLocalize("connector.update.cmd.example"),
-		Hidden:  true,
```
wtrocki, line 38: We should follow create being hidden. It will be odd to have only update visible :D

pkg/cmd/connector/update/update.go
```diff
@@ -35,85 +41,178 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
...
 		},
 	}
-
-	flags := connectorcmdutil.NewFlagSet(cmd, f)
-	flags.StringVar(&opts.file, "file", "", f.Localizer.MustLocalize("connector.file.flag.description"))
-	flags.AddConnectorID(&opts.id)
+	flags := flagutil.NewFlagSet(cmd, f.Localizer)
+	flags.StringVar(&opts.name, "name", "", f.Localizer.MustLocalize("connector.flag.name.description"))
+	flags.StringVar(&opts.namespaceID, "namespace-id", "", f.Localizer.MustLocalize("connector.flag.name.description"))
+	flags.StringVar(&opts.kafkaID, "kafka-id", "", f.Localizer.MustLocalize("connector.flag.name.description"))
```
wtrocki, line 57:
```suggestion
	flags.StringVar(&opts.kafkaID, "kafka-id", "", f.Localizer.MustLocalize("connector.flag.kafka-id.description"))
```

pkg/cmd/connector/update/update.go
```diff
@@ -35,85 +41,178 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
...
+			return runUpdate(opts)
 		},
 	}
-
-	flags := connectorcmdutil.NewFlagSet(cmd, f)
-	flags.StringVar(&opts.file, "file", "", f.Localizer.MustLocalize("connector.file.flag.description"))
-	flags.AddConnectorID(&opts.id)
+	flags := flagutil.NewFlagSet(cmd, f.Localizer)
+	flags.StringVar(&opts.name, "name", "", f.Localizer.MustLocalize("connector.flag.name.description"))
+	flags.StringVar(&opts.namespaceID, "namespace-id", "", f.Localizer.MustLocalize("connector.flag.name.description"))
```
wtrocki, line 56:
```suggestion
	flags.StringVar(&opts.namespaceID, "namespace-id", "", f.Localizer.MustLocalize("connector.flag.name" ".description"))
```

pkg/cmd/connector/update/update.go
jackdelahunt: My idea of the connector update was that it is always pointing to a connector that is created already and you want to update it. Updating the json build file I think is a different idea that shouldn't be covered by the connector update command

pkg/cmd/connector/update/update.go
```diff
@@ -35,85 +41,184 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
...
+	if err = dump.Formatted(opts.f.IOStreams.Out, opts.outputFormat, updated); err != nil {
+		return err
+	}
+
+	opts.f.Logger.Info(opts.f.Localizer.MustLocalize("connector.update.info.success"))
+
+	return nil
+}
+
+func setValuesInInteractiveMode(connector *connectormgmtclient.Connector, opts *options, conn *connection.Connection) (map[string]interface{}, error) {
```
wtrocki, line 114: Is that duplicated between create and update?

pkg/cmd/connector/update/update.go
```diff
@@ -35,85 +41,184 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
...
+
+	if opts.namespaceID == "" {
+		namespaceId, err := askForValue("Namespace ID", connector.GetNamespaceId(), opts)
+		if err != nil {
+			return nil, err
+		}
+		opts.namespaceID = namespaceId
+	}
+
+	patch["namespace_id"] = opts.namespaceID
```
wtrocki, line 156: Probably better to have it inside the if statement.

pkg/cmd/connector/update/update.go
```diff
@@ -35,85 +41,184 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
...
+	if err2 != nil {
+		connector, _ = connectorutil.InteractiveSelect(conn, opts.f)
+	}
+
+	patch, err3 := setValuesInInteractiveMode(connector, opts, &conn)
+	if err3 != nil {
+		return err3
+	}
+
+	connectorSpecification, err4 := askForConnectorDetails(&api, connector, opts)
```
wtrocki, line 83: If I want to update just one element like connector name would i need to fill the details?

pkg/cmd/connector/update/update.go
```diff
@@ -35,85 +41,184 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
...
+	opts.f.Logger.Info(opts.f.Localizer.MustLocalize("connector.update.info.success"))
+
+	return nil
+}
+
+func setValuesInInteractiveMode(connector *connectormgmtclient.Connector, opts *options, conn *connection.Connection) (map[string]interface{}, error) {
+
+	patch := make(map[string]interface{})
+
+	if opts.name == "" {
```
wtrocki, line 118: This feels like dead code. Any *valid* connector cannot have name as empty.
Also logic is strange as if I would like to change name I would not be asked for it etc.

pkg/cmd/connector/update/update.go
wtrocki: > My idea of the connector update was that it is always pointing to a connector that is created already and you want to update it. 

That would be also my understanding of updating item. 

>  Updating the json build file I think is a different idea t

I did not implied that would be new file. I meant that users can stream connector changes and patch it:

- Sending patch request 
- Sending whole file that user edited. etc.

all of that can be done by streaming (reading standard input) in update command. 

This would be alternative way to handle updates etc.

wtrocki:
I see numerous ways of handing updates but this PR doesn't seem to handle well some very basic use cases. 

As we discussed update command can be have exactly as update command - asking users every question. However this would not be great experience. 

We can let users to send patch requests for things they want to change - however this would imply some challenging interface/validation for patch changes.

Another option which I have suggested would be to fetch connector file and give user ability to edit it with OpenEditor - that seems trivial and would give users the best experience IMHO.

wtrocki:
See: https://github.com/redhat-developer/app-services-cli/pull/1713

pkg/cmd/connector/update/update.go
```diff
@@ -35,85 +35,131 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
...
+		connector.SetNamespaceId(opts.namespaceID)
+		connectorChanged = true
+	}
+	if opts.kafkaID != "" {
+		connector.Kafka.SetId(opts.kafkaID)
+		connectorChanged = true
+	}
+
+	if !connectorChanged {
+		err = runEditor(connector)
```
wtrocki, line 89: I forgot to put some message to user informing them that now we will start editor for them and they 
can change editor by changing $EDITOR env. Like EDITOR="Code -w"

wtrocki:
This looks really good. It will enable us to release that command instantly.

wtrocki (approved):

wtrocki:
Verified

PR #1708, opened 2022-08-08
fix: remove mas-auth  url

wtrocki:
URL is no longer valid and have been included in +2 releases as deprecated. We can remove it


jackdelahunt:
LGTM :+1:

PR #1715, opened 2022-08-10
feat: added info about changing editor in connector update

jackdelahunt:
Adds prompt and help to change editor in connector update

### Type of change

- [x] Documentation change
- [ ] Other (please specify)


pkg/cmd/connector/update/update.go
```diff
@@ -55,6 +55,7 @@ func NewUpdateCommand(f *factory.Factory) *cobra.Command {
 	return cmd
 }
 
+// nolint:funlen
```
wtrocki, line 58: We tend to never add that

wtrocki (approved):

PR #1721, opened 2022-08-18
chore(service-account): use error codes from SDK

rkpattnaik780:
Error handling for service-account operations.

### Type of change

- [ ] Documentation change
- [X] SDK Integration


pkg/cmd/serviceaccount/svcaccountcmdutil/errorutils/error_sdk.go
```diff
@@ -0,0 +1,34 @@
+package errorutils
```
rkpattnaik780, line 1: @wtrocki should we move this file to SDK, lik the other SDKs have?

pkg/cmd/serviceaccount/create/create.go
```diff
@@ -140,8 +141,13 @@ func runCreate(opts *options) error {
 		defer httpRes.Body.Close()
 	}
 
-	if err != nil {
-		return err
+	if apiErr := errorutils.GetAPIError(err); apiErr != nil {
+		switch apiErr.GetError() {
+		case "service_account_limit_exceeded":
```
wtrocki, line 146: I think we have Enums defined in SDK for those

pkg/cmd/serviceaccount/svcaccountcmdutil/errorutils/error_sdk.go
wtrocki: +1 on adding this to SDK. Thank you

pkg/cmd/serviceaccount/create/create.go
rkpattnaik780: I tried that first. The enums are defined in a different manner in the openAPI [file](https://github.com/redhat-developer/app-services-sdk-go/blob/main/.openapi/service-accounts.yaml#L346), more like accepted options.

pkg/cmd/serviceaccount/svcaccountcmdutil/errorutils/error_sdk.go
rkpattnaik780: Raised PR https://github.com/redhat-developer/app-services-sdk-go/pull/555.

pkg/core/localize/locales/en/cmd/serviceaccount.en.toml
```diff
@@ -43,6 +43,9 @@ one = 'internal server error'
 one = 'you are forbidden to {{.Operation}} this service account'
 other = 'you are forbidden to {{.Operation}} these service accounts'
 
+[serviceAccount.common.error.limitExceeded]
+one = 'account can not have more than 50 service accounts'
```
wtrocki, line 47: We cannot put any backend driven business logic into message. 
50 account limit is just config on backend that can change etc.

```suggestion
one = 'account can not have more than 50 service accounts'
```

wtrocki (approved):

wtrocki (approved):
As discussed. This is good to be merged

PR #1724, opened 2022-08-19
fix: can now update kafka when only owner

jackdelahunt:
Running `rhoas kafka update reauthentication false` was not allowed as you needed to be a org admin.

This pr allows for kafka owners to update the kafka reauthentication option and also skips prompt they are not allowed to change as they are not org admins.

### Verification Steps

1. Build main and run `rhoas kafka update reauthentication=false` when not being org admin, this will fail
2. Build this branch run `rhoas kafka update reauthentication false` and should work fine 
3. run rhoas `rhoas kafka update` and should only be able to change reauthentication an not owner
4. Do the same as org admin and you can change owner

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


jackdelahunt:
go.sum changes were from `make format` and gh actions fails if not included.

wtrocki:
> run rhoas rhoas kafka update and should only be able to change reauthentication an not owner

So backend doesnt allow to change owner? One cannot give their kafka to someone else?

jackdelahunt:
> So backend doesn't allow to change owner? One cannot give their kafka to someone else?

Not sure about the actual action itself, but in the implementation that we have we try and select the owner from a list of other people and to get this list it requires to to be org admin.

pkg/cmd/kafka/update/update.go
```diff
@@ -213,9 +209,14 @@ func run(opts *options) error {
...
-	if err != nil {
-		return err
+
+	if opts.userIsOrgAdmin {
+		opts.owner, err = selectOwnerInteractive(opts.Context, opts)
+		if err != nil {
+			return err
+		}
+	} else {
+		opts.owner = kafkaInstance.GetOwner()
```
wtrocki, line 219: I think we should askForNewOwnerName - as opposed to select for admin.
Current logic does assign the same owner.

jackdelahunt: as in enter the name directly as the select is for admins only?

wtrocki: Only admins can fetch users but anyone can put username that they know and should be able to give their kafka away.

wtrocki (approved):

PR #1716, opened 2022-08-10
chore: integrate service-account SDK

rkpattnaik780:
Integrate `serviceaccount-sdk` with the RHOAS CLI

### Verification Steps
1. Build the CLI from the changes and try out the service account commands.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation change
- [X] Dependency update

### ToDo

- [X] Format the time in `service-account list`
- ~~[ ] Distinguish between field "name" and "description" in `ServiceAccountData` object~~

PR #1732, opened 2022-08-23
fix: revert service accounts integration

wtrocki:
Revert changes introduced in #1716.


wtrocki:
@rkpattnaik780 We are looking to release connectors. Service accounts integration should stay on the feature branch as we would need to give more heads up to people as this will introduce some breaking changes.

jackdelahunt (approved):
LGTM :+1:

PR #1742, opened 2022-08-31
fix: make request command public

wtrocki:
When chatting with @dimakis we seen that request command is not public. adding this as PR instead of issue as change is trivial


jackdelahunt (approved):
Fix CI and then LGTM 👍

PR #1743, opened 2022-09-02
feat(service-account): add support for service account SDK

rkpattnaik780:
Added a flag to login to control output format of service-account commands.

### Verification Steps

1. Run `rhoas service-account list -o json`. It should print output with control plane SDK (an object)
2. Login with the following flag:
```
 ./rhoas login --enable-auth-v2 
```
3.  Run `rhoas service-account list -o json` again. It should print output with service account SDK (an array)

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/login/login.go
+++ b/pkg/cmd/login/login.go
@@ -57,6 +57,8 @@ type options struct {
 	insecureSkipTLSVerify bool
 	printURL              bool
 	offlineToken          string
+	// Temporary workaround to be removed
+	enableAuthV2 bool
 }
 
 // NewLoginCmd gets the command that's log the user in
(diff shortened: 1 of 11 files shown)


pkg/cmd/login/login.go
```diff
@@ -96,6 +98,7 @@ func NewLoginCmd(f *factory.Factory) *cobra.Command {
 	cmd.Flags().BoolVar(&opts.printURL, "print-sso-url", false, opts.localizer.MustLocalize("login.flag.printSsoUrl"))
 	cmd.Flags().StringArrayVar(&opts.scopes, "scope", kcconnection.DefaultScopes, opts.localizer.MustLocalize("login.flag.scope"))
 	cmd.Flags().StringVarP(&opts.offlineToken, "token", "t", "", opts.localizer.MustLocalize("login.flag.token", localize.NewEntry("OfflineTokenURL", build.OfflineTokenURL)))
+	cmd.Flags().BoolVar(&opts.enableAuthV2, "enable-auth-v2", false, opts.localizer.MustLocalize("login.flag.enableAuthV2"))
```
rkpattnaik780, line 101: Since the output formats are controlled for only two commands, will it be better to put it at individual command level and not at config level?

wtrocki: Makes sense.

wtrocki (approved):

wtrocki:
WIth that change in place we can easily merge and release those changes next week.

PR #1745, opened 2022-09-14
feat(service-registry): added artifact owner-get owner-set commands

SafarMirek:
See https://github.com/Apicurio/apicurio-cli/issues/9

### Get owner of the artifact
```
[miroslavsafar@m-bean-ipmi app-services-cli]$ ./rhoas service-registry artifact owner-get --help
Get owner of the artifact

Usage:
  rhoas service-registry artifact owner-get [flags]

Examples:
## Get owner of the artifact with artifact id 'example-name' in group 'example-group'
$ rhoas service-registry artifact owner-get --artifact-id example-name --group example-group


Flags:
      --artifact-id string   ID of the artifact
  -g, --group string         Artifact group (default "default")
      --instance-id string   ID of the Service Registry instance to be used (by default, uses the currently selected instance)

Global Flags:
  -h, --help      Show help for a command
  -v, --verbose   Enable verbose mode
```
### Set owner of the artifact
```
[miroslavsafar@m-bean-ipmi app-services-cli]$ ./rhoas service-registry artifact owner-set --help
Set owner of the specific artifact

Usage:
  rhoas service-registry artifact owner-set [flags]

Examples:
## Set owner of the artifact with artifact id 'example-name' in group 'example-group' to 'new-owner-name'
$ rhoas service-registry artifact owner-get --artifact-id example-name --group example-group --owner new-owner-name


Flags:
      --artifact-id string   ID of the artifact
  -g, --group string         Artifact group (default "default")
      --instance-id string   ID of the Service Registry instance to be used (by default, uses the currently selected instance)
      --owner string         Name of new owner

Global Flags:
  -h, --help      Show help for a command
  -v, --verbose   Enable verbose mode
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


rkpattnaik780:
Hi @SafarMirek . I am getting issues trying to run these commands:

```
NotFoundException: RESTEASY003210: Could not find resource for full path: http://bu98.serviceregistry.rhcloud.com/apis/registry/v2/groups/default/artifacts/tweets-topic-2/owner
```

Do we need to use some other environment?

rkpattnaik780:
Hi @SafarMirek . I am getting issues trying to run these commands:

```
NotFoundException: RESTEASY003210: Could not find resource for full path: http://bu98.serviceregistry.rhcloud.com/apis/registry/v2/groups/default/artifacts/tweets-topic-2/owner
```

Do we need to use some other environment?

rkpattnaik780:
Also the cli uses the convention: [verb]-[noun], like set-kafka, set-service-registry.
Will it be okay if we change these commands to get-owner and set-owner?

SafarMirek:
Currently, this feature is I think only in staging and in Operate First environment.

You can use cli with Operate First using this login command:
```
rhoas login --api-gateway https://fleet-manager-mt-apicurio-apicurio-registry.apps.smaug.na.operate-first.cloud --auth-url https://auth.apicur.io/auth/realms/operate-first-apicurio --client-id apicurio-cli
```

SafarMirek:
> Also the cli uses the convention: [verb]-[noun], like set-kafka, set-service-registry. Will it be okay if we change these commands to get-owner and set-owner?

I inspired myself in metadata commands:
```
Available Commands:
  create       Create new artifact from file or standard input
  delete       Deletes an artifact or all artifacts in a given group
  download     Download artifacts from Service Registry using global identifiers
  export       Export data from Service Registry instance
  get          Get artifact by ID, group, and version
  import       Import data into a Service Registry instance
  list         List artifacts
  metadata-get Get artifact metadata
  metadata-set Update artifact metadata
  state-set    Set artifact state
  update       Update artifact
  versions     Get latest artifact versions by artifact-id and group
```

EricWittmann:
LGTM - the names of the commands are consistent with existing metadata and state related commands.  So I think unless we wanted to change them all, then `owner-set` and `owner-get` seem like the right choices.

rkpattnaik780:
> I inspired myself in metadata commands:
I missed those commands, thanks for the clarifying @SafarMirek .

pkg/cmd/registry/artifact/owner/get.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	request := a.MetadataApi.GetArtifactOwner(opts.f.Context, opts.group, opts.artifact)
+
+	artifactOwner, _, err := request.Execute()
+	if err != nil {
+		return registrycmdutil.TransformInstanceError(err)
+	}
+
+	opts.f.Logger.Info(icon.SuccessPrefix(), *artifactOwner.Owner)
```
rkpattnaik780, line 101: Not sure if adding success prefix to owner name adds value, I think we can remove it, what do you think?

rkpattnaik780 (commented):
Nice PR @SafarMirek ! :)
Leaving some minor suggestions.
Also a question: do you really feel we should add interactive mode for owner operations?

pkg/cmd/registry/artifact/owner/get.go
SafarMirek: +1

rkpattnaik780:
> > Also a question: do you really feel we should add interactive mode for owner operations?
> 
> @rkpattnaik780 Does it have any downside?

No downsides at all. I personally prefer prompting when users have many flags to input or many options and also when options for one flag are dependent on the values provided for previous flags. I would have liked interactivity more if we could chose the owner and artifact-id, but that could be an overkill. Wdyt?

SafarMirek:
I am not sure there is a way to show all users who can be new owners. 

There is a way to let the user choose the artifact-id, but I am not sure it will be good UX because there could be hundreds of artifacts in the registry and in this scenario it would not make sense. Wdyt?

wtrocki:
I agree with Marek. While it is generally an good idea for this case it might be risky to show so many options (or good UX to have it with so many items)

rkpattnaik780:
> There is a way to let the user choose the artifact-id, but I am not sure it will be good UX because there could be hundreds of artifacts in the registry and in this scenario it would not make sense. Wdyt?

Great points @SafarMirek . I will agree as well, let's keep the interactive mode as it is.

rkpattnaik780 (approved):

PR #1746, opened 2022-09-19
Supporting gitops

wtrocki:
## Motivation

Enable users to fetch state for their organization and personal account required for kafka creation.
This approach enables granular set of commands that can be useful for users to get information about where kafka is deployed and how they are billed by creating kafka.  

### Fetching cloud providers and regions

```
[wtrocki@graphapi app-services-cli (supporting-gitops)]$ rhoas kafka providers 
  CLOUD PROVIDER   REGION     
 ---------------- ----------- 
  aws              eu-west-1  
  aws              us-east-1  
```

### Fetching billing types

This command fetching billing information - We need to extend this by including plans etc.
Depends on @rkpattnaik780 investigation on how we can represent all of that in gitops setup.

```
[wtrocki@graphapi app-services-cli (supporting-gitops ✗)]$ rhoas kafka billing
  BILLING    
 ----------- 
  developer  

```

### DIFF
--- a/docs/commands/rhoas_kafka.md
+++ b/docs/commands/rhoas_kafka.md
@@ -41,11 +41,13 @@ rhoas kafka topic create --name mytopic
 
 * [rhoas](rhoas.md)	 - RHOAS CLI
 * [rhoas kafka acl](rhoas_kafka_acl.md)	 - Manage Kafka ACLs for users and service accounts
+* [rhoas kafka billing](rhoas_kafka_billing.md)	 - List Kafka Billing Types
 * [rhoas kafka consumer-group](rhoas_kafka_consumer-group.md)	 - Describe, list, and delete consumer groups for the current Kafka instance
 * [rhoas kafka create](rhoas_kafka_create.md)	 - Create a Kafka instance
 * [rhoas kafka delete](rhoas_kafka_delete.md)	 - Delete a Kafka instance
 * [rhoas kafka describe](rhoas_kafka_describe.md)	 - View configuration details of a Kafka instance
 * [rhoas kafka list](rhoas_kafka_list.md)	 - List all Kafka instances
+* [rhoas kafka providers](rhoas_kafka_providers.md)	 - List Kafka Cloud Providers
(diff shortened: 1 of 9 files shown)


wtrocki:
@rkpattnaik780 feel free to add your results of investigation if needed.

rkpattnaik780:
Output of `rhoas kafka providers`:

```
  CLOUD PROVIDER (6)    REGION      SUPPORTED PLAN  
 --------------------- ----------- ---------------- 
  aws                   eu-west-1   developer.x1    
  aws                   eu-west-1   standard.x1     
  aws                   eu-west-1   standard.x2     
  aws                   us-east-1   developer.x1    
  aws                   us-east-1   standard.x1     
  aws                   us-east-1   standard.x2 
```

Output of `rhoas kafka billing`:

```
BILLING       MARKETPLACE ACCOUNT ID   MARKETPLACE PROVIDER  
 ------------- ------------------------ ---------------------- 
  marketplace   568056954830             aws                   
  standard                                               
```

dimakis:
I was just checking this out to check the response. 
This is great, I wonder if we could do with more information? like some more of an explainer as to what those options are? or at least pointing to docs somewhere that does explain what we mean by these terms, `standard` `marketplace` etc.

IMO even a small reminder to the user would suffice here. Trying to make it as easy as possible.

wtrocki:
Awesome work!
Made change for some ordering to make it more readable.

> Only developer instances are available

Would that be harder for people who just want to copy paste values?
IMHO Single value is simpler to parse by human than text :D

> or at least pointing to docs somewhere that does explain what we mean by these terms, standard marketplace etc.

good luck with that :)

rkpattnaik780:
> Would that be harder for people who just want to copy paste values?
> IMHO Single value is simpler to parse by human than text :D

Earlier implementation printed "developer" as the billing type, which should not be the case, the developer instances have billing types "marketplace" and "standard".
But again, those can be rejected by API.

wtrocki:
Knowing that is better to leave it this way

PR #1751, opened 2022-09-27
feat: do not check for context existence if a custom context file is set

lburgazzoli:
As today, it a custom context file is set, then the code check for the existence of the context file and fails if it does not exist, however, this is not true for custom configuration location. This PR aims to make config and context behaving consistently. 

### Verification Steps

1. Set `RHOASCONFIG` to an not existing file
2. Run `rhoas`
3. No error should be reported and the `RHOASCONFIG` file should have been created

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/core/servicecontext/file.go
+++ b/pkg/core/servicecontext/file.go
@@ -97,12 +97,6 @@ func (c *File) Remove() error {
 func (c *File) Location() (path string, err error) {
 
 	if rhoasContext := os.Getenv(ContextEnvName); rhoasContext != "" {
-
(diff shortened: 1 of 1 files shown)


rkpattnaik780 (approved):
This makes the handling of context in cli consistent with handling of context. I verified that OC creates the file as well, if it is missing.
Thanks a lot for this change!

rkpattnaik780:
The change has been released in [v0.51.2](https://github.com/redhat-developer/app-services-cli/releases/tag/v0.51.2).

PR #1768, opened 2022-10-19
refactor(context): use dataplane fetch approach

rkpattnaik780:
This PR highlights the problem with existing approach and explores a new approach to do minimal calls for service commands.

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/topic/list/list.go
+++ b/pkg/cmd/kafka/topic/list/list.go
@@ -16,6 +16,7 @@ import (
 	"github.com/redhat-developer/app-services-cli/pkg/shared/contextutil"
 	"github.com/redhat-developer/app-services-cli/pkg/shared/factory"
 	kafkainstanceclient "github.com/redhat-developer/app-services-sdk-go/kafkainstance/apiv1/client"
+	kafkamgmtclient "github.com/redhat-developer/app-services-sdk-go/kafkamgmt/apiv1/client"
 
 	"github.com/spf13/cobra"
 
@@ -30,11 +31,15 @@ type options struct {
 	Context        context.Context
 	ServiceContext servicecontext.IContext
 
-	kafkaID string
-	output  string
-	search  string
-	page    int32
-	size    int32
+	kafkaID       string
+	output        string
+	search        string
+	page          int32
+	size          int32
+	api           *kafkainstanceclient.APIClient
+	kafkaInstance *kafkamgmtclient.KafkaRequest
+
+	f *factory.Factory
 }
 
 type topicRow struct {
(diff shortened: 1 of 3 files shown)


pkg/shared/contextutil/util.go
```diff
@@ -93,6 +97,37 @@ func GetCurrentRegistryInstance(f *factory.Factory) (*registrymgmtclient.Registr
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	fmt.Println("inside get kafka dp")
+
+	api, kafkaInstance, err := conn.API().KafkaAdmin(currCtx.KafkaID)
+
+	fmt.Println("error - ", err)
+	if kafkamgmtv1errors.IsAPIError(err, kafkamgmtv1errors.ERROR_7) {
```
rkpattnaik780, line 122: Problem with this approach is, there is no way to identify a 404 error using error codes, which makes it difficult to map an error to a custom error message like we have for `GetCurrentKafkaInstance`. The only way to go seems to be using a `String.contains`, which is not a preferred way for error handling.

rkpattnaik780: Cc @jackdelahunt @dimakis 

Need a second opinion if we should follow the `string.contains` hack to optimize, or just let the data plane commands as they are (making two calls that return the current Kafka instance).

pkg/cmd/kafka/topic/list/list.go
```diff
@@ -61,9 +67,9 @@ func NewListTopicCommand(f *factory.Factory) *cobra.Command {
 		Long:    opts.localizer.MustLocalize("kafka.topic.list.cmd.longDescription"),
 		Example: opts.localizer.MustLocalize("kafka.topic.list.cmd.example"),
 		Args:    cobra.NoArgs,
-		RunE: func(cmd *cobra.Command, _ []string) error {
+		RunE: func(cmd *cobra.Command, _ []string) (err error) {
```
jackdelahunt, line 70: Should we be using named return types, personally i dont think they are great but its personal preference for you

pkg/shared/contextutil/util.go
```diff
@@ -93,6 +97,37 @@ func GetCurrentRegistryInstance(f *factory.Factory) (*registrymgmtclient.Registr
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	currCtx, err := GetCurrentContext(svcContext, f.Localizer)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	fmt.Println("inside get kafka dp")
```
jackdelahunt, line 117: need to remove

pkg/shared/contextutil/util.go
jackdelahunt: > there is no way to identify a 404 error using error codes

I was not aware of this can you explain why this is?

jackdelahunt:
The approach to use one call is good here just had a few other questions ^

Issue #1738, opened 2022-08-29
Improve error message when namespace set in context has been removed


rkpattnaik780:
## Description

**Version**: 0.50.0

Show more descriptive message when namespace set in the context has been removed. CLI provides limited information at the moment.

The `if` statement [here](https://github.com/redhat-developer/app-services-cli/blob/e69eae60b7656e93fdd8394f980f83d59444996b/pkg/shared/contextutil/util.go#L189) needs to be modified to check if the error returned is a not found error and return customized error message like the one for [Kafka service](https://github.com/redhat-developer/app-services-cli/blob/main/pkg/shared/contextutil/util.go#L72).

```
> rhoas context status         
❌ 410 Gone. Run the command in verbose mode using the -v flag to see more information
```

### Expected vs actual behaviour

Message displayed should be similar to what happens with Kafka:

```
rhoas status                              
❌ Kafka instance in your context doesn't exist.
Your instance might have been removed.
You can update your context by creating a new instance or by using the "rhoas context set-kafka" command
```

rkpattnaik780:
Cc @jackdelahunt

Thenitesh143:
I have just completed my first contribution. Is this a good project for a beginner like me. If yes please assign it to me. It's my humble request

rkpattnaik780:
Go ahead @Thenitesh143

Thenitesh143:
Can someone help me i can't find the error message and don't understand what to do with namespace

rkpattnaik780:
Hello Nitesh
The error `❌ 410 Gone` is returned from the server. The code that returns the error is [here](https://github.com/redhat-developer/app-services-cli/blob/e69eae60b7656e93fdd8394f980f83d59444996b/pkg/shared/contextutil/util.go#L189). Currently the error is being returned as it is. We want to return a custom error like we have for [Kafka instances](https://github.com/redhat-developer/app-services-cli/blob/main/pkg/shared/contextutil/util.go#L72). Notice how for Kafka, we return the error differently.

For connector, the error handling will be very similar to [this snippet](https://github.com/redhat-developer/app-services-cli/blob/e69eae60b7656e93fdd8394f980f83d59444996b/pkg/cmd/connector/build/build.go#L113). Let me know if that helps :).

Thenitesh143:
I'm no more working on this issue

yati1998:
Can I work on this issue?

rkpattnaik780:
Go ahead @yati1998

PR #1769, opened 2022-10-19
error: return detailed error message for ns not found

yati1998:
This commit improves the error handling in case the namespace is not found. It returns more detailed message rather than just returning the error.

Fixes: #1738

Signed-off-by: yati1998 <ypadia@redhat.com>


rkpattnaik780:
Thanks a lot for this PR @yati1998 . Great work!!

PR #1771, opened 2022-10-20
feat(service-account): add pagination to list command

rkpattnaik780:
Add pagination to `rhoas service-account list` command.

As users are used to running `rhoas service-account list | grep <username>` to get their service accounts, the default size has been set to 100.

### Verification Steps

1. `rhoas service-account list` should display upto 100 service accounts.

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/serviceaccount/list/list.go
+++ b/pkg/cmd/serviceaccount/list/list.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/redhat-developer/app-services-cli/internal/build"
+	"github.com/redhat-developer/app-services-cli/pkg/core/cmdutil"
 	"github.com/redhat-developer/app-services-cli/pkg/core/cmdutil/flagutil"
 	"github.com/redhat-developer/app-services-cli/pkg/core/config"
 	"github.com/redhat-developer/app-services-cli/pkg/core/ioutil/dump"
@@ -28,6 +30,8 @@ type options struct {
 
 	output       string
 	enableAuthV2 bool
+	page         int32
+	size         int32
 }
 
 // svcAcctRow contains the properties used to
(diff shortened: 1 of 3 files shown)


pkg/cmd/serviceaccount/list/list.go
```diff
@@ -71,6 +75,10 @@ func NewListCommand(f *factory.Factory) *cobra.Command {
 
 	_ = cmd.Flags().MarkDeprecated("enable-auth-v2", opts.localizer.MustLocalize("serviceAccount.common.flag.deprecated.enableAuthV2"))
 
+	cmd.Flags().Int32VarP(&opts.page, "page", "", int32(cmdutil.ConvertPageValueToInt32(build.DefaultPageNumber)), opts.localizer.MustLocalize("serviceAccount.list.flag.page.description"))
```
dimakis, line 78: Are we adding a `--page` flag? The flag doesn't work for me:

```shell
╰$ ./rhoas service-account --page 2
❌ Unknown flag: --page. Run the command in verbose mode using the -v flag to see more information
```

Another feature that might be nice is letting the user know how many service accounts there are? I know we output the number of service accounts displayed using -> `(100)`, but as a user I'd like to know if there are more, and if so how to get them. I'm unsure if this is difficult?

This would mean that the `-h` has to be updated too.

rkpattnaik780: Hi @dimakis 
The command you'd like to run is `rhoas service-account list --page 2`.

We have thought of improving the overall experience with pagination in CLI by providing additional information. Though it doesn't seem to be possible for service-accounts as of now.

The response we get is an array, so there is no way to know the total no. of service accounts.

jackdelahunt:
Running `rhoas service-account list` works as epxected. and so does `rhoas service-account list --size=100` as they are pretty much the same thing.

Something that is interesting though is when running `rhoas service-account list --size=101 -v` I get the following error.
```
{"fields":{"max":"must be less than or equal to 100"},"error":"invalid_field","error_description":"Request failed field validation"}
```

Wondering if we can have a custom error message for this as this just returns 400 without verbose mode.

rkpattnaik780:
> Wondering if we can have a custom error message for this as this just returns 400 without verbose mode.

Nice catch @jackdelahunt :).
Unfortunately we dont have custom error message for this, but we can add client level validation to ensure page > 0 and size <= 100

rkpattnaik780:
Custom validations have been added @jackdelahunt 

```
./rhoas service-account list --page 0
❌ Invalid page number 0, minimum value is 1. Run the command in verbose mode using the -v flag to see more information

➜  cli git:(svc_acct_pagination) ✗ ./rhoas service-account list --page -1
❌ Invalid page number -1, minimum value is 1. Run the command in verbose mode using the -v flag to see more information

➜  cli git:(svc_acct_pagination) ✗ ./rhoas service-account list --size 101
❌ Invalid size 101, maximum value is 100. Run the command in verbose mode using the -v flag to see more information

➜  cli git:(svc_acct_pagination) ✗ ./rhoas service-account list --size 0  
❌ Invalid size 0, minimum value is 1. Run the command in verbose mode using the -v flag to see more information
```

jackdelahunt (approved):
LGTM 👍

PR #1779, opened 2022-10-28
Localize strings in request command

### LINKED ISSUE #1767: Localize strings in request command
The short description, long description and example for the `request` command currently use [raw strings](https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/request/request.go#L38-L45). They should be localized as in other commands. You can refer to [this example](https://github.com/redhat-developer/app-services-cli/blob/main/pkg/cmd/whoami/whoami.go#L34-L36).

A new `toml` file needs to be added to `pkg/core/localize/locales/en/cmd`, whose keys should then be used in the `request` command using the `localizer` object.

This will also fix the indentation of `rhoas request` help text.

patrikhl:
Closes #1767

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/docs/commands/rhoas_request.md
+++ b/docs/commands/rhoas_request.md
@@ -6,6 +6,7 @@ Allows you to perform API requests against the API server
 
 Command allows users you to perform API requests against the API server.
 
+
 ```
 rhoas request [flags]
 ```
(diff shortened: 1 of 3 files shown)


rkpattnaik780 (approved):
Thanks a lot for the PR @patrikhl !!
Great work!

PR #1792, opened 2023-01-04
fix(kafka create): sync marketplace provider with cloud provider

rkpattnaik780:
Selection of marketplace provider should be dependent on the cloud provider.

### Verification Steps

1. Use mock server.
2. Run Kafka create in interactive mode. User should not be able to select `gcp` as marketplace when `aws` is chosen as the cloud provider:
```
rhoas kafka create --dry-run -v
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -2,6 +2,7 @@ package create
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -257,7 +258,13 @@ func runCreate(opts *options) error {
 				CloudAccountID: opts.marketplaceAcctId,
 			}
 
-			userQuota, err = accountmgmtutil.SelectQuotaForUser(f, orgQuotas, marketplaceInfo)
+			if opts.marketplace != "" && opts.marketplace != accountmgmtutil.RedHatMarketPlace {
+				if opts.marketplace != opts.provider {
+					return errors.New("selected marketplace is not supported for the cloud provider")
+				}
+			}
+
+			userQuota, err = accountmgmtutil.SelectQuotaForUser(f, orgQuotas, marketplaceInfo, opts.provider)
 			if err != nil {
 				return err
 			}
@@ -477,7 +484,13 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 		return nil, err
 	}
 
-	availableBillingModels := FetchSupportedBillingModels(orgQuota)
+	availableBillingModels := FetchSupportedBillingModels(orgQuota, answers.CloudProvider)
+
+	if len(availableBillingModels) == 0 && len(orgQuota.MarketplaceQuotas) > 0 {
+		return nil, errors.New("standard instances are unavailable for the cloud provider, try another provider")
+	}
+
+	fmt.Println("Supported Billing Models - ", availableBillingModels)
 
 	if len(availableBillingModels) > 0 {
 		if len(availableBillingModels) == 1 {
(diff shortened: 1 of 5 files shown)


pkg/cmd/kafka/create/create.go
```diff
@@ -477,7 +484,13 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
 	}
 
-	availableBillingModels := FetchSupportedBillingModels(orgQuota)
+	availableBillingModels := FetchSupportedBillingModels(orgQuota, answers.CloudProvider)
+
+	if len(availableBillingModels) == 0 && len(orgQuota.MarketplaceQuotas) > 0 {
+		return nil, errors.New("standard instances are unavailable for the cloud provider, try another provider")
+	}
+
+	fmt.Println("Supported Billing Models - ", availableBillingModels)
```
jackdelahunt: Debugging print statements 🙃

pkg/cmd/kafka/create/create.go
```diff
@@ -477,7 +484,13 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 		return nil, err
 	}
 
-	availableBillingModels := FetchSupportedBillingModels(orgQuota)
+	availableBillingModels := FetchSupportedBillingModels(orgQuota, answers.CloudProvider)
+
+	if len(availableBillingModels) == 0 && len(orgQuota.MarketplaceQuotas) > 0 {
+		return nil, errors.New("standard instances are unavailable for the cloud provider, try another provider")
```
jackdelahunt: These error messages should be added to our lang files right? Also one line 263

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -131,12 +136,31 @@ func SelectQuotaForUser(f *factory.Factory, orgQuota *OrgQuotas, marketplaceInfo
...
+						break
+					}
+				}
+			}
+
+			orgQuota.MarketplaceQuotas = uniqueQuota(filteredMarketPlaceQuotas)
+		}
+
+		if len(orgQuota.MarketplaceQuotas) == 0 {
+			return nil, errors.New("no marketplace quota available for given provider")
```
jackdelahunt: Localise

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -289,3 +346,15 @@ func unique(s []string) []string {
...
+	inResult := make(map[QuotaSpec]bool)
+	var result []QuotaSpec
+	for _, quota := range s {
+		if _, ok := inResult[quota]; !ok {
+			inResult[quota] = true
+			result = append(result, quota)
+		}
+	}
+	return result
+}
```
jackdelahunt, line 360: This is a small thing but I had to read this a few times to figure out what was happening (we are removing all duplicates from the slice). Maybe just giving this a rename would make it clearer for code reading in the future.

jackdelahunt (changes requested):

pkg/shared/accountmgmtutil/ams.go
jackdelahunt: Something like, removeDuplicateQuotaSpec maybe?

rkpattnaik780: Having unique as prefix makes more sense imo, I have added an inline description for the method. Can you check it and let me know wdyt?

jackdelahunt: Yeah, the comment is fine, I tend to rather document with the naming of things then comments as they can be outdated and not read most of the time.

I have a small rule, if my comment isnt 3 or more lines long it probably means the code can be made clearer without it (or there is something very niche happening that is just difficult so its fine). 

But that is just how I go about it so this is fine 👍

jackdelahunt (approved):

PR #1800, opened 2023-01-30
feat(kafka create): enable creating long-lived trial instances

rkpattnaik780:
Add additional billing model `eval` to create long lived trial instances.

### Verification Steps
1. Create a local binary of the CLI:
```
make binary
```
2. Login to an org having quotas for long lived trial instances
```
rhoas login --api-gateway staging
```
3. Run kafka create command setting `eval` as billing model
```
./rhoas kafka create --name test-instance --billing-model eval --dry-run -v
```
4. The logs should show that the selected quota object with name "eval".
5. Request body should have billing model "eval".

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/shared/accountmgmtutil/ams.go
+++ b/pkg/shared/accountmgmtutil/ams.go
@@ -93,16 +93,17 @@ func GetOrgQuotas(f *factory.Factory, spec *remote.AmsConfig) (*OrgQuotas, error
 		for i := range quotaResources {
 			quotaResource := quotaResources[i]
 			if quotaResource.GetResourceName() == spec.ResourceName {
-				if quotaResource.GetProduct() == spec.TrialProductQuotaID {
+				switch quotaResource.GetProduct() {
+				case spec.TrialProductQuotaID:
 					trialQuotas = append(trialQuotas, QuotaSpec{QuotaTrialType, 0, quotaResource.BillingModel, nil})
-				} else if quotaResource.GetProduct() == spec.InstanceQuotaID {
+				case spec.InstanceQuotaID:
 					remainingQuota := int(quota.GetAllowed() - quota.GetConsumed())
 					if quotaResource.BillingModel == QuotaStandardType {
 						standardQuotas = append(standardQuotas, QuotaSpec{QuotaStandardType, remainingQuota, quotaResource.BillingModel, nil})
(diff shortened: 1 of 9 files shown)


ziccardi:
If you specify the `EVAL` billing model, KFM requires that you specify the instance type you want to create (as of today, `standard` is the only accepted value).

rkpattnaik780:
Hello @ziccardi I have addressed your comments and off-thread discussions. Just wanted to confirm if the `instance-type` here is set in `Plan` along with size or will it be a new parameter in the Kafka create payload?

ziccardi:
@Hi @rkpattnaik780 ! After yesterday's chat, I looked back into the code and made some tests. I was wrong: specifying the instance type/plan is not mandatory. You can specify only the billing model. Sorry for the confusion.

pkg/shared/accountmgmtutil/ams.go
```diff
@@ -86,56 +87,80 @@ func GetOrgQuotas(f *factory.Factory, spec *remote.AmsConfig) (*OrgQuotas, error
...
 					trialQuotas = append(trialQuotas, QuotaSpec{QuotaTrialType, 0, quotaResource.BillingModel, nil})
-				} else if quotaResource.GetProduct() == spec.InstanceQuotaID {
+				case spec.InstanceQuotaID:
 					remainingQuota := int(quota.GetAllowed() - quota.GetConsumed())
 					if quotaResource.BillingModel == QuotaStandardType {
 						standardQuotas = append(standardQuotas, QuotaSpec{QuotaStandardType, remainingQuota, quotaResource.BillingModel, nil})
 					} else if quotaResource.BillingModel == QuotaMarketplaceType {
 						marketplaceQuotas = append(marketplaceQuotas, QuotaSpec{QuotaMarketplaceType, remainingQuota, quotaResource.BillingModel, quota.CloudAccounts})
 					}
+				case "RHOSAKEval":
```
jackdelahunt, line 106: Do we need to be using a string literal here

jackdelahunt (approved):

pkg/shared/accountmgmtutil/ams.go
rkpattnaik780: These literals are to be fetched from a GitHub file, it will work only after merging this PR, I will do it in a follow up PR.

PR #1794, opened 2023-01-18
feat: addition of the new dedicated command to select ocm cluster

dimakis:
This is still not completed but what is still needed is listed via TODOs in the code.

I'm just looking to see reviews on what I've done thus far

### Verification Steps

1. An OCM cluster is needed to verify this so I'll need to hop on to verify with the reviewer
### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation change
- [ ] Other (please specify)

PR #1803, opened 2023-02-03
Add custom cluster mgmt url flag

dimakis:
This allows the user to specify a custom api-gateway for their clustermgmt (ocm) which allows the CLI to install stage Addons if the targeted KFM deployment is stage.

Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/dedicated/register/registercluster.go
+++ b/pkg/cmd/dedicated/register/registercluster.go
@@ -378,17 +378,12 @@ func selectAccessPrivateNetworkInteractivePrompt(opts *options) error {
 		Help:    opts.f.Localizer.MustLocalize("dedicated.registerCluster.prompt.selectPublicNetworkAccess.help"),
 		Default: false,
 	}
-	accessKafkasViaPublicNetwork := false
-	err := survey.AskOne(prompt, &accessKafkasViaPublicNetwork)
+	accessFromPublicNetwork := true
+	err := survey.AskOne(prompt, accessFromPublicNetwork)
 	if err != nil {
 		return err
 	}
-	if accessKafkasViaPublicNetwork {
-		opts.accessKafkasViaPrivateNetwork = false
-	} else {
-		opts.accessKafkasViaPrivateNetwork = true
-	}
-
+	opts.accessKafkasViaPrivateNetwork = !accessFromPublicNetwork
 	return nil
 }
 
@@ -440,7 +435,7 @@ func getStrimziAddonIdByEnv(con *config.Config) string {
 	return strimziAddonIdQE
 }
 
-func getKafkaFleetManagerAddonIdByEnv(con *config.Config) string {
+func getKafkaFleetShardAddonIdByEnv(con *config.Config) string {
 	if con.APIUrl == build.ProductionAPIURL {
 		return fleetshardAddonId
 	}
@@ -481,7 +476,7 @@ func registerClusterWithKasFleetManager(opts *options) error {
 	if err != nil {
 		return err
 	}
-	err = createAddonWithParams(opts, getKafkaFleetManagerAddonIdByEnv(con), response.FleetshardParameters)
+	err = createAddonWithParams(opts, getKafkaFleetShardAddonIdByEnv(con), response.FleetshardParameters)
 	if err != nil {
 		return err
 	}
--- a/pkg/cmd/dedicated/register/registercluster.go
+++ b/pkg/cmd/dedicated/register/registercluster.go
@@ -33,19 +33,18 @@ type options struct {
 
 // list of consts should come from KFM
 const (
-	machinePoolId                  = "kafka-standard"
-	machinePoolTaintKey            = "bf2.org/kafkaInstanceProfileType"
-	machinePoolTaintEffect         = "NoExecute"
-	machinePoolTaintValue          = "standard"
-	machinePoolInstanceType        = "r5.xlarge"
-	machinePoolLabelKey            = "bf2.org/kafkaInstanceProfileType"
-	machinePoolLabelValue          = "standard"
-	clusterReadyState              = "ready"
-	defaultClusterManagementAPIURL = "https://api.openshift.com"
-	fleetshardAddonId              = "kas-fleetshard-operator"
-	strimziAddonId                 = "managed-kafka"
-	fleetshardAddonIdQE            = "kas-fleetshard-operator-qe"
-	strimziAddonIdQE               = "managed-kafka-qe"
+	machinePoolId           = "kafka-standard"
+	machinePoolTaintKey     = "bf2.org/kafkaInstanceProfileType"
+	machinePoolTaintEffect  = "NoExecute"
+	machinePoolTaintValue   = "standard"
+	machinePoolInstanceType = "r5.xlarge"
+	machinePoolLabelKey     = "bf2.org/kafkaInstanceProfileType"
+	machinePoolLabelValue   = "standard"
+	clusterReadyState       = "ready"
+	fleetshardAddonId       = "kas-fleetshard-operator"
+	strimziAddonId          = "managed-kafka"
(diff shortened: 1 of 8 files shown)


pkg/cmd/dedicated/register/registercluster.go
```diff
@@ -465,11 +473,15 @@ func registerClusterWithKasFleetManager(opts *options) error {
...
+	con, err := opts.f.Config.Load()
+	if err != nil {
+		return err
+	}
+	err = createAddonWithParams(opts, getStrimziAddonIdByEnv(con), nil)
 	if err != nil {
 		return err
 	}
-	err = createAddonWithParams(opts, fleetshardAddonId, response.FleetshardParameters)
+	err = createAddonWithParams(opts, getKafkaFleetManagerAddonIdByEnv(con), response.FleetshardParameters)
```
machi1990, line 484:
```suggestion
	err = createAddonWithParams(opts, getKafkaFleetShardAddonIdByEnv(con), response.FleetshardParameters)
```

pkg/cmd/dedicated/register/registercluster.go
```diff
@@ -439,6 +433,20 @@ func createAddonWithParams(opts *options, addonId string, params *[]kafkamgmtcli
...
 }
 
+func getStrimziAddonIdByEnv(con *config.Config) string {
+	if con.APIUrl == build.ProductionAPIURL {
+		return strimziAddonId
+	}
+	return strimziAddonIdQE
+}
+
+func getKafkaFleetManagerAddonIdByEnv(con *config.Config) string {
```
machi1990, line 443:
```suggestion
func getKafkaFleetShardAddonIdByEnv(con *config.Config) string {
```

machi1990 (commented):
left one suggestion on function name.

reviewed the change, it responds to MVP requirement.
In the near term future, let's progress the conversation about KFM having a metadata API that drives most of the things here.

machi1990 (changes requested):
hi @dimakis this is semi unrelated but can we sort out the wording in https://github.com/redhat-developer/app-services-cli/blob/5f68f800ab0d67d90c34301aede29fc9a767e838/pkg/core/localize/locales/en/cmd/dedicated.en.toml#L29

it should read `Kafkas` and not `Kakfas`  thanks ;-)

machi1990 (changes requested):
The lines https://github.com/redhat-developer/app-services-cli/blob/5f68f800ab0d67d90c34301aede29fc9a767e838/pkg/cmd/dedicated/register/registercluster.go#L386-L390 are inversed.

It should be 
```go
if accessKafkasViaPublicNetwork {
		opts.accessKafkasViaPrivateNetwork = true
	} else {
		opts.accessKafkasViaPrivateNetwork = false
	}
```
or simply 
```go
opts.accessKafkasViaPrivateNetwork=accessKafkasViaPublicNetwork
```

While we are at it, the whole method can be refactored to
```go
prompt := &survey.Confirm{
		Message: opts.f.Localizer.MustLocalize("dedicated.registerCluster.prompt.selectPublicNetworkAccess.message"),
		Help:    opts.f.Localizer.MustLocalize("dedicated.registerCluster.prompt.selectPublicNetworkAccess.help"),
		Default: opts.accessKafkasViaPrivateNetwork,
	}
	err := survey.AskOne(prompt, &opts.accessKafkasViaPrivateNetwork)
	if err != nil {
		return err
	}
return nil
```

dimakis:
> 

do you suspect that a private network will be the default? the reason I have that as is, is that I presumed public would be default. This was also mentioned by Keith Wall. I'm easy either way really.

machi1990:
> > 
> 
> @machi1990 do you suspect that a private network will be the default? the reason I have that as is, is that I presumed public would be default. This was also mentioned by Keith Wall. I'm easy either way really.

Hi @dimakis there is a bug in that logic which I highlighted. If someone choose to allow for private kafkas - i.e `accessKafkasViaPublicNetwork set to "true"` then with the logic current in place, it'll send the value as `false`. And vice versa.

machi1990:
> > 
> 
> @machi1990 do you suspect that a private network will be the default? the reason I have that as is, is that I presumed public would be default. This was also mentioned by Keith Wall. I'm easy either way really.

Also note: the suggested fix does not change the default value.

dimakis:
> > @machi1990 do you suspect that a private network will be the default? the reason I have that as is, is that I presumed public would be default. This was also mentioned by Keith Wall. I'm easy either way really.
> 
> Hi @dimakis there is a bug in that logic which I highlighted. If someone choose to allow for private kafkas - i.e `accessKafkasViaPublicNetwork set to "true"` then with the logic current in place, it'll send the value as `false`. And vice versa.

@machi1990 It is not a bug, its a feature 😛 
The wording is changed to public, i.e. `would you like your kafkas to have access to a public network` hence the inversion.

machi1990:
> > > @machi1990 do you suspect that a private network will be the default? the reason I have that as is, is that I presumed public would be default. This was also mentioned by Keith Wall. I'm easy either way really.
> > 
> > 
> > Hi @dimakis there is a bug in that logic which I highlighted. If someone choose to allow for private kafkas - i.e `accessKafkasViaPublicNetwork set to "true"` then with the logic current in place, it'll send the value as `false`. And vice versa.
> 
> @machi1990 It is not a bug, its a feature stuck_out_tongue The wording is changed to public, i.e. `would you like your kafkas to have access to a public network` hence the inversion.
> 
> it is a little confusing yes, as the var in the endpoint is `private`

Right. That's confusing. And the fact that it defaults to `no` public access is also confusing.

pkg/cmd/dedicated/register/registercluster.go
```diff
@@ -384,17 +378,14 @@ func selectAccessPrivateNetworkInteractivePrompt(opts *options) error {
...
-	accessKafkasViaPublicNetwork := false
-	err := survey.AskOne(prompt, &accessKafkasViaPublicNetwork)
+	accessFromPublicNetwork := true
+	err := survey.AskOne(prompt, accessFromPublicNetwork)
 	if err != nil {
 		return err
 	}
-	if accessKafkasViaPublicNetwork {
+	if !accessFromPublicNetwork {
 		opts.accessKafkasViaPrivateNetwork = false
```
machi1990, line 387:
```suggestion
opts.accessKafkasViaPrivateNetwork = !accessFromPublicNetwork
```

pkg/cmd/dedicated/register/registercluster.go
```diff
@@ -384,17 +378,13 @@ func selectAccessPrivateNetworkInteractivePrompt(opts *options) error {
...
+	err := survey.AskOne(prompt, accessFromPublicNetwork)
 	if err != nil {
 		return err
 	}
-	if accessKafkasViaPublicNetwork {
-		opts.accessKafkasViaPrivateNetwork = false
-	} else {
-		opts.accessKafkasViaPrivateNetwork = true
+opts.accessKafkasViaPrivateNetwork = !accessFromPublicNetwork
 	}
```
machi1990, line 387:
```suggestion
```

machi1990: trailing brace

pkg/core/localize/locales/en/cmd/dedicated.en.toml
```diff
@@ -26,7 +26,7 @@ one = 'The ID of the OpenShift cluster to register:'
 one = 'Select the ready cluster to register'
 
 [dedicated.registerCluster.prompt.selectPublicNetworkAccess.message]
-one = 'Would you like your Kakfas to be accessible via a public network?'
+one = 'Would you like your Kafkas to be accessible via a public network?'
```
rkpattnaik780, line 29: Wondering if we should use `Kafka instances` instead of `Kafkas`

jackdelahunt:
Closed because work is done in other pr #1794

PR #1804, opened 2023-02-08
Add cluster id to create kafka

dimakis:
Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


pkg/cmd/kafka/create/create.go
```diff
@@ -237,6 +239,7 @@ func runCreate(opts *options) error {
 			Name:          opts.name,
 			Region:        &opts.region,
 			CloudProvider: &opts.provider,
+			ClusterId:     *kafkamgmtclient.NewNullableString(&opts.clusterId),
```
machi1990, line 242: @dimakis this is awesome work, I'd like to us to please let's take a note that we'll need to extend this over to https://github.com/redhat-developer/app-services-cli/blob/36badbc88c93c4eef81aab174a5a76a13257ad3d/pkg/cmd/kafka/create/create.go#L449 as well.

i.e what is done here is fine for non interactive mode but we need to build on top of this by thinking on how we can going to extend this to interactive mode. We'll need to get UX involved in this.

i.e what kind of flow we want customers to have. `list existings clusters -> pick one -> etc`. That can be done in a follow PR and we'll need to involve UX on this. Let's sync with them.

pkg/core/localize/locales/en/cmd/dedicated.en.toml
```diff
@@ -26,7 +26,7 @@ one = 'The ID of the OpenShift cluster to register:'
 one = 'Select the ready cluster to register'
 
 [dedicated.registerCluster.prompt.selectPublicNetworkAccess.message]
-one = 'Would you like your Kakfas to be accessible via a public network?'
+one = 'Would you like your Kafkas to be accessible via a public network?'
```
machi1990, line 29: nice

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -1144,3 +1144,7 @@ one = 'Kafka instance "{{.Name}}" is not ready yet'
 [kafka.update.flag.owner]
 description = 'Description for the --owner flag'
 one = 'ID of the Kafka instance owner'
+
+[kafka.create.flag.clusterId.description]
+description = 'Description for the --cluster-id flag'
+one = 'ID of the Customer-Cloud dataplane cluster to create the Kafka instance on.'
```
machi1990, line 1150:
```suggestion
one = 'ID of the Customer-Cloud data plane cluster to create the Kafka instance on.'
```

machi1990 (approved):
it coughs and wheezes!

PR #1805, opened 2023-02-09
feat(kafka): add promote command

rkpattnaik780:
`rhoas kafka promote` command to promote eval instances to use paid subscriptions.

### Verification Steps

1. Create an eval instance:
```
rhoas kafka create --name test-instance --billing-model eval
```
2. Run promote command to change billing model to standard.
```
rhoas kafka promote --billing-model standard
```

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


ziccardi:
I gave a run to the code and it works very well.

I have only a few comments:

1. If you run the `promote` command and the `kafka` is not in a valid state for promotion (ready/suspended/resuming), the cli gives a misleading error : `Only Kafka instances with billing model "eval" can be promoted`
2. The `promote` is valid only for `eval` instances, but when I list the kafkas, the billing model is not visualised. How can I know what are the kafkas I can visualise?
3. When the `promote` is called and it is successful, it returns an 'instance has been successfully promoted'. That is misleading. When you invoke the promotion endpoint, the `promotion` request gets accepted. A reconciler then manages the promotion. At this point, the promotion can still fail. The `cli` should take into consideration the `promotion_status` and the `promotion_details` fields.
4. The billing_model is not shown in the kafka list: I have no way to know the current billing model and I don't have any way to check that the promotion has been really done.

rkpattnaik780:
Hello @ziccardi 
Thanks for the review. While going through the wrong error message issue, it turns out that the custom error code  `KAFKAS-MGMT-9` is associated with two failed reasons - `kafka request \"cfivi8ubtv3kjdprf0v0\" has a kafka billing model \"standard\". Only kafka requests with a kafka billing model 'eval' can be promoted` and `kafka request \"cfivi8ubtv3kjdprf0v0\" with status \"accepted\" cannot be promoted: promotable status are: [ready suspended resuming]"`. Is there any reason we have two errors mapped to single error code? Shouldn't they be unique?

ziccardi:
> Is there any reason we have two errors mapped to single error code? Shouldn't they be unique?

Currently we return those errors as `general errors`, so they share the same code. If you need that to use different codes, please fire a jira and we will look into it

rkpattnaik780:
> > Is there any reason we have two errors mapped to single error code? Shouldn't they be unique?
> 
> Currently we return those errors as `general errors`, so they share the same code. If you need those to use different codes, please fire a jira and we will look into it

Thank you for explaining. Will be creating a JIRA.

dimakis (approved):
LGTM, much better success message 👍

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -54,6 +54,15 @@ one = 'Kafka instance ID. Uses the current instance if not set'
 [kafkas.common.flag.output.description]
 one = 'Format in which to display the Kafka instances (choose from: "json", "yml", "yaml")'
 
+[kafka.common.flag.marketplaceId.description]
+one = 'Cloud Account ID for the marketplace'
```
jbyrne-redhat, line 58: Suggest: "Account ID for cloud marketplace"

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -54,6 +54,15 @@ one = 'Kafka instance ID. Uses the current instance if not set'
...
+
+[kafka.common.flag.billingModel.description]
+one = 'Billing model to be used'
+
+[kafka.common.flag.marketplaceType.description]
+one = 'Name of the marketplace where the instance is purchased on'
```
jbyrne-redhat, line 64: Suggest: 'Name of the cloud marketplace where the instance was purchased'

pkg/core/localize/locales/en/cmd/kafka.en.toml
```diff
@@ -396,6 +396,9 @@ one = 'unable to create new Kafka instance at this time in specified cloud provi
 [kafka.create.error.instance.limit]
 one = 'maximum number of allowed kafka instances has been reached. Please review all instances that your user has access to and delete one or more instances before creating a new one'
```
jbyrne-redhat, line 397: Suggest: 'the maximum number of Kafka instances for your organization has been reached. Before you can create a new instance, you must delete one or more of the existing instances'

jbyrne-redhat (commented):
Descriptions look good overall, @rkpattnaik780 . A couple of minor suggestions.

PR #1806, opened 2023-02-09
fix(dedicated): attempts to correctly parse the cluster ingress dns name

machi1990:
This is so that "apps." prefix is always included and required by KFM when it is running under certain configuration

Related to https://github.com/bf2fc6cc711aee1a0c2a/kas-installer/issues/266#issuecomment-1424156487

### Verification Steps

1. `make binary`
2. `./rhoas dedicated register-cluster  -v` and observe that the KFM cluster registration payload now contains a valid `cluster_ingress_dns_name` field i.e with `apps.` prefix in it.

### Type of change

- ~~[ ] New feature (non-breaking change which adds functionality)~~
- ~~[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)~~
- ~~[ ] Documentation change~~
- ~~[ ] Other (please specify)~~

### DIFF
--- a/pkg/cmd/dedicated/register/registercluster.go
+++ b/pkg/cmd/dedicated/register/registercluster.go
@@ -3,9 +3,10 @@ package register
 import (
 	"context"
 	"fmt"
+	"strings"
+
 	"github.com/redhat-developer/app-services-cli/internal/build"
 	"github.com/redhat-developer/app-services-cli/pkg/core/config"
-	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	clustersmgmtv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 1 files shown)


pkg/cmd/dedicated/register/registercluster.go
```diff
@@ -181,13 +182,19 @@ func runClusterSelectionInteractivePrompt(opts *options) error {
...
 	}
-	return strings.SplitAfter(clusterIngressDNSName, ".apps.")[1], nil
+
+	splits := strings.SplitAfterN(clusterIngressDNSName, "console-openshift-console.", 2)
+	if len(splits) == 2 {
+		return splits[1], nil
+	}
+
+	return "", fmt.Errorf("could not construct cluster_ingress_dns_name")
 }
```
machi1990, line 198: The logic might not be needed from the CLI in the next releases of KFM - i.e KFM might take over this logic and do it on a case by case basis.

MikeEdgar (approved):
lgtm - this is working for me.

machi1990:
hey @dimakis can you've a look as well? Thanks

Might be good to have another release of the CLI afterward.

machi1990:
> @machi1990 I've tested it, and LGTM

Thanks @dimakis 

>  I do not know enough about when "CNAME registration is not used". 

That's a KFM configuration that among other things, controls how Kafka URLs are generated. When CNAME is not activated, the URLs are based of the cluster dns. And it is in this particular scenario (usually under development or testing environments setups) where a missing `apps.` was going to be problematic for those Kafkas. Once that happens, the Kafka is more or less not usabled.

> are multiple ways for a DNS to be obtained?

Yes. Another alternative is to directly grab the value from the `clusters/<cluster-id>/ingress` endpoints - but that only works for public accessible clusters. 

>  and is `.apps.` always in the DNS name itself?

Yes, `apps` is always going to be there. That's the default ingress in OpenShift.

machi1990:
@dimakis once this is merged, please proceed with a new alpha release as QE might be looking to start testing this using kas-installer.

dimakis:
LGTM

PR #1816, opened 2023-02-23
feat: now listing customer cloud in rhoas kafka list

jackdelahunt:
Closes # 

### Verification Steps

1. Have an enterprise cluster
2. `rhoas kafka list` you will see the customer cloud info listed

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/list/list.go
+++ b/pkg/cmd/kafka/list/list.go
@@ -34,6 +34,7 @@ type kafkaRow struct {
 	Status        string `json:"status" header:"Status"`
 	CloudProvider string `json:"cloud_provider" header:"Cloud Provider"`
 	Region        string `json:"region" header:"Region"`
+	CustomerCloud string `json:"customer_cloud" header:"Customer Cloud"`
 }
 
 type options struct {
@@ -160,6 +161,14 @@ func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId st
 		if k.GetId() == selectedId {
 			name = fmt.Sprintf("%s %s", name, icon.Emoji("✔", "(current)"))
 		}
+
+		var customerCloud string
+		if id, ok := k.GetClusterIdOk(); ok {
+			customerCloud = *id
+		} else {
+			customerCloud = "Red Hat Infrastructure"
+		}
+
 		row := kafkaRow{
 			ID:            k.GetId(),
 			Name:          name,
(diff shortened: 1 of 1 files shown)


pkg/cmd/kafka/list/list.go
```diff
@@ -160,13 +161,22 @@ func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId st
 		if k.GetId() == selectedId {
 			name = fmt.Sprintf("%s %s", name, icon.Emoji("✔", "(current)"))
 		}
+
+		var customerCloud string
+		if id, ok := k.GetClusterIdOk(); ok {
+			customerCloud = *id
+		} else {
+			customerCloud = "Redhat Infrastructure"
```
machi1990, line 169:
```suggestion
			customerCloud = "Red Hat Infrastructure"
```
should it be?

pkg/cmd/kafka/list/list.go
```diff
@@ -34,6 +34,7 @@ type kafkaRow struct {
 	Status        string `json:"status" header:"Status"`
 	CloudProvider string `json:"cloud_provider" header:"Cloud Provider"`
 	Region        string `json:"region" header:"Region"`
+	CustomerCloud string `json:"customer_cloud" header:"CustomerCloud"`
```
machi1990, line 37: Was this the agreed name?

machi1990: If not already agreed, let's revisit this by using the agreed header name.

machi1990: Also the header has to be `Customer Cloud` if we stick with that name:-)

machi1990 (changes requested):
Left naming suggestions.

If the names are not agreed at the moment, let's revisit them at some point.

pkg/cmd/kafka/list/list.go
machi1990, line 37:
```suggestion
	CustomerCloud string `json:"customer_cloud" header:"Customer Cloud"`
```

jackdelahunt: We have not officially agreed on a name for the output, @dimakis what do you think?

dimakis: There is no agreed names as of yet, I've brought this up a couple of times at meetings but so far we don't have names for `Red Hat OpenShift Cluster` or  `Your OpenShift Cluster`

jackdelahunt: From those two I think `Red Hat OpenShift Cluster` sounds good to me

machi1990: Let's go with what we've hear and raise it in the sync call.

PR #1815, opened 2023-02-22
fix: pass no params during the installation of managed kafka addon

dimakis:
Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation change
- [ ] Other (please specify)

PR #1817, opened 2023-02-23
feat: addition of the new list cluster cmd

dimakis:
You must have an enterprise cluster registered with KFM to view clusters:
```shell
rhoas dedicated list
```

Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)


pkg/cmd/dedicated/listclusters/list_clusters.go
```diff
@@ -0,0 +1,176 @@
...
+	opts.kfmClusterList = clist
+	f.Logger.Info(response)
+	if len(opts.kfmClusterList.Items) == 0 {
+		return f.Localizer.MustLocalizeError("dedicated.listClusters.log.info.noClusters")
+	}
+	return nil
+}
+
+func getPaginatedClusterList(opts *options) error {
+	cl, err := clustermgmt.GetClusterList(opts.f, opts.accessToken, opts.clusterManagementApiUrl, opts.page, opts.limit)
```
machi1990, line 107: I'll suggest filtering by only the clusters that have been returned by `listEnterpriseClusters`.

i.e from `opts.kfmClusterList.Items`, you could map the ids and use them to make a call to OCM with a search parameter like:

```
ocm get /api/clusters_mgmt/v1/clusters --parameter search="id = 'id1' or id = 'id2' or .... id='idN'"
```
This will reduce the need for https://github.com/redhat-developer/app-services-cli/blob/bf1a206b76c6be17aa95ed2e793f6235148f83ef/pkg/cmd/dedicated/listclusters/list_clusters.go#L116 (even though this can still be removed with what you current have) or https://github.com/redhat-developer/app-services-cli/blob/bf1a206b76c6be17aa95ed2e793f6235148f83ef/pkg/cmd/dedicated/listclusters/list_clusters.go#L145

machi1990: When you use the search suggestion - you'll then make one call to ocm, the search parameter and the `opts.limit` to control the bucket size will then be set to `len(opts.kfmClusterList.Items)` the length of the items returned by registered enterprise clusters.

dimakis: Well that is a much better approach, will sort this 👍

pkg/cmd/dedicated/register/registercluster.go
```diff
@@ -373,7 +373,7 @@ func registerClusterWithKasFleetManager(opts *options) error {
 	if err != nil {
 		return err
 	}
-	err = clustermgmt.CreateAddonWithParams(opts.f, opts.clusterManagementApiUrl, opts.accessToken, getStrimziAddonIdByEnv(con), response.FleetshardParameters, opts.selectedCluster.ID())
+	err = clustermgmt.CreateAddonWithParams(opts.f, opts.clusterManagementApiUrl, opts.accessToken, getStrimziAddonIdByEnv(con), nil, opts.selectedCluster.ID())
```
machi1990, line 376: hmm this was done in https://github.com/redhat-developer/app-services-cli/pull/1815 ? A rebase maybe ;-)

dimakis: Yes, it was. Great catch 👍

pkg/cmd/dedicated/listclusters/list_clusters.go
```diff
@@ -0,0 +1,176 @@
...
+	return cls
+}
+
+func mapRegisteredClusters(opts *options) error {
+	for _, cluster := range opts.clustermgmtClusterList {
+		// if cluster is registered, add it to the list of registered clusters
+		if isRegistered, kfmcluster := isClusterRegistered(&cluster, &opts.kfmClusterList); isRegistered {
+			opts.registeredClusters = append(opts.registeredClusters, clusterRow{
+				ID:            cluster.ID(),
+				Name:          cluster.Name(),
```
machi1990, line 131:
```suggestion
				Name:          cluster.DisplayName(),
```

The `.name` field is static and creation time only. The display name can be changed by the user and we should be using this one instead. Can we verify other location as well and update accordingly?

machi1990: e.g https://github.com/redhat-developer/app-services-cli/blob/bf1a206b76c6be17aa95ed2e793f6235148f83ef/pkg/cmd/dedicated/register/registercluster.go#L144 but in a separate PR to this one

pkg/cmd/dedicated/listclusters/list_clusters.go
```diff
@@ -0,0 +1,176 @@
...
+	// show limit or total
+	if opts.limit > len(opts.registeredClusters) {
+		lmt = localize.NewEntry("CCClusterLimit", len(opts.registeredClusters))
+	}
+	opts.f.Logger.Info(opts.f.Localizer.MustLocalize("dedicated.listClusters.common.limitTotal", lmt, totalClusters))
+	start := (opts.page - 1) * opts.limit
+	end := start + len(opts.registeredClusters)
+	opts.f.Logger.Info(fmt.Sprintf("[%v %v : %v - %v]", opts.f.Localizer.MustLocalize("dedicated.listClusters.common.page"), opts.page, start, end))
+	return nil
+}
```
machi1990, line 176: Let's investigate displaying the output using the existing options already used in the CLI e.g https://github.com/redhat-developer/app-services-cli/blob/bf1a206b76c6be17aa95ed2e793f6235148f83ef/pkg/cmd/kafka/list/list.go#L129-L150

machi1990: Notably line 146 or line 149

pkg/cmd/dedicated/listclusters/list_clusters.go
dimakis: DisplayName is not in the OCM SDK as discussed 👍

pkg/cmd/dedicated/listclusters/list.go
```diff
@@ -0,0 +1,137 @@
...
+	return nil
+}
+
+func createSearchString(opts *options) string {
+	searchString := ""
+	for idx, kfmcluster := range opts.kfmClusterList.Items {
+		searchString += fmt.Sprintf("id = '%s'", kfmcluster.Id)
+		if idx > 0 {
+			searchString += " or "
+		}
```
machi1990, line 97: What happens if there are two clusters? Will this construct a valid search string?

pkg/cmd/dedicated/listclusters/list.go
```diff
@@ -91,10 +91,10 @@ func listEnterpriseClusters(opts *options, f *factory.Factory) error {
 func createSearchString(opts *options) string {
 	searchString := ""
 	for idx, kfmcluster := range opts.kfmClusterList.Items {
-		searchString += fmt.Sprintf("id = '%s'", kfmcluster.Id)
-		if idx > 0 {
+		if idx >= 1 && idx < len(opts.kfmClusterList.Items) {
```
machi1990, line 94: will `idx` be ever equal to `len(opts.kfmClusterList.Items)`?

pkg/cmd/dedicated/listclusters/list_test.go
```diff
@@ -0,0 +1,59 @@
+package listclusters
+
+import (
+	kafkamgmtclient "github.com/redhat-developer/app-services-sdk-go/kafkamgmt/apiv1/client"
+	"testing"
+)
+
+func TestCreateSearchString(t *testing.T) {
```
machi1990, line 8: <3

pkg/cmd/dedicated/listclusters/list.go
dimakis: Nope. Damn this is amateur hour over here

dimakis: for idx == 0 and `len(opts.kfmClusterList.Items)` == 0 🤔

dimakis: but in that case there are no items to range over so the loop never gets executed

pkg/cmd/dedicated/listclusters/list_test.go
dimakis: I thought this might make you happy 😄

pkg/cmd/dedicated/listclusters/list.go
machi1990: Yes. I wonder the usefulness of that check length check then.

Good thing is you've test - let's see what they say :-D

pkg/cmd/dedicated/listclusters/list_test.go
machi1990: :100:

pkg/cmd/dedicated/listclusters/list.go
```diff
@@ -0,0 +1,137 @@
...
+				crl = append(crl, clusterRow{
+					Name:          ocmCluster.Name(),
+					ID:            kfmcluster.Id,
+					Status:        *kfmcluster.Status,
+					CloudProvider: ocmCluster.CloudProvider().ID(),
+					Region:        ocmCluster.Region().ID(),
+				})
+			}
+		}
+	}
```
machi1990, line 125: nitpick: consider building a map of `cluster_id ->  clustermgmtCluster` to improve search and get rid of the second loop .
Instead of line 114 you'll the have 
```go
ocmCluster := theMap[kfmcluster.Id]
crl = append(crl, clusterRow{
					Name:          ocmCluster.Name(),
					ID:            kfmcluster.Id,
					Status:        *kfmcluster.Status,
					CloudProvider: ocmCluster.CloudProvider().ID(),
					Region:        ocmCluster.Region().ID(),
				})
```

pkg/cmd/dedicated/listclusters/list.go
```diff
@@ -0,0 +1,137 @@
...
+	ctx := context.Background()
+	api := conn.API()
+	cl := api.KafkaMgmtEnterprise().GetEnterpriseOsdClusters(ctx)
+	clist, response, err := cl.Execute()
+	if err != nil {
+		return err
+	}
+	opts.kfmClusterList = &clist
+	f.Logger.Info(response)
```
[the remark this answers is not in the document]
machi1990: let's remove this as it gets in the way of the printed clusters table @dimakis

pkg/cmd/dedicated/listclusters/list.go
```diff
@@ -0,0 +1,140 @@
...
+	if len(opts.kfmClusterList.Items) == 0 {
+		return f.Localizer.MustLocalizeError("dedicated.listClusters.log.info.noClusters")
```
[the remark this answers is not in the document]
machi1990: The `dedicated.listClusters.log.info.noClusters` message doesn't exist

machi1990 (approved):
lgtm once the suggestions are implemented

pkg/cmd/dedicated/listclusters/list.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return err
+	}
+	opts.clustermgmtClusterList = clist
+	opts.registeredClusters = kfmListToClusterRowList(opts)
+	err = displayRegisteredClusters(opts)
+	if err != nil {
+		return err
+	}
+	return nil
```
machi1990, line 68:
```suggestion
	return displayRegisteredClusters(opts)
```

pkg/cmd/dedicated/listclusters/list.go
```diff
@@ -0,0 +1,132 @@
...
+	ID            string `json:"id" header:"ID"`
+	Status        string `json:"status" header:"Status"`
+	CloudProvider string `json:"cloud_provider" header:"Cloud Provider"`
+	Region        string `json:"region" header:"Region"`
+}
+
+type options struct {
+	outputFormat            string
+	search                  string
+	kfmClusterList          *kafkamgmtclient.EnterpriseClusterList
```
machi1990, line 25:
```suggestion
```

Are these two used anywhere?

pkg/cmd/dedicated/listclusters/list.go
```diff
@@ -0,0 +1,132 @@
...
+type clusterRow struct {
+	Name          string `json:"name" header:"Name"`
+	ID            string `json:"id" header:"ID"`
+	Status        string `json:"status" header:"Status"`
+	CloudProvider string `json:"cloud_provider" header:"Cloud Provider"`
+	Region        string `json:"region" header:"Region"`
+}
+
+type options struct {
+	outputFormat            string
```
machi1990, line 23: Might be good to allow configuring the output format

machi1990: In the meantime this is not used

dimakis: Yes, I started with the option to format the output in json/yml etc. 
obviously I should have cleaned this up once I changed the functionality

pkg/cmd/dedicated/listclusters/list.go
dimakis: kfmClusterlist is used as the list of enterprise clusters returned by the KFM call

PR #1818, opened 2023-02-28
refactor: moving list enterprise clusters to util package

dimakis:
Moving this out for easy resuse

Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/shared/kafkautil/dedicated_utils.go
+++ b/pkg/shared/kafkautil/dedicated_utils.go
@@ -1,0 +1,26 @@
+package kafkautil
+
+import (
+	"context"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/factory"
+	kafkamgmtclient "github.com/redhat-developer/app-services-sdk-core/app-services-sdk-go/kafkamgmt/apiv1/client"
+)
+
+func ListEnterpriseClusters(f *factory.Factory) (*kafkamgmtclient.EnterpriseClusterList, error) {
+	conn, err := f.Connection()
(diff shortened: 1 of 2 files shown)


pkg/shared/kafkautil/dedicated_utils.go
```diff
@@ -0,0 +1,26 @@
+package kafkautil
+
+import (
+	"context"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/factory"
+	kafkamgmtclient "github.com/redhat-developer/app-services-sdk-core/app-services-sdk-go/kafkamgmt/apiv1/client"
+)
+
+func ListEnterpriseClusters(f *factory.Factory) (*kafkamgmtclient.EnterpriseClusterList, error) {
```
rkpattnaik780, line 9: `shared utils` is a place where methods defined are used in different subcommands. For example - `kafkautil.InteractiveSelect` is used in `kafka`, `connectors` and `cluster` subcommands.
`ListEnterpriseClusters` is used just in dedicated list, will it make sense to move it to `pkg/cmd/dedicated/dedicatedcmdutil/dedicated_util.go`?

dimakis: it will also be used in the `create kafka` cmd too, so that's the reason I moved it to shared

jackdelahunt (approved):

rkpattnaik780 (approved):

PR #1824, opened 2023-03-01
Enterprise create kafka interactive refactor2

dimakis:
Some refactoring of the create kafka cmd.

interactive prompt for cluster selection now shows name and ID. 

flags still need to be updated i.e. if flags are passed skip the interactive logic

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -3,6 +3,9 @@ package create
 import (
 	"encoding/json"
 	"fmt"
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/dedicated/listclusters"
+	"github.com/redhat-developer/app-services-cli/pkg/shared/connection/api/clustermgmt"
 	"github.com/redhat-developer/app-services-cli/pkg/shared/kafkautil"
 	"os"
 	"os/signal"
@@ -70,6 +73,7 @@ type options struct {
 
 	kfmClusterList       *kafkamgmtclient.EnterpriseClusterList
 	selectedCluster      *kafkamgmtclient.EnterpriseClusterListItem
+	clusterMap           *map[string]v1.Cluster
 	useEnterpriseCluster bool
 
 	f *factory.Factory
@@ -460,15 +464,13 @@ func setEnterpriseClusterList(opts *options) error {
 		return err
 	}
 	opts.kfmClusterList = clist
-	return nil
-}
 
-func createPromptSliceFromEnterpriseClusterList(clusterList *kafkamgmtclient.EnterpriseClusterList) []string {
-	clusterListSlice := make([]string, 0)
-	for _, cluster := range clusterList.Items {
-		clusterListSlice = append(clusterListSlice, *cluster.ClusterId)
+	clusterMap, err := getClusterNameMap(opts.f, clist)
+	if err != nil {
+		return err
 	}
-	return clusterListSlice
+	opts.clusterMap = clusterMap
+	return nil
 }
 
 func selectEnterpriseOrRHInfraPrompt(opts *options) error {
@@ -516,7 +518,7 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 		return nil, err
 	}
 
-	// Get the list of enterprise clusters in the users organization if there are any
+	// Get the list of enterprise clusters in the users organization if there are any, creates a map of cluster ids to names to include names in the prompt
 	err = setEnterpriseClusterList(opts)
 	if err != nil {
 		return nil, err
@@ -556,6 +558,7 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 		return nil, opts.f.Localizer.MustLocalizeError("kafka.create.provider.error.noStandardInstancesAvailable")
 	}
 
+	// prompting for billing model if there are more than one available, otherwise using the only one available
 	if len(availableBillingModels) > 0 {
 		if len(availableBillingModels) == 1 {
 			answers.BillingModel = availableBillingModels[0]
@@ -571,17 +574,12 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 		}
 	}
 
+	// if billing model is marketplace, prompt for marketplace provider and account id
 	if answers.BillingModel == accountmgmtutil.QuotaMarketplaceType {
 		answers, err = marketplaceQuotaPrompt(orgQuota, answers, f, err)
 		if err != nil {
 			return nil, err
 		}
-	}
-
-	marketplaceInfo := accountmgmtutil.MarketplaceInfo{
-		BillingModel:   answers.BillingModel,
-		Provider:       answers.Marketplace,
-		CloudAccountID: answers.MarketplaceAcctID,
 	}
 
 	var sizes []kafkamgmtclient.SupportedKafkaSize
@@ -630,6 +628,7 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 		var clusterIdSNS kafkamgmtclient.NullableString
 		clusterIdSNS.Set(opts.selectedCluster.ClusterId)
 
+		// need a better way to set billing model
 		var billingModelEnterprise kafkamgmtclient.NullableString
 		var enterprise = "enterprise"
 		billingModelEnterprise.Set(&enterprise)
@@ -641,11 +640,18 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 			BillingCloudAccountId: nullString,
 			Marketplace:           nullString,
 			ClusterId:             clusterIdSNS,
-			//	set plan here
 		}
 		// enterprise quota spec, kfm will use the default, this should be returned by the `select quota` call
 		payload.SetPlan(mapAmsTypeToBackendType(&orgQuota.EnterpriseQuotas[0]) + "." + answers.Size)
 	} else {
+
+		// This flow is for the provisioning of a standard kafka instance
+		marketplaceInfo := accountmgmtutil.MarketplaceInfo{
+			BillingModel:   answers.BillingModel,
+			Provider:       answers.Marketplace,
+			CloudAccountID: answers.MarketplaceAcctID,
+		}
+
 		userQuota, err := accountmgmtutil.SelectQuotaForUser(f, orgQuota, marketplaceInfo, answers.CloudProvider)
 		if err != nil {
 			return nil, err
@@ -773,11 +779,51 @@ func cloudProviderPrompt(f *factory.Factory, answers *promptAnswers) (*promptAns
 	return answers, nil
 }
 
+// This may need to altered as the `name` are mutable on ocm side
+func getClusterNameMap(f *factory.Factory, clusterList *kafkamgmtclient.EnterpriseClusterList) (*map[string]v1.Cluster, error) {
+	//	for each cluster in the list, get the name from ocm and add it to the cluster list
+	str := listclusters.CreateSearchString(clusterList)
+	ocmClusterList, err := clustermgmt.GetClusterListByIds(f, "", "", str, len(clusterList.Items))
+	if err != nil {
+		return nil, err
+	}
+	clusterMap := make(map[string]v1.Cluster)
+	for _, cluster := range ocmClusterList.Slice() {
+		clusterMap[cluster.ID()] = *cluster
+	}
+	return &clusterMap, nil
+
+}
+
+func validateClusters(clusterList *kafkamgmtclient.EnterpriseClusterList) *kafkamgmtclient.EnterpriseClusterList {
+	// if cluster is in a ready state add it to the list of clusters
+	returnList := kafkamgmtclient.EnterpriseClusterList{}
+	for _, cluster := range clusterList.Items {
+		if *cluster.Status == "ready" {
+			returnList.Items = append(returnList.Items, cluster)
+		}
+	}
+	// TO-DO add logic to return cluster capacity
+
+	return &returnList
+}
+
+func createPromptOptionsFromClusters(clusterList *kafkamgmtclient.EnterpriseClusterList, clusterMap *map[string]v1.Cluster) *[]string {
+	promptOptions := []string{}
+	validatedClusters := validateClusters(clusterList)
+	for _, cluster := range validatedClusters.Items {
+		ocmCluster := (*clusterMap)[cluster.GetId()]
+		display := ocmCluster.Name() + " (" + cluster.GetId() + ")"
+		promptOptions = append(promptOptions, display)
+	}
+	return &promptOptions
+}
+
 func selectClusterPrompt(opts *options) error {
-	promptOptions := createPromptSliceFromEnterpriseClusterList(opts.kfmClusterList)
+	promptOptions := createPromptOptionsFromClusters(opts.kfmClusterList, opts.clusterMap)
 	promptForCluster := &survey.Select{
 		Message: opts.f.Localizer.MustLocalize("kafka.create.input.cluster.selectClusterMessage"),
-		Options: promptOptions,
+		Options: *promptOptions,
 	}
 
 	var index int
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -612,13 +612,13 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 			}
 		}
 	}
-	
+
 	marketplaceInfo := accountmgmtutil.MarketplaceInfo{
 		BillingModel:   answers.BillingModel,
 		Provider:       answers.Marketplace,
 		CloudAccountID: answers.MarketplaceAcctID,
 	}
-	
+
 	var sizes []kafkamgmtclient.SupportedKafkaSize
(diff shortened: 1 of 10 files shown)


pkg/cmd/kafka/create/create.go
```diff
@@ -579,48 +574,84 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+
+		if selectedSize.GetCapacityConsumed() > cluster.CapacityInformation.RemainingKafkaStreamingUnits {
+			return nil, opts.f.Localizer.MustLocalizeError("kafka.create.cluster.error.noEnoughCapacity", localize.NewEntry("Size", selectedSize.GetId()))
+		}
+
+		answers.Size = selectedSize.GetId()
+
+		if len(orgQuota.EnterpriseQuotas) == 0 {
+			return nil, fmt.Errorf("No enterprise quota available")
+		}
```
machi1990, line 626: This is not checking if the user has remaining quota, it just check if the user at some point had enterprise quota. 
We should improve this onto checking for `remaining quota >= size.quota_consumed`.

pkg/cmd/kafka/create/create.go
```diff
@@ -579,48 +574,84 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
-	}
 
-	marketplaceInfo := accountmgmtutil.MarketplaceInfo{
-		BillingModel:   answers.BillingModel,
-		Provider:       answers.Marketplace,
-		CloudAccountID: answers.MarketplaceAcctID,
-	}
+		sizes = supportedKafkaTypes[0].Sizes
+
+		index, err := promptUserForSizes(opts.f, &sizes)
```
machi1990, line 611: no need to prompt for size if `opts.size` the flag has been set

pkg/cmd/kafka/create/create.go
```diff
@@ -509,15 +511,22 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
-	err := setEnterpriseClusterList(opts)
+	answers := &promptAnswers{}
+
+	err := survey.AskOne(promptName, &answers.Name, survey.WithValidator(validator.ValidateName), survey.WithValidator(validator.ValidateNameIsAvailable))
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the list of enterprise clusters in the users organization if there are any, creates a map of cluster ids to names to include names in the prompt
+	err = setEnterpriseClusterList(opts)
```
machi1990, line 522: no need to prompt for cluster if opts.cluster_id has been set

machi1990, line 522: The https://github.com/redhat-developer/app-services-cli/blob/d9826b3b71bf87661a7a5cd04d842d50029dfb68/pkg/cmd/kafka/create/create.go#L460 needs to handle the `403` error being returned when making a call to KFM to list the clusters.
The error could be returned when the user doesn't not have quota. In that case, we should propose the normal flow as well.

pkg/cmd/kafka/create/create.go
dimakis: but it is getting it from here https://github.com/dimakis/app-services-cli/blob/d9826b3b71bf87661a7a5cd04d842d50029dfb68/pkg/shared/accountmgmtutil/ams.go#L111
this is where the quota is checked. makes sense?

dimakis: and it is called in create here: https://github.com/dimakis/app-services-cli/blob/d9826b3b71bf87661a7a5cd04d842d50029dfb68/pkg/cmd/kafka/create/create.go#L550

machi1990: All that is fine, but what's missing is checking that the user has truly remaining quota. I'll explain myself

> but it is getting it from here https://github.com/dimakis/app-services-cli/blob/d9826b3b71bf87661a7a5cd04d842d50029dfb68/pkg/shared/accountmgmtutil/ams.go#L111
this is where the quota is checked. makes sense?

It could be that in the `remainingQuota` is `0`. I do not see anywhere in the code where we check for this and we'll endup attempting to create a Kafka even if there is no remaining quota.

pkg/cmd/kafka/create/create.go
machi1990, line 611: this prompt shows the following label
```
? Instance type:  [Use arrows to move, type to filter]
  x1
> x2
```
Should it be `Instance size` instead?

pkg/cmd/kafka/create/create.go
machi1990, line 626: I am getting 
```
 No enterprise quota available.
```

All the time when I try to create a Kafka, is that expected?

pkg/cmd/kafka/create/create.go
```diff
@@ -579,48 +574,84 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+		if selectedSize.GetCapacityConsumed() > cluster.CapacityInformation.RemainingKafkaStreamingUnits {
+			return nil, opts.f.Localizer.MustLocalizeError("kafka.create.cluster.error.noEnoughCapacity", localize.NewEntry("Size", selectedSize.GetId()))
+		}
+
+		answers.Size = selectedSize.GetId()
+
+		if len(orgQuota.EnterpriseQuotas) == 0 {
+			return nil, fmt.Errorf("No enterprise quota available")
+		}
+		// casting to nullable string
```
andreaTP, line 627: I suspect that this is coming from the generated SDK. This code looks a bit wired, would it be possible to refactor the creation of `KafkaRequestPayload` in a Factory method?

pkg/cmd/kafka/create/create.go
jackdelahunt: Will do

jackdelahunt: Yes

pkg/cmd/kafka/create/create.go
```diff
@@ -579,48 +574,84 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+		var nullString kafkamgmtclient.NullableString
+		payload = &kafkamgmtclient.KafkaRequestPayload{
+			Name:                  answers.Name,
+			BillingModel:          billingModelEnterprise,
+			BillingCloudAccountId: nullString,
+			Marketplace:           nullString,
+			ClusterId:             clusterIdSNS,
+		}
+		// enterprise quota spec, kfm will use the default, this should be returned by the `select quota` call
+		payload.SetPlan(mapAmsTypeToBackendType(&orgQuota.EnterpriseQuotas[0]) + "." + answers.Size)
```
andreaTP, line 645: Please extract the selection of `&orgQuota.EnterpriseQuotas[0]` before this line:
https://github.com/redhat-developer/app-services-cli/pull/1824/files#diff-5a5c1a033aadb8090e822a7ebdb919ec8ecbd989c46f5e84047ea197aa987293R609

i.e. closer to this meaningful comment:
https://github.com/redhat-developer/app-services-cli/pull/1824/files#diff-5a5c1a033aadb8090e822a7ebdb919ec8ecbd989c46f5e84047ea197aa987293R603-R604

So that we don't forget about updating the index when the limitation got removed.

pkg/cmd/kafka/create/create.go
```diff
@@ -653,26 +684,18 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
-			planPrompt := &survey.Select{
-				Message: f.Localizer.MustLocalize("kafka.create.input.plan.message"),
-				Options: sizeLabels,
-			}
-
-			err = survey.AskOne(planPrompt, &answers.Size)
-			if err != nil {
-				return nil, err
-			}
+		index, err := promptUserForSizes(opts.f, &sizes)
```
andreaTP, line 692: Should we skip the prompt if it has been set through the options?

pkg/cmd/kafka/create/create.go
```diff
@@ -700,35 +723,107 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+				return nil, err
+			}
 		}
 	}
-	return payload, nil
+	return answers, nil
+}
+
+func cloudProviderPrompt(f *factory.Factory, answers *promptAnswers) (*promptAnswers, error) {
+	cloudProviderNames, err := GetEnabledCloudProviderNames(f)
```
andreaTP, line 765: Should we check if there is only one choice and avoid the following prompt?

pkg/cmd/kafka/create/create.go
```diff
@@ -700,35 +723,107 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+	}
+
+	err = survey.AskOne(cloudProviderPrompt, &answers.CloudProvider)
+	if err != nil {
+		return nil, err
+	}
+	return answers, nil
+}
+
+// This may need to altered as the `name` are mutable on ocm side
```
andreaTP, line 782: is this comment updated? I see that we are using `ID`

pkg/cmd/kafka/create/create.go
```diff
@@ -700,35 +723,107 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+
+func validateClusters(clusterList *kafkamgmtclient.EnterpriseClusterList) *kafkamgmtclient.EnterpriseClusterList {
+	// if cluster is in a ready state add it to the list of clusters
+	returnList := kafkamgmtclient.EnterpriseClusterList{}
+	for _, cluster := range clusterList.Items {
+		if *cluster.Status == "ready" {
+			returnList.Items = append(returnList.Items, cluster)
+		}
+	}
+	// TO-DO add logic to return cluster capacity
```
andreaTP, line 806: Should we target this TODO in the context of this PR?

pkg/cmd/kafka/create/create.go
```diff
@@ -700,35 +723,107 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+
+	return &returnList
+}
+
+func createPromptOptionsFromClusters(clusterList *kafkamgmtclient.EnterpriseClusterList, clusterMap *map[string]v1.Cluster) *[]string {
+	promptOptions := []string{}
+	validatedClusters := validateClusters(clusterList)
+	for _, cluster := range validatedClusters.Items {
+		ocmCluster := (*clusterMap)[cluster.GetId()]
```
andreaTP, line 815: [nitpick] those lines can be easily refactored to an external function, probably the code will become a little better readable.

andreaTP (commented):
Thanks! Nice work! Left a few comments around 🙂

pkg/shared/remote/service-constants.json
```diff
@@ -9,7 +9,8 @@
       "quotaProductId": "RHOSAK",
       "trialQuotaProductId": "RHOSAKTrial",
       "longLivedQuotaProductId": "RHOSAKEval",
-      "resourceName": "rhosak"
+      "resourceName": "rhosak",
+      "enterpriseProductQuotaId": "RHOSAKCC"
```
machi1990, line 13: We'll also need to add this constant here https://github.com/redhat-developer/app-services-ui/blob/main/static/configs/service-constants.json

pkg/cmd/kafka/create/create.go
machi1990: This is due to https://github.com/redhat-developer/app-services-cli/pull/1824#discussion_r1122953539

Until that's progressed, I guess @dimakis has a workaround for hardcoding the productID

pkg/cmd/kafka/create/create.go
jackdelahunt: So I was adding this and discovered that we didn't know exactly how the command handles flags. From what I can see when adding a flag like `size` or `name` the command won't just skip prompts in interactive mode, it skips interactive mode all together. Meaning there is no need t handle it in this case as above ^^

pkg/cmd/kafka/create/create.go
jackdelahunt: This is happening when we are in interactive mode meaning the `cluster_id` has not been set anyway

pkg/cmd/kafka/create/create.go
```diff
@@ -579,48 +574,84 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+		sizes = supportedKafkaTypes[0].Sizes
+
+		index, err := promptUserForSizes(opts.f, &sizes)
+		if err != nil {
+			return nil, err
+		}
+
+		selectedSize := sizes[index]
+
+		if selectedSize.GetCapacityConsumed() > cluster.CapacityInformation.RemainingKafkaStreamingUnits {
```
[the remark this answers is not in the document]
jackdelahunt: I thought it should be `>` if there is an equal amount of streaming units to the capacity consumed for the size then it should be okay to use. Maybe I am wrong though

pkg/cmd/kafka/create/create.go
jackdelahunt: From how other options work, if any option is set we skip the interactive selection and make the payload from that meaning because we are currently in interactive mode we know size has not been set

pkg/cmd/kafka/create/create.go
```diff
@@ -700,35 +723,107 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+func getClusterNameMap(f *factory.Factory, clusterList *kafkamgmtclient.EnterpriseClusterList) (*map[string]v1.Cluster, error) {
+	//	for each cluster in the list, get the name from ocm and add it to the cluster list
+	str := listclusters.CreateSearchString(clusterList)
+	ocmClusterList, err := clustermgmt.GetClusterListByIds(f, "", "", str, len(clusterList.Items))
+	if err != nil {
+		return nil, err
+	}
+	clusterMap := make(map[string]v1.Cluster)
+	for _, cluster := range ocmClusterList.Slice() {
+		clusterMap[cluster.ID()] = *cluster
```
jackdelahunt, line 792: Isn't this supposed to be name?

pkg/cmd/kafka/create/create.go
andreaTP: 👍  fair!

pkg/cmd/kafka/create/create.go
dimakis: nope, the name is only coming from the ocm side, ID is common between both and immutable

pkg/cmd/kafka/create/create.go
dimakis: this was a question I wanted to ask a little more broadly, to determine capacity for each cluster we've got to do a seperate call to ocm for each cluster, to the best of my knowledge. 
is this overkill? I mean I know it would be nice to see this info here but I was just wondering?

pkg/cmd/kafka/create/create.go
dimakis: we are using both, the cluster name and the cluster id in this format:
`clustername (clusterId)`
The issue is that ocm name is a mutable field, however, we're doing the same thing that console.dot and the ocm cli do

pkg/cmd/kafka/create/create.go
```diff
@@ -579,48 +574,84 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
 
-			validMarketplaceAcctIDs := FetchValidMarketplaceAccounts(orgQuota.MarketplaceQuotas, answers.Marketplace)
+	if opts.useEnterpriseCluster {
+		/*
+			if using dedicated cluster option then get the supported sizes for this cluster
+			based on the sizes it supports the user to select the one they want to use
+			then check if the size uses <= the remaining streaming units it takes up
+			on the cluster
+		*/
+		cluster, err := getClusterDetails(opts.f, opts.selectedCluster.GetId())
```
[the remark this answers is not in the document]
dimakis: so here, its everytime someone goes to select a hybrid (enterprise) cluster to provision a kafka on. The `getClusterDetails` returns capacity information about that cluster i.e. what is the cluster's capacity for the  provisioning of a kafka. 
This information should be fresh I believe.

pkg/cmd/kafka/create/create.go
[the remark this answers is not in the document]
jackdelahunt: Talked with @dimakis and we have confirmed the `>` is the correct check for this.

pkg/cmd/kafka/create/create.go
jackdelahunt: Done

pkg/cmd/kafka/create/create.go
jackdelahunt: Done

pkg/cmd/kafka/create/create.go
```diff
@@ -579,48 +574,84 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
...
+
+		selectedSize := sizes[index]
+
+		if selectedSize.GetCapacityConsumed() > cluster.CapacityInformation.RemainingKafkaStreamingUnits {
+			return nil, opts.f.Localizer.MustLocalizeError("kafka.create.cluster.error.noEnoughCapacity", localize.NewEntry("Size", selectedSize.GetId()))
+		}
+
+		answers.Size = selectedSize.GetId()
+
+		if len(orgQuota.EnterpriseQuotas) == 0 {
```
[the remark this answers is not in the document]
jackdelahunt: Done

PR #1825, opened 2023-03-03
feat: added name of cluster to kafka list

jackdelahunt:
### Verification Steps

1. Create a cluster that is on a dedicated cluster
2. Run `rhoas kafka list`
3. See the cluster colom now contains the name and id of the cluster

 
image

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/list/list.go
+++ b/pkg/cmd/kafka/list/list.go
@@ -50,10 +50,7 @@ type options struct {
 // NewListCommand creates a new command for listing kafkas.
 func NewListCommand(f *factory.Factory) *cobra.Command {
 	opts := &options{
-		page:   0,
-		limit:  100,
-		search: "",
-		f:      f,
+		f: f,
 	}
 
 	cmd := &cobra.Command{
@@ -83,7 +80,7 @@ func NewListCommand(f *factory.Factory) *cobra.Command {
 
 	flags.AddOutput(&opts.outputFormat)
 	flags.IntVar(&opts.page, "page", int(cmdutil.ConvertPageValueToInt32(build.DefaultPageNumber)), opts.f.Localizer.MustLocalize("kafka.list.flag.page"))
-	flags.IntVar(&opts.limit, "limit", 100, opts.f.Localizer.MustLocalize("kafka.list.flag.limit"))
+	flags.IntVar(&opts.limit, "limit", int(cmdutil.ConvertPageValueToInt32(build.DefaultPageSize)), opts.f.Localizer.MustLocalize("kafka.list.flag.limit"))
 	flags.StringVar(&opts.search, "search", "", opts.f.Localizer.MustLocalize("kafka.list.flag.search"))
 	flags.StringVar(&opts.clusterManagementApiUrl, "cluster-mgmt-api-url", "", f.Localizer.MustLocalize("dedicated.registerCluster.flag.clusterMgmtApiUrl.description"))
 	flags.StringVar(&opts.accessToken, "access-token", "", f.Localizer.MustLocalize("dedicated.registercluster.flag.accessToken.description"))
@@ -141,9 +138,9 @@ func runList(opts *options) error {
 		}
 
 		if currCtx.KafkaID != "" {
-			rows = mapResponseItemsToRows(response.GetItems(), currCtx.KafkaID, &clusterIdMap)
+			rows = mapResponseItemsToRows(opts, response.GetItems(), currCtx.KafkaID, &clusterIdMap)
 		} else {
-			rows = mapResponseItemsToRows(response.GetItems(), "-", &clusterIdMap)
+			rows = mapResponseItemsToRows(opts, response.GetItems(), "-", &clusterIdMap)
 		}
 		dump.Table(opts.f.IOStreams.Out, rows)
 		opts.f.Logger.Info("")
@@ -153,7 +150,7 @@ func runList(opts *options) error {
 	return nil
 }
 
-func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId string, clusterIdMap *map[string]*v1.Cluster) []kafkaRow {
+func mapResponseItemsToRows(opts *options, kafkas []kafkamgmtclient.KafkaRequest, selectedId string, clusterIdMap *map[string]*v1.Cluster) []kafkaRow {
 	rows := make([]kafkaRow, len(kafkas))
 
 	for i := range kafkas {
@@ -168,7 +165,7 @@ func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId st
 			cluster := (*clusterIdMap)[*id]
 			customerCloud = fmt.Sprintf("%v (%v)", cluster.Name(), cluster.ID())
 		} else {
-			customerCloud = "Red Hat Infrastructure"
+			customerCloud = opts.f.Localizer.MustLocalize("kafka.list.output.customerCloud.redhat")
 		}
 
 		row := kafkaRow{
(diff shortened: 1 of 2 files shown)


pkg/cmd/kafka/list/list.go
```diff
@@ -185,6 +180,43 @@ func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId st
...
+			kafkaClusterIds = append(kafkaClusterIds, kafka.GetClusterId())
+		}
+	}
+
+	idToCluster := make(map[string]*v1.Cluster)
+
+	// if no kafkas have a cluster id assigned then we can skip the call to get
+	// the clusters as we dont need their info
+	if len(kafkaClusterIds) > 0 {
+		clusterList, err := clustermgmt.GetClusterListByIds(opts.f, "", "", createSearchString(kafkaClusterIds), 100)
```
jackdelahunt, line 196: Two questions:

1. Do we add a cluster mgmt url to `rhoas kafka list` as a flag and the access token
2. To know the size of the request needed we need to make another api call, should this be done?

machi1990: > Do we add a cluster mgmt url to rhoas kafka list as a flag and the access token

Yes please. These flags are to be hidden.

machi1990: > To know the size of the request needed we need to make another api call, should this be done?

No extra API call is needed.

The size should be at maximum the length of the `kafkaClusterIds` slice - we could use this instead of hardcoding 100.

machi1990, line 196:
```suggestion
		clusterList, err := clustermgmt.GetClusterListByIds(opts.f, "", "", createSearchString(kafkaClusterIds), len(kafkaClusterIds))
```

pkg/cmd/kafka/list/list.go
```diff
@@ -185,6 +180,43 @@ func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId st
 	return rows
 }
 
+func getClusterIdMapFromKafkas(opts *options, kafkas []kafkamgmtclient.KafkaRequest) (map[string]*v1.Cluster, error) {
+	kafkaClusterIds := make([]string, len(kafkas))
+	for _, kafka := range kafkas {
+		if kafka.GetClusterId() != "" {
+			kafkaClusterIds = append(kafkaClusterIds, kafka.GetClusterId())
+		}
+	}
```
machi1990, line 189: You could further improve this by making sure that the IDs are unique.. but it is good enough already

pkg/cmd/kafka/list/list.go
jackdelahunt: > Yes please. Additionally, these flags are to be hidden.

Agh okay nice, I thought it might be strange but that fact they are hidden is okay

machi1990: @jackdelahunt This was marked as resolved but I am still seeing the hardcoded 100. Is that expected?

pkg/cmd/kafka/list/list.go
jackdelahunt: I tried to think about this, how would you make sure the IDs are unique in a way that would be faster then what happens now which is the chance to add the same id and insert the same key, value pair into the map?

I tried to think of a solution but I am not sure.

jackdelahunt:
Changes suggested by Manyanda are applied 👍

pkg/cmd/kafka/list/list.go
machi1990: > value pair into the map?

a `Set` data type is what would have been suitable here. But golang doesn't have that and a map can be used instead.  IM0, what you've now is okay as at the API is capable of handling that via the search query that's constructed.

pkg/cmd/kafka/list/list.go
```diff
@@ -185,6 +187,43 @@ func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId st
...
+
+	idToCluster := make(map[string]*v1.Cluster)
+
+	// if no kafkas have a cluster id assigned then we can skip the call to get
+	// the clusters as we dont need their info
+	if len(kafkaClusterIds) > 0 {
+		clusterList, err := clustermgmt.GetClusterListByIds(opts.f, opts.clusterManagementApiUrl, opts.accessToken, createSearchString(kafkaClusterIds), len(idToCluster))
```
machi1990, line 203:
```suggestion
		clusterList, err := clustermgmt.GetClusterListByIds(opts.f, opts.clusterManagementApiUrl, opts.accessToken, createSearchString(kafkaClusterIds), len(kafkaClusterIds))
```

`len(idToCluster)` is `0` at this stage as it is set afterward post this call.

pkg/cmd/kafka/list/list.go
```diff
@@ -88,31 +79,36 @@ func NewListCommand(f *factory.Factory) *cobra.Command {
...
 
-	flags := kafkaFlagutil.NewFlagSet(cmd, opts.localizer)
+	flags := kafkaFlagutil.NewFlagSet(cmd, opts.f.Localizer)
 
 	flags.AddOutput(&opts.outputFormat)
-	flags.IntVar(&opts.page, "page", int(cmdutil.ConvertPageValueToInt32(build.DefaultPageNumber)), opts.localizer.MustLocalize("kafka.list.flag.page"))
-	flags.IntVar(&opts.limit, "limit", 100, opts.localizer.MustLocalize("kafka.list.flag.limit"))
-	flags.StringVar(&opts.search, "search", "", opts.localizer.MustLocalize("kafka.list.flag.search"))
+	flags.IntVar(&opts.page, "page", int(cmdutil.ConvertPageValueToInt32(build.DefaultPageNumber)), opts.f.Localizer.MustLocalize("kafka.list.flag.page"))
+	flags.IntVar(&opts.limit, "limit", 100, opts.f.Localizer.MustLocalize("kafka.list.flag.limit"))
```
dimakis, line 86: can we use the default limit here, similar to the way we deal with page?

pkg/cmd/kafka/list/list.go
```diff
@@ -164,7 +165,8 @@ func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId st
 
 		var customerCloud string
 		if id, ok := k.GetClusterIdOk(); ok {
-			customerCloud = *id
+			cluster := (*clusterIdMap)[*id]
```
dimakis, line 168: This is not printing out a cluster ID for me, just a pair of emptys parenthesis

pkg/cmd/kafka/list/list.go
jackdelahunt: Done

pkg/cmd/kafka/list/list.go
```diff
@@ -164,7 +165,8 @@ func mapResponseItemsToRows(kafkas []kafkamgmtclient.KafkaRequest, selectedId st
...
+			customerCloud = fmt.Sprintf("%v (%v)", cluster.Name(), cluster.ID())
 		} else {
 			customerCloud = "Red Hat Infrastructure"
```
dimakis, line 171: This should be localized.

pkg/cmd/kafka/list/list.go
```diff
@@ -38,46 +37,38 @@ type kafkaRow struct {
...
-		page:           0,
-		limit:          100,
-		search:         "",
-		Connection:     f.Connection,
-		Logger:         f.Logger,
-		IO:             f.IOStreams,
-		localizer:      f.Localizer,
-		Context:        f.Context,
-		ServiceContext: f.ServiceContext,
+		page:   0,
```
dimakis, line 53: why are we setting these here? the best place comes from the default values in the flag?

pkg/cmd/kafka/list/list.go
dimakis: thas the `Red Hat Infrastucture`

pkg/cmd/kafka/list/list.go
dimakis: it does work now since the last push, I'm wondering if this is the best way to display it? I get it makes sense during an interactive selection prompt, but since we already have a table with one value for column then I think we should stick to that. i.e. have a `Hybrid Cluster Name` and a `Hybrid Cluster ID`

pkg/cmd/kafka/list/list.go
jackdelahunt: Done

pkg/cmd/kafka/list/list.go
jackdelahunt: Done

pkg/cmd/kafka/list/list.go
jackdelahunt: Weird not sure why I added that 🤔

pkg/cmd/kafka/list/list.go
jackdelahunt: Yeah that is an interesting point, what would we put there, and I am thinking if all you have is normal kafkas then two columns of info that doesn't apply might be messy

jackdelahunt: @dimakis and I will get back to this conversation around formats and naming in the future

dimakis (approved):
LGTM, Great Work

PR #1836, opened 2023-03-13
refactor: hide flags, improves logic around passing cluster flag in r…

dimakis:
…egister-cluster

addresses review comments

Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/dedicated/deregister/deregister.go
+++ b/pkg/cmd/dedicated/deregister/deregister.go
@@ -5,7 +5,10 @@ import (
 	"fmt"
 	"github.com/AlecAivazis/survey/v2"
 	clustersmgmtv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/redhat-developer/app-services-cli/internal/build"
+	"github.com/redhat-developer/app-services-cli/pkg/cmd/dedicated/dedicatedcmdutil"
 	kafkaFlagutil "github.com/redhat-developer/app-services-cli/pkg/cmd/kafka/flagutil"
+	"github.com/redhat-developer/app-services-cli/pkg/core/cmdutil"
 	"github.com/redhat-developer/app-services-cli/pkg/core/localize"
 	"github.com/redhat-developer/app-services-cli/pkg/shared/connection/api/clustermgmt"
 	ocmUtils "github.com/redhat-developer/app-services-cli/pkg/shared/connection/api/clustermgmt"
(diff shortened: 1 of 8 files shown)


pkg/cmd/dedicated/deregister/deregister.go
```diff
@@ -54,6 +56,8 @@ func NewDeRegisterClusterCommand(f *factory.Factory) *cobra.Command {
 	flags.StringVar(&opts.accessToken, "access-token", "", f.Localizer.MustLocalize("dedicated.deregistercluster.flag.accessToken.description"))
 	flags.StringVar(&opts.selectedClusterId, "cluster-id", "", f.Localizer.MustLocalize("dedicated.deregisterCluster.flag.clusterId.description"))
 
+	_ = flags.MarkHidden("cluster-mgmt-api-url")
```
andreaTP, line 59: Can you refactor those 2 lines into a function?
It will make it easier to be removed, I hope 🙂

pkg/cmd/dedicated/register/registercluster.go
```diff
@@ -117,23 +115,29 @@ func runRegisterClusterCmd(opts *options) (err error) {
...
-func validateClusters(clusters *clustersmgmtv1.ClusterList, cls []clustersmgmtv1.Cluster) []clustersmgmtv1.Cluster {
-	for _, cluster := range clusters.Slice() {
-		if cluster.State() == clusterReadyState && cluster.MultiAZ() == true {
-			cls = append(cls, *cluster)
-		}
+func validClusterString() string {
+	return "multi_az='true' AND state='ready'"
+}
+
+func getPaginatedClusterList(opts *options) error {
```
andreaTP, line 131: I'm not sure about this, and it might be common practice in Go ... consider this a nitpick.
I expect a method named `getPaginatedClusterList` to actually return the cluster list instead of enriching an object passed as an argument.

pkg/cmd/dedicated/register/registercluster.go
```diff
@@ -117,23 +115,29 @@ func runRegisterClusterCmd(opts *options) (err error) {
...
+	cl, err := clustermgmt.GetClusterListWithSearchParams(opts.f, opts.clusterManagementApiUrl, opts.accessToken, validClusterString(), opts.pageNumber, opts.pageSize)
+	if err != nil {
+		return err
 	}
-	return cls
+
+	for _, cluster := range cl.Slice() {
+		opts.clusterList = append(opts.clusterList, *cluster)
```
andreaTP, line 138: What happens if this function gets called twice?
Should we avoid possible duplicates?

pkg/core/localize/locales/en/cmd/dedicated.en.toml
```diff
@@ -26,7 +26,10 @@ one = 'The access token to use to authenticate with the OpenShift Cluster Manage
 one = 'The ID of the OpenShift cluster to deregister'
 
 [dedicated.deregisterCluster.run.noClusterFound]
-one = 'No valid OpenShift clusters found'
+one = '''
+No valid OpenShift clusters found.
+Try passing a different page number or increasing the page size.
```
andreaTP, line 31: Are we asking the user to be aware of the API pagination?
I think this should remain hidden as an implementation detail and never be exposed to the user.

andreaTP (commented):
Nice job! A couple of comments @dimakis !

pkg/core/localize/locales/en/cmd/dedicated.en.toml
dimakis: This is here incase the user passes in a page via the page flag and the the cluster isn't in said page.
should this not be there then for this  case??

andreaTP: oh, didn't knew we already have a "page" flag 😞

PR #1837, opened 2023-03-20
feat: make enterprise flow the primary flow

dimakis:
Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -516,7 +516,7 @@ func checkForLegacyQuota(opts *options, orgQuotas *accountmgmtutil.OrgQuotas) {
 			opts.useEnterpriseFlow = true
 		}
 	}
-	// to-do may have to deal with trial (devoloper instances) quota here
+	// to-do may have to deal with trial (developer instances) quota here
 
 	// Check if the user has a legacy quota
 	for _, quota := range orgQuotas.StandardQuotas {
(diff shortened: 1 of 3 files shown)


pkg/cmd/kafka/create/create.go
```diff
@@ -582,7 +612,7 @@ func promptKafkaPayload(opts *options, constants *remote.DynamicServiceConstants
 	}
 
 	// If the user is not using an enterprise cluster, prompt them to select a cloud provider
-	if !opts.useEnterpriseCluster {
+	if opts.hasLegacyQuota && opts.useLegacyFlow {
```
jackdelahunt, line 615: Nit pick but couldn't this be just `opts.useLegacyFlow` as if that is true then `opts.hasLegacyQuota` is always true

jackdelahunt (approved):
I had one comment but it was smalled and can be ignored if you want

pkg/cmd/kafka/create/create.go
dimakis: makes sense 👍

pkg/cmd/kafka/create/create.go
```diff
@@ -501,29 +503,64 @@ func selectEnterpriseOrRHInfraPrompt(opts *options) error {
...
+	// check if org has legacy (non-hybrid) quota
+	checkForLegacyQuota(opts, orgQuota)
+
+	var enterpriseQuota accountmgmtutil.QuotaSpec
 
 	answers := &promptAnswers{}
 
-	err := survey.AskOne(promptName, &answers.Name, survey.WithValidator(validator.ValidateName), survey.WithValidator(validator.ValidateNameIsAvailable))
+	if !opts.useEnterpriseFlow {
+		f.Logger.Info(opts.f.Localizer.MustLocalize("kafka.create.info.enterpriseQuota"))
```
jackdelahunt, line 561: So are logging to show how to get enterprise quota every time they don't use the new standard flow even if they have legacy?

dimakis: yes, seeing as enterprise is rhosak now, we want to sell it. if they don't have enterprise quota, tell them where to get it to give us money. This may change when know what the craic is with trials but until then I think this is the best method

jackdelahunt (approved):
Just need CI to work

dimakis:
> Just need CI to work

are you looking for the failing test to pass? this test always fails right? 🤦

PR #1840, opened 2023-03-22
feat: now support trial instances for kafka create

jackdelahunt:
### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -6,10 +6,12 @@ import (
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/redhat-developer/app-services-cli/internal/build"
 	"github.com/redhat-developer/app-services-cli/pkg/cmd/dedicated/dedicatedcmdutil"
+	"github.com/redhat-developer/app-services-cli/pkg/core/auth/token"
 	"github.com/redhat-developer/app-services-cli/pkg/shared/connection/api/clustermgmt"
 	"github.com/redhat-developer/app-services-cli/pkg/shared/kafkautil"
 	"os"
 	"os/signal"
+	"strconv"
 	"time"
 
 	kafkaFlagutil "github.com/redhat-developer/app-services-cli/pkg/cmd/kafka/flagutil"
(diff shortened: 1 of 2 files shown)


pkg/cmd/kafka/create/create.go
```diff
@@ -809,12 +833,48 @@ func determineFlowFromQuota(opts *options) error {
...
+
+	userName, ok := token.GetUsername(cfg.AccessToken)
+	if !ok {
+		return false, opts.f.Localizer.MustLocalizeError("kafka.create.error.cannotFindUserDetails")
+	}
+
+	list, _, err := api.KafkaMgmt().GetKafkas(opts.f.Context).Page(strconv.Itoa(1)).Size(strconv.Itoa(99)).Execute()
+	if err != nil {
+		return false, err
+	}
```
jackdelahunt, line 866: Another example of when having ONLY pagination creates pain, I am not sure what to do here to make sure we are always getting every Kafka in the org @dimakis any thoughts

pkg/cmd/kafka/create/create.go
```diff
@@ -809,12 +833,48 @@ func determineFlowFromQuota(opts *options) error {
...
+
+	for i := int32(0); i < list.GetSize(); i++ {
+		kafka := list.Items[i]
+		if kafka.GetOwner() == userName && kafka.GetInstanceTypeName() == TrialInstanceType {
```
machi1990, line 870: ideally the API should allow to filter per instance_type. So that the call in line 863 is reduced to only filtering the developer instances that are in a org.
Can you open a JIRA about this and ping @pbraun

pkg/cmd/kafka/create/create.go
machi1990: left a comment about what needs to be done from the API: https://github.com/redhat-developer/app-services-cli/pull/1840/files#r1146358149 then this call only becomes about filtering for a developer kafka owned by the user. If the list is empty then we are good. If the list is not empty then it'll be one element.

pkg/cmd/kafka/create/create.go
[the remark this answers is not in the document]
jackdelahunt: Done

PR #1842, opened 2023-03-23
fix: allows for the creation of trial instances

dimakis:
Closes # 

### Verification Steps

1. Do x
2. Do y

### Type of change

- [ ] Documentation change
- [ ] Other (please specify)

### DIFF
--- a/pkg/cmd/kafka/create/create.go
+++ b/pkg/cmd/kafka/create/create.go
@@ -78,6 +78,7 @@ type options struct {
 	useEnterpriseFlow       bool
 	hasLegacyQuota          bool
 	useLegacyFlow           bool
+	useTrialFlow            bool
 	clusterManagementApiUrl string
 	accessToken             string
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/kafka/create/create.go
```diff
@@ -518,29 +519,29 @@ func checkForLegacyQuota(opts *options, orgQuotas *accountmgmtutil.OrgQuotas) {
...
 			return
 		}
 	}
-	for _, quota := range orgQuotas.EvalQuotas {
+
+	// Check if the user has a trial quota, if so they ONLY have trial quota
+	for _, quota := range orgQuotas.TrialQuotas {
 		if quota.Quota > 0 {
-			opts.hasLegacyQuota = true
+			opts.useTrialFlow = true
```
jackdelahunt, line 544: Didn't we find out trial is always 0 and it is only known if they have trial quota left if they have no trial instances running?

dimakis: yes you are correct. I should have checked the changes, more thoroughly. give me a min and I'll correct 👍

PR #1850, opened 2023-04-11
feat: now mapping cluster status to custom status

jackdelahunt:
Adds custom status output for the status of the cluster

Before:
image

After:
image

### Verification Steps

1. Do `rhoas kafka oc list` to see the new output

### Type of change


jackdelahunt:
Note: I am use the phrase `Unregistering` here even though we normally think of it as `Deregistering`, this choice was made by the UI team so it needs to be consistent to how they will map the status

